PR #5, opened 2023-07-20
Frankly

henpemaz:
Merge slash rework of @[FranklyGD](https://github.com/henpemaz/Rain-Meadow/commits?author=FranklyGD)'s local lobby version of the mod.

Currently very buggy still don't merge just peek

Limited to loopback local, and non interoperable with steam in the same build, but as easy as switching a build setting.

Includes a new internal api-ish if we want to introduce even more matchmaking systems.


henpemaz:
Good stuff keep it up @FranklyGD

PR #8, opened 2023-07-30
realized po state only updates when necessary

cheapcents:


henpemaz:
Good stuff! Keep it up @cheapcents !

PR #9, opened 2023-08-01
ping calculations, shows up in debug view

cheapcents:


Online/DebugOverlay.cs
```diff
@@ -152,35 +197,37 @@ public static void Update(RainWorldGame self, float dt)
...
-                        if (player.eventsWritten)
-                        {
-                            label.color = new Color(1, 0.5f, 0);
-                        }
-                        overlayContainer.AddChild(label);
-                        outgoingLabels.Add(label);
-                        line++;
-                    }
+                    var player = idv.player;
+                    var playerTruePing = Math.Max(1, player.ping - 16);
```
cheapcents, line 203: can OnlineManager's rawUpdate fire more than 60 times p second? If so, this hardcoded var should be changed...

PR #11, opened 2023-08-03
Basic Overseer position synchronization

Dev-Lope:
Synchronizes overseer positions

The graphics module doesnt synchronize, causing issues with zips for remotes.

Other overseer synchronizations are not handled by this PR. See e.g. overseer type, death and holograms.


henpemaz:
Very good stuff keep it up!

PR #13, opened 2023-08-10
Slightly better error handling when joining lobbies, added loading/error dialog to the lobby list and sending the user there when joining through steam

MatheusVigaro:


henpemaz:
Looks great!

PR #15, opened 2023-08-21
Sync dangle fruit

Chrometurtle:
Simple dangle fruit syncing. It won't crash now


Game/RainMeadow.EntityHooks.cs
```diff
@@ -29,6 +30,17 @@ private void EntityHooks()
...
+            On.UpdatableAndDeletable.Destroy += UpdatableAndDeletable_Destroy;
+        }
+
+        private void UpdatableAndDeletable_Destroy(On.UpdatableAndDeletable.orig_Destroy orig, UpdatableAndDeletable self)
+        {
+            //Adding a delay on mark for garbage collection. Need to give network a chance to see an object has been destroyed
+            Task.Run(async () => {
+                await Task.Delay(250);
+                orig(self);
+            });
```
henpemaz, line 43: I disagree, you're "waiting for state to sync" while instead this should be handled as an event.
Is the entity system not picking this up as an "entity removed" event?

GameModes/OnlineGameModeHelpers.cs
```diff
@@ -12,7 +12,7 @@ public static class OnlineGameModeHelpers
             //PlacedObject.Type.PuffBall,
             //PlacedObject.Type.TempleGuard,
             PlacedObject.Type.LightFixture,
-            //PlacedObject.Type.DangleFruit,
+            PlacedObject.Type.DangleFruit,
```
henpemaz, line 15: Fruits aren't cosmetic. Lets figure out a way to fit this into our gamemode API, if we need more methods/filters lets make them!

Online/State/RealizedDangleFruitState.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        public override void ReadTo(OnlineEntity onlineEntity)
+        {
+            var fruit = (DangleFruit)((OnlinePhysicalObject)onlineEntity).apo.realizedObject;
+            fruit.bites = bites;
+            if (bites < 3) { RainMeadow.Debug($"Bites written to: {bites}"); }
+            if (hasStalk && fruit.stalk == null)
+            {
+                fruit.stalk = new DangleFruit.Stalk(fruit, fruit.room, this.chunkStates[0].pos);
+                fruit.room.AddObject(fruit.stalk);
```
henpemaz, line 39: Nitpick but if this was the whole reason you made chunkstates protected for, you could have instead just run base() first then use fruit.chunks[0].pos

PR #18, opened 2023-09-14
Arena menu and more configurations

Silvyger:
- Functioning arena mode for first player only
- Minor project adjustments for convenience
- Adds more configurations to avoid having a constant working change


Chrometurtle (approved):
lgtm

PR #19, opened 2023-09-19
Entity fixes

henpemaz:


Chrometurtle (approved):
If you could drop a comment to explain the OnlineField Group syntax you used that would be great

Online/Resource/WorldSession.RainCycle.cs
```diff
@@ -14,7 +7,7 @@ public class RainCycleData : OnlineState
         {
             [OnlineField]
             public int cycleLength = 0;
-            [OnlineField]
+            [OnlineField(group:"timer")]
```
Chrometurtle, line 10: This Line

henpemaz:
> If you could drop a comment to explain the OnlineField Group syntax you used that would be great

That one value is ever changing, so if we're grouping changes in the delta it's beneficial it's on its own group 
Deltas are more than "something in this whole object changed". there's field groups and one change flag per group

PR #23, opened 2023-09-24
simple fly-sync

Chrometurtle:


Game/RainMeadow.GameHooks.cs
```diff
@@ -29,13 +29,26 @@ private void GameHooks()
...
             
             // Arena specific
             On.GameSession.AddPlayer += GameSession_AddPlayer;
         }
 
+        private void Room_ReadyForAI(On.Room.orig_ReadyForAI orig, Room self)
+        {
+            if (OnlineManager.lobby != null)
+            {
+                if (!RoomSession.map.TryGetValue(self.abstractRoom, out var rs)) return;
```
henpemaz, line 44: Doesn't run orig if you do it this way. This is a "should never happen" scenario though so probably doesn't matter.

Online/State/AbstractCreatureState.cs
```diff
@@ -22,6 +22,7 @@ protected override RealizedPhysicalObjectState GetRealizedState(OnlinePhysicalOb
             if (onlineObject.apo.realizedObject is Player) return new RealizedPlayerState((OnlineCreature)onlineObject);
             if (onlineObject.apo.realizedObject is Overseer) return new RealizedOverseerState((OnlineCreature)onlineObject);
             if (onlineObject.apo.realizedObject is Creature) return new RealizedCreatureState((OnlineCreature)onlineObject);
+            if (onlineObject.apo.realizedObject is Fly) return new RealizedFlyState((OnlineCreature)onlineObject);
```
henpemaz, line 25: Fly is creature so this never runs (the previous line returns)

Game/RainMeadow.GameHooks.cs
Chrometurtle: Found a better function to hook into

Online/State/AbstractCreatureState.cs
Chrometurtle: fixed

PR #24, opened 2023-09-26
Add player list in lobby menu

RaineGit:
Adds a player list in the lobby menu and a player count in the lobby selection menu


henpemaz:
Very cool!

PR #33, opened 2023-12-06
Fix vanilla arena bugs

SortaUnknown:


henpemaz:
Good stuff keep it up!

PR #35, opened 2023-12-09
Fix arena crash

Chrometurtle:
Turn on custom music when the game mode is meadow only.


henpemaz:
Good stuff, keep it up!

PR #51, opened 2024-01-15
Story/42 save state sync

Chrometurtle:


Chrometurtle:
TODO:
- Wakeup in the proper shelter, not the lobby owners
- connect the check for ready/continue to savestate sync
- Save to multiplayer specific file

PR #54, opened 2024-01-19
StoryMenu

### LINKED ISSUE #43: Story: Lobby needs character selection
Not an urgent need yet, just opening to track it

6fears7:
image

Tinkered with some of the existing data we had on hand from MeadowMenu and combined it with LobbyMenu to form "Story Menu". It's ugly, copy and pasted code with some slight tweaks, but wanted to demo as a POC.


6fears7:
@henpemaz, would like your opinion. 

I'm wondering if you would want MeadowMenu to also support the menu-with-lobby design (meaning: we can set a UI base and then split from it), or if you'd prefer to keep MeadowMenu and StoryMenu completely separated designs with code that only overlaps by coincidence.

6fears7:
Spoke with henpe, we will NOT share the same UI design for MeadowMenu.

6fears7:
Really agonized over this decision but:

I think it's better for us to pursue the StoryMenu process manager versus SlugcatSelectMenu for a few reasons:

1. If we run the OG selection menu, we will have to do a number of hooks for items like pulling from multiplayer save data, hooking lobby information, etc.. 

2. This would also make it harder down the line to implement any custom emoji work someone might want to do for story mode. 

StoryMenu would allow us the flexibility to control exactly what we want on-screen and want to do when the game starts. The trade-off is that we will have to write some code that duplicates work the game could already do for us if we just leaned into the SCSelectMenu. 


TL;DR - We either do a lot of hooks, or we do less hooks with more code writing

6fears7:
image

6fears7:
Just gotta rebase from some lobby changes y’all made and then we can test. 

Long term May need to figure this out better but the .Remove on a subObject doesn’t seem to actually delete the item so if a client and host both have different UI elements (like a button), despite them showing up there is an invisible barrier blocking the client unless the element is moved. 

For now I just moved the client start button to the more to the center of the frame so it’s clickable. Function over form in the meantime.

6fears7:
@Chrometurtle we are good to go on the logic side, now to test Steam.

6fears7:
image

henpemaz:
I don't understand the need for the `DidStartGame` RPC.

I presumed only the lobby host would be able to start the game, and in that case it can update the lobby state without an rpc.

Is the idea here that 'anyone' can become the world-host by pressing the start button, so it might not be the same as the lobby-host?

6fears7:
When discussing with Turtle, he mentioned clients can click "Enter" before the host does and as a result there lies a potential for a race condition if two people go in at the same time, which creates issues on the initial setup. So the RPC protects against that until after the host is able to initiate the game session first.

henpemaz:
It's good!

Story/StoryMenu.cs
```diff
@@ -0,0 +1,337 @@
...
+
+        private void SetupMenuItems()
+        {
+
+            // Host button
+            this.hostStartButton = new EventfulHoldButton(this, this.pages[0], base.Translate("ENTER"), new Vector2(683f, 85f), 40f);
+            this.hostStartButton.OnClick += (_) => { StartGame(); };
+            hostStartButton.buttonBehav.greyedOut = false;
+
+            // TODO: clientWaitingButton needs to not require x/y shift to function. Look into .Remove() on subObjects.
```
henpemaz, line 236: you might be looking for `MenuObject.RemoveSubObject` (removes child from update tree) and `MenuObject.RemoveSprites` (cleanup graphics), both are needed

henpemaz (approved):

Chrometurtle (approved):
lgtm

PR #90, opened 2024-02-22
Storymenu rework

6fears7:
Notables from this PR:

- StoryMenu is reworked to leverage more IL to save significant pain for added characters and interactions with the menu
- MSC slugcat selections are soft added (commented out). There's more work that needs to be done regarding the MSC check if the slugs are unlocked as it panics currently. 
- I have no idea what Sofanthiel's scene ID is or if they have one (Night?)
- Host can now play as the three original slugcats in outskirts with the expected stats  / functionality, clients inherit host's decision
- Arrows are disabled for client to force them to take the host's slugcat name / functionality
- This can be merged after the UI development at @henpemaz's discretion to prevent merge pains
- Even if you play as another slugcat, you will default to Color.white if you don't choose another option. Your tunnel traversal color will reflect who you really are. 
- Review at your discretion, there is no rush on this. Had a lot of free time today.


6fears7:
Closing this for now to disable character selection save slot wiping

PR #97, opened 2024-02-24
Storymenu rework

6fears7:
See #43 for more details. Will be draft until UI rework is complete, then I will turn off host ability to choose other slugcats to preserve save file corruption until we can get some new save files up


6fears7:
@Chrometurtle thisbis ready to go.

Features: 
- Meadow remix options 
- Experimental section in remix menu for unsafe but toggleable features, warns the user to backup data 
- reworks story menu for the future 
- Adds toggleable experimental feature for hosts to choose campaigns that aren't survivor and for clients to choose slugcats different than the host 
- adds a new OnlineField to StoryLobbyData - current campaign. 
- host chooses a campaign to play, information is passed to clients before they enter

6fears7:
After discord discussion, I will update the PR to lay groundwork for save slots so that players aren't choosing a real slugcat name they're choosing an alias we are repurposing.

`MeadowOnline was the default savestate for our online survivor. Idk where exactly the starting point is in your PR, but my starting assumption would be in `SlugcatCustomSelection`, instead of SlugcatStats.Name we do Ext_slugcatStatsname`

6fears7:
I will also shift these into a SlugcatStatsHook.cs likely as we'll need to keep building that out if we want full functionality, unless you know a different route to take

6fears7:
should fix #104  . Calling original method on slugcat ctor resolved collision issue

PR #108, opened 2024-03-04
friends name display: hope the fade away doesn't end in a rim shot

6fears7:
You can't hear it, but @Chrometurtle is crying right now from the simplicity of this PR

### DIFF
--- a/Story/OnlineUIComponents/OnlinePlayerArrow.cs
+++ b/Story/OnlineUIComponents/OnlinePlayerArrow.cs
@@ -42,25 +42,36 @@ namespace RainMeadow
         public override void Update()
         {
             base.Update();
-            this.lastAlpha = this.alpha;
-            this.lastBlink = this.blink;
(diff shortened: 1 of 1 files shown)


Story/OnlineUIComponents/OnlinePlayerArrow.cs
```diff
@@ -42,25 +42,36 @@ public OnlinePlayerArrow(PlayerSpecificOnlineHud owner) : base(owner)
...
+
             }
-            
-            this.alpha = Custom.LerpAndTick(this.alpha, owner.needed ? 1 : 0, 0.08f, 0.033333335f);
 
-            this.counter++;
         }
 
-        public override void Draw(float timeStacker)
+    public override void Draw(float timeStacker)
```
Chrometurtle, line 74: D:< tabbing

Chrometurtle: lgtm otherwise

PR #113, opened 2024-03-06
vulture mask sync

6fears7:
Tested and checked, I think this item falls into the same category as a "rock" in terms of its properties requiring syncing. 

Two player in-game: P1 could grab the mask and wear it, P2 could pull the mask off P1 and wear it. Mask could be dropped. 

As this is my first item to sync I'm not sure if there's usually a 'tell' when it's not working but unless creature in-game behavior has to be synced (which I would think it wouldn't even be handled here), then this is working as I expect, hence my lack of changes.

### DIFF
--- a/GameModes/OnlineGameModeHelpers.cs
+++ b/GameModes/OnlineGameModeHelpers.cs
@@ -94,7 +94,7 @@ namespace RainMeadow
             PlacedObject.Type.BubbleGrass,
             PlacedObject.Type.Hazer,
             PlacedObject.Type.DeadHazer,
-            PlacedObject.Type.VultureMask,        //AbstractPhysicalObject Child MEDIUM
+            PlacedObject.Type.VultureMask,
             //PlacedObject.Type.HangingPearls,      //MSC?
             PlacedObject.Type.Lantern,
         };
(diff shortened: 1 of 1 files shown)


Chrometurtle:
Vulture mask requires an online counterpart for it's abstract class. It's not that simple

Chrometurtle (changes requested):
Needs Online counterpart for abstract vulture mask

Chrometurtle:
Observe commit 0ae79ecaf1383084ca26e47319ca0126cd00b78c

6fears7:
I'm tracking on why it might need an Online counterpart, but then I look at something like[ PlaceObject.Type.Jellyfish](https://github.com/henpemaz/Rain-Meadow/compare/main...sync-story-items#diff-995a3ae1dbbc6cf44db34850736e40ac50d67b163b0418025afc89bba3cc7211R82) or PuffBall and there's no specific  Online counterpart class to speak of aside from existing as a HashSet inside Player Grabbable items, and I'm wondering if 

1. They were just accidentally left out 
or 
2. There weren't any properties about it that required synchronous intervention, which is what seems like is also true for VultureMask.

Chrometurtle:
wrong on both accounts.

Look at the vulture mask code inside rainworld. The abstract counterparts are not simply abstractPhysicalObject

6fears7:
Ohhhhhh

Chrometurtle:
kk, did a deep dive on abstractVultureMask. The things that would need to be synced only fix graphical issues if we included downpour DLC. For now, no work needs to be done since I have no plans on getting downpour working.

PR #115, opened 2024-03-10
Meadow things

henpemaz:
Emote pickers embetterned

Emote handling embetterned

Creature head/look control prototype

Creature controller unification work

Colorpicker vastly simplified, works better on non-mouse scenarios

Death protection and fall protection for meadow-mode


Menu/Objects/OpTinyColorPicker.cs
```diff
@@ -1,20 +1,24 @@
...
+        private int loseFocusCounter;
 
         public OpTinyColorPicker(Menu.Menu menu, Vector2 pos, string defaultHex) : base(pos, new Vector2(30, 30))
         {
-            //this.colorPicker = new OpColorPicker(pos + new Vector2(-60, 24), "", defaultHex);
-            this.colorPicker = FlatColorPicker.MakeFlatColorpicker(menu, pos + new Vector2(-60, 30), defaultHex);
+            this.colorPicker = new OpColorPicker(new Configurable<Color>(MenuColorEffect.HexToColor(defaultHex)), pos);
+            UIelementWrapper wrapper = new UIelementWrapper((menu as SmartMenu).tabWrapper, colorPicker);
+            colorPicker.Hide();
+
```
6fears7, line 21: @henpemaz I meant to bring this up awhile ago but got distracted. Thoughts on making the color picker re-sizable? 

Or would you prefer me to implement a separate one that has custom sizing?

```
// pseudo code
public OpTinyColorPicker(Menu.Menu menu, Vector2 pos, string defaultHex, Vector2 customSize)

this._size = customSize // I believe _size was the property that managed the size of the square.
```

henpemaz: I have no idea if that would work and given it bakes a texture to display the hsl map as well as allowing mouse input I really doubt it works well while scaled, you can give it a try though, doesn't have to be separate as long as it works

PR #123, opened 2024-03-15
Sync story items

Chrometurtle:


Online/Entity/EntityDefinition.cs
```diff
@@ -14,6 +14,13 @@ public abstract class EntityDefinition : OnlineState, IIdentifiable<OnlineEntity
 
         public EntityDefinition() : base() { }
 
+        public EntityDefinition(EntityDefinition ed) : base() 
+        { 
+            this.entityId = ed.entityId;
```
Chrometurtle, line 19: @henpemaz is this bad practice in c#?

PR #135, opened 2024-04-08
Mostly Meadow things

henpemaz:


Ext_.cs
```diff
@@ -31,11 +31,16 @@ public class Ext_SceneID
             internal static MenuScene.SceneID Slugcat_MeadowSquidcicada = new("Slugcat_MeadowSquidcicada", true);
             internal static MenuScene.SceneID Slugcat_MeadowLizard = new("Slugcat_MeadowLizard", true);
             internal static MenuScene.SceneID Slugcat_MeadowScav = new("Slugcat_MeadowScav", true);
+            internal static MenuScene.SceneID Slugcat_MeadowEggbug = new("Slugcat_MeadowEggbug", true);
```
Chrometurtle, line 34: Would it be possible to get a separate list of added SlugcatStats.names so that for the saveManager I can just iterate over it to determine what is not a campaign slugcat?

PR #142, opened 2024-04-09
Easy mode hibernation

Chrometurtle:


Game/RainMeadow.EntityHooks.cs
```diff
@@ -284,162 +284,88 @@ private void OverWorld_WorldLoaded(On.OverWorld.orig_WorldLoaded orig, OverWorld
 
                 Room room = null;
 
-                if (true)
```
henpemaz, line 287: git really didn't like the indentation change, is this all just removing the if(true) and braces?

Chrometurtle: yeah, there's an additional change for changedRegion.

PR #147, opened 2024-04-20
Draft: bare minimum for generic hud between modes

6fears7:
This adds some basic Arena files in preparation for the larger changes coming between myself and @Silvyger  have done. 

Makes the online hud generic across game modes (useful for arena as pictured):

image


6fears7:
Going to implement this after getting arena changed between myself and Silver so the PR will already have all the networking arena needs for this to fill work

PR #153, opened 2024-05-15
Arena world loading

6fears7:


Online/Resource/OnlineResource.Transactions.cs
```diff
@@ -8,8 +8,56 @@ public abstract partial class OnlineResource
         public void Request()
         {
             RainMeadow.Debug(this);
-            if (isPending) throw new InvalidOperationException("pending");
-            if (isAvailable) throw new InvalidOperationException("available");
+            if (isPending)
+                if (RainMeadow.isArenaMode(out var _))
```
6fears7, line 12: @henpemaz this file is where I spent a lot of time in pain. The solution below is what I found to work, FWIW.

PR #168, opened 2024-07-17
Arena fix creature despawn

6fears7:
- Fixes creature despawn crashing client games


6fears7:
image


Running this code prevents the creature desync for arena, though how to manage it in story would require more investigation since arena needs and story needs are quite different for entity ownership

Issue #175, opened 2024-07-23
Arena: Prevent join until host is ready


6fears7:
Thinking there's some racing involved with players trying to join the world session before the host has established the game env. 

Host should be realized in-game and then call for the others to join.

6fears7:
This is done but will be merged along with the fix for spear violence (#173 ).

Issue #173, opened 2024-07-23
Arena: Ownership loss on spear stick


6fears7:
- [x] Check sSpawning avatars placement 
- [x] Remove orig loop to create players in SpawnPlayers (qol)

- [x] Ask game politely

6fears7:
Fixed the ownership loss. victim state is not being displayed correctly, however:

- [ ] Player states seem to not be syncing on violence.

Player 1 kills P2. P2- screen shows death, P1 shows just immobile slugcat but still blinking and not being registered as state = dead.

But falling off map works fine.

Need to grab logs for this.

PR #176, opened 2024-07-24
fix spear granting authority

6fears7:
… need to investigate the spear stick pending forever


6fears7:
Resolves #173

6fears7:
Resolves #175

6fears7:
In this branch:

- Spear fixed
- Client race condition fixes
- Creature state fix

Creature state fix was just re-adding realized creature's state to be opposite of abstract creature's state in the new if condition. I believe the reason this now works is due to Violence RPC calling Die() before the ReadTo does, which would explain why I saw it being used twice in the logs.

6fears7:
resolves #180

Issue #182, opened 2024-07-29
Arena: NextLevel is not stable for all clients


6fears7:
Once again in painful territority.

- [x] Test severing connection from host / clients completely in new world vs old world. No errors for missing entities should throw.
- [x] Revisit the changes made to the Request/Release paradigm for Arena and see if we can backtrack
- [x] I need to make a solution that also accommodates playing the same level back to back (may need to follow similar logic to re-using gates by copying the world session?)

Perhaps the flow should be:

1. Host exits current resource
2. Host requests next resource
3. Wait for host to establish session
4. Notify clients to leave (RPC or other)
5. Client request to join resource

Was initially concerned about clients having ownership of previous world, but maybe we need to play hot potato for the release to work properly. Having to enable re-usable worlds is going to force a change to occur in the current logic.

6fears7:
https://github.com/henpemaz/Rain-Meadow/blob/c4996552942fc388ed3f43c0388d8668cba6777b/Game/RainMeadow.LoadingHooks.cs#L264

Might be able to leverage this _Update to determine if the upcoming room is the same as the current room and re-use it.

6fears7:
Did some work with NextLevel where if the worldsession is available and active we deactivate it, otherwise we release it asap. 

Currently trying to figure out the next step. I can make a request at the end of NextLevel to a new arena world session but I get "world is not set" thrown because we aren't at the world loader step yet.

But if I release the world session in NextLevel I must request it then too. 

So thinking out loud: What instead I should do is perform the release / deactivate logic in the WorldLoader step so that if we have a worldsession that is available / active we turn it off then clear the way for the new world. 

Will test and report results.

PR #183, opened 2024-07-31
arena level switchery

6fears7:
Still a WIP. This enables a level to be replayed, still trying to sort out the client logic for leaving and joining this world.


6fears7:
Got a working prototype. Works for both playing the same arena levels and new levels. Just need to grab some bodies to test this with.

6fears7:
Closes #182

6fears7:
Rewrote entry / exit logic to leap frog host and client behavior. This works well, much better than previously. Tested online. Will find out over time if it scales but currently is promising.

PR #186, opened 2024-08-01
fixes starve pre cycle2

6fears7:

PR #187, opened 2024-08-02
better fix for starve /sleep

6fears7:


Story/RainMeadow.StoryHooks.cs
```diff
@@ -329,16 +329,10 @@ private void RainWorldGame_GoToDeathScreen(On.RainWorldGame.orig_GoToDeathScreen
             {
                 foreach (OnlinePlayer player in OnlineManager.players)
                 {
-                    if (OnlineManager.lobby.isOwner)
-                    {
-                        RPCs.GoToDeathScreen();
```
Chrometurtle, line 334: This is an optimization

Chrometurtle (changes requested):
You're removing an optimization

Story/RainMeadow.StoryHooks.cs
6fears7:
```
                  if (player.id == OnlineManager.lobby.owner.id)
                    {
                        RPCs.GoToDeathScreen();
                    }
                    else
                    {

                        player.InvokeRPC(RPCs.GoToDeathScreen);
                    }
```

Keeps optimization and prevents unintended repeat invocations

Chrometurtle: oh? What was causing the repeat invocations. There should have only ever been 1 lobby owner

6fears7: I believe it was from having a separate function called, "MovePlayersToDeathScreen" that simply iterated over all players and was called during this looping, so players were getting double-dosed with calls to move to sleep / death and dying of starvation rapidly. 

https://github.com/henpemaz/Rain-Meadow/pull/186/files

This solution should provide the same behavior while preventing repeats.

Chrometurtle (approved):

Story/RainMeadow.StoryHooks.cs
Chrometurtle: yup checks out.

PR #189, opened 2024-08-06
Sync grenades

6fears7:
Work from a branch a long time ago. Might be able to test this in story to measure impact to arena if I can't get grenades to spawn easily. 

Current impact to arena without this change: Desync on hit. 

- [x] Host holds two grenades, throws one. Both blow up. Fix this.
- [x] Determine why ReadTo is not printing logs. Something is amiss.


6fears7:
no more desyncs thanks to null serializations for OEs. Current struggle is determining how to get the explosions from scav grenades from the host to show for clients. Seems like clients can throw nades and hosts views it no problem. More investigating is required.

6fears7:
Findings so far:

I was curious about determining what the Scav bomb's CollisionWithTerrain boolean value was returning so I could see why clients weren't getting the data they needed to be notified to explode. 

https://github.com/henpemaz/Rain-Meadow/pull/189/files#diff-207aee8899fabbb3762ac236b6dfd0a751fa721c1bcf60eae2ad4fe4f11e39adR60


The data never prints. 

In contrast, when printing the same boolean on RealizedSpear I was receiving the expected information. I believe I also tested printing on WeaponState and was getting a mix of true and false. 

Scav Bomb's OnlineState and PhysObjectState are being set.

Bottom Line: Scav Bomb should be printing data in the ReadTo. Figure out why

6fears7:
Finally got bombs to sync. I was losing my mind why my ReadTo wasn't printing any info and I realized that the Weapon realizedobj state was 'on top of'  the scav bomb so it was defaulting to Weapon first and as a result my synced properties weren't being read in realized scav bomb state. 

Order matters.

image


https://github.com/user-attachments/assets/5ed49bed-2e5e-4659-95bc-e31d438060e3


Now for the second problem: Grenades in hand explode when another grenade is thrown

Online/Entity/OnlinePhysicalObject.cs
```diff
@@ -343,5 +343,14 @@ public static void HitByExplosion(OnlinePhysicalObject objectHit, float hitfac)
         {
             objectHit?.apo.realizedObject?.HitByExplosion(hitfac, null, 0);
         }
+
+
+        [RPCMethod]
+        public static void ScavengerBombExplode(OnlinePhysicalObject scavBomb)
+        {
+
+            (scavBomb.apo.realizedObject as ScavengerBomb)?.Explode(null);
```
Chrometurtle, line 352: You might want to pass in the exact pos of the bomb when it explodes so that the location of where it explodes is EXACTLY the same. Otherwise lag might cause a bomb to explode earlier for receiving clients

6fears7:
Before / After

https://github.com/user-attachments/assets/02c6814a-9d5f-48ca-b4b7-c67d1dc4b648

Game/RainMeadow.GameplayHooks.cs
```diff
@@ -345,6 +390,13 @@ private void CreatureOnViolence(On.Creature.orig_Violence orig, Creature self, B
                 {
                     if (!OnlinePhysicalObject.map.TryGetValue(trueVillain.abstractPhysicalObject, out var onlineTrueVillain))
                     {
+                        if (trueVillain.abstractPhysicalObject.type == AbstractPhysicalObject.AbstractObjectType.ScavengerBomb)
+                        {
+                            // scav bombs exit quickly, and that's ok.
+                            OnlinePhysicalObject onlineVillain = null;
+                            (onlineVictim as OnlineCreature).RPCCreatureViolence(onlineVillain, hitchunk?.index, hitappendage, directionandmomentum, type, damage, stunbonus);
```
6fears7, line 397: by the time client gets violenced, villain OE as already exited resource.

6fears7:
@Chrometurtle thanks for the pos tip. Saved me a lot of time. If you wanna give a second opinion on anything here that would be great.

Chrometurtle:
yeah let me review this one

GameModes/OnlineGameModeHelpers.cs
```diff
@@ -89,7 +89,7 @@ public static class OnlineGameModeHelpers
             PlacedObject.Type.FirecrackerPlant,
             PlacedObject.Type.VultureGrub,
             PlacedObject.Type.DeadVultureGrub,
-            //PlacedObject.Type.ScavengerTreasury,  //Spear, Explosive spear, Scav bomb, Pearl, & Laterns
```
Chrometurtle, line 92: Keep this commented out

PR #201, opened 2024-08-19
Sync spear lifetimes

### LINKED ISSUE #199: Sync: spear lifetime
important when transferring ownership

(this will also fix vertical spears being treated as horizontal poles, because the game keeps track of vertical/horizontal using the sign of the `stuckInWallCycles` property in `AbstractSpear`)

youbitchoc:
track spear lifetime in `RealizedSpearState`

bugs (needs testing):
- spear sprites can sometimes appear behind player (does this happen with jolly co-op?)
- plays the wrong sound when downspearing (might not be due to this change, could be a separate issue?)

closes #199

### DIFF
--- a/Online/State/RealizedSpearState.cs
+++ b/Online/State/RealizedSpearState.cs
@@ -17,12 +17,15 @@ namespace RainMeadow
         private sbyte stuckBodyPart;
         [OnlineField(group = "spear")]
         private float stuckRotation;
+        [OnlineField(group = "spear")]
+        private int stuckInWallCycles;
 
         public RealizedSpearState() { }
         public RealizedSpearState(OnlinePhysicalObject onlineEntity) : base(onlineEntity)
         {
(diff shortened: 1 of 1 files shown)


6fears7:
Thanks for the effort and testing in this. 

Going to see about addressing the sound before this is merged. Feel free to update this MR with information if you decide to investigate that before I do.

youbitchoc:
actually I'm not sure if the sound bug is due to the changes

Chrometurtle:
You'll need to look into the spear update() code. The way it currently syncs the audio wouldn't queue for players receiving the new "spear in ground". You'd have to detect it and play the sound yourself.

6fears7:
Looked into this. 

If "spinning" bool is synced, we get the proper stick in ground effect. However, this then impacts other SFX like the spear sticking in the wall or being picked up. 

As it stands, I'm fine with accepting the current missing sfx for ground spear as the trade off is everyone's ability to use a  down spear.

PR #205, opened 2024-08-20
Prevent clients initiating game over before host

6fears7:
Prevents clients from clicking screen during game over text update. Host will bring them into death screen


6fears7:
image

Story/RainMeadow.StoryHooks.cs
```diff
@@ -57,7 +57,33 @@ private void StoryHooks()
...
+        }
+
+        private void TextPrompt_Update(On.HUD.TextPrompt.orig_Update orig, TextPrompt self)
+        {
 
+            if (OnlineManager.lobby == null)
+            {
+                orig(self);
+                return;
+            }
```
henpemaz, line 70: early returns look pretty but make logic hard to track.

right now if an online gamemode that isn't storygamemode would try and use this nothing would run (and I'd say that's unexpected)

can you rework so that you're only altering the behavior for storygamemode?

Story/RainMeadow.StoryHooks.cs
```diff
@@ -57,7 +57,28 @@ private void StoryHooks()
             On.SLOracleSwarmer.BitByPlayer += SLOracleSwarmer_BitByPlayer;
             On.CoralBrain.CoralNeuronSystem.PlaceSwarmers += OnCoralNeuronSystem_PlaceSwarmers;
             On.SSOracleSwarmer.NewRoom += SSOracleSwarmer_NewRoom;
+            On.HUD.TextPrompt.Update += TextPrompt_Update;
+        }
+
+        private void TextPrompt_Update(On.HUD.TextPrompt.orig_Update orig, TextPrompt self)
```
Chrometurtle, line 63: There has to be a better place to do this than in update. You don't need to be Overwriting the gameoverstring every update cycle

Chrometurtle (changes requested):

Story/RainMeadow.StoryHooks.cs
6fears7: Unfortunately, the TextPrompt Update overwrites the gameoverstring both in the ctor and every Update cycle with the func, `UpdateGameOverString(string)` due to listening for what input the user has (controller, kb) and updates the message for how to initiate game over accordingly. If custom game over for clients  is set in the ctor alone, the Update frame would override it.

Chrometurtle: Does hooking into updateGameOverString get over written?

6fears7: Good thinking. Can set the client game over string differently. Will still need the restartNotAllowed here to block clients from going to game over screen.

6fears7: Updated. Blocks clients from intiating, less invasive hooking, and better blocking on text prompt so it only happens during death screen ID

6fears7: Will update this weekend to allow clients to remove the death text so they can properly quit the game if they don't feel like waiting for the host.

6fears7:
Video of change

https://github.com/user-attachments/assets/bcee7e47-56d7-48b0-920c-692ff66d163a

Story/StoryHooks.cs
```diff
@@ -102,19 +156,19 @@ private void SLOracleSwarmer_BitByPlayer(On.SLOracleSwarmer.orig_BitByPlayer ori
...
             orig(self, grasp, eu);
-            if (self.slatedForDeletetion == true) 
+            if (self.slatedForDeletetion == true)
             {
                 SwarmerEaten();
             }
         }
 
-        private void SwarmerEaten() 
+        private void SwarmerEaten()
```
Chrometurtle, line 165: Next time do these formatting changes in the different PR. It makes reading diffs harder

Story/StoryHooks.cs
```diff
@@ -57,14 +58,67 @@ private void StoryHooks()
...
+
+        private void TextPrompt_Update(On.HUD.TextPrompt.orig_Update orig, TextPrompt self)
+        {
+            orig(self);
+            if (OnlineManager.lobby.gameMode is StoryGameMode)
+            {
+
+                if (!OnlineManager.lobby.isOwner && self.currentlyShowing == TextPrompt.InfoID.GameOver)
+                {
+                    self.restartNotAllowed = 1; // block clients from GoToDeathScreen
```
Chrometurtle, line 99: Not sure how this blocks clients. It looks like it's only used to determine if the hud should play the game over sound.

Story/StoryHooks.cs
```diff
@@ -57,14 +58,67 @@ private void StoryHooks()
...
+                    bool touchedInput = false;
+
+                    // let clients still have access to pause menu
+                    for (int j = 0; j < self.hud.rainWorld.options.controls.Length; j++)
+                    {
+                        touchedInput = ((self.hud.rainWorld.options.controls[j].gamePad || !self.defaultMapControls[j]) ? (touchedInput || self.hud.rainWorld.options.controls[j].GetButton(5) || RWInput.CheckPauseButton(0, inMenu: false)) : (touchedInput || self.hud.rainWorld.options.controls[j].GetButton(11)));
+                    }
+                    if (touchedInput)
+                    {
+                        self.gameOverMode = false;
```
Chrometurtle, line 109: Hook into gameovermodeActive instead of doing it in an update function

Chrometurtle (changes requested):

Story/StoryHooks.cs
Chrometurtle: or hook into showpausemenu

6fears7: Going to wait for @youbitchoc to address or abandon the following changes they desire to make that I saw in Discord:

```
host dies, someone shelters -> should count as win
host dies, everyone dies too -> should count as loss
```

since that would impact how we currently are discussing changing gameover text menus.

Story/StoryHooks.cs
6fears7: Yeah it happens from me hitting shift+ctrl then k to autoformat  the current document.

Story/StoryHooks.cs
6fears7: image

`restartNotAllowed `decrements during Update. When it hits 0 and a player inputs the Option / whatever button is mapped, they will be process switched to GameOver. So in order to provide the player the ability to be dead and still have access to the menu and prevent game over mode,  I have to block the timer.

`lastAnyKey `could be overwritten instead, but because it's assigned to if player touched input in Update() if not in death screen, we'd have to re-write more of Update() instead of just holding the timer at 1.

Story/StoryHooks.cs
6fears7: Unfortunately, if gameOverMode is not managed here, it creates a few issues. If it is set to false in the ctor, then they will not have the text show up to wait for resuce, shelter etc. They will also lack the power to remove the TextPrompt.

 To that second point, the planned inclusion of [spectating ](https://github.com/henpemaz/Rain-Meadow/pull/210) in the future would benefit from the lack of TextPrompt running as a QoL while watching others. 

PauseMenu_SpawnExitContinueButtons could be used when activated to also remove the gameOverMode so the text prompt is removed, which would only be activated on a user's action versus during Update():

ex:
```
        private void PauseMenu_SpawnExitContinueButtons(On.Menu.PauseMenu.orig_SpawnExitContinueButtons orig, Menu.PauseMenu self)
        {
            self.game.cameras[0].hud.textPrompt.gameOverMode = false;
        }
```

 but then we would be reaching across to another class to alter the variable versus hooking the class directly. 

Let me know what you think.

Chrometurtle: If that's the case I would add in additional modes handled by us. Rather than continuing to leverage gameOverMode. Since technically it's not gameOverMode anymore, it's spectatingMode. This'll give us more granular control on which messages to display

6fears7: Sounds good.

6fears7: I went back and tested this some more:

`if (touchedInput)
                    {
                        self.gameOverMode = false; }` 

if we do not do this, clients cannot access the pause menu when they're dead, which means they cannot quit the game. TextPrompt_Update modes block activation of PauseMenu.

Options:

1. We don't display any TextPrompt for the client on GameOver. Gives them access to PauseMenu to quit.
2. We read-only display TextPrompt, and clients can only exit game after Sleep/Death screen
3. We let clients disable gameOverMode while dead to access PauseMenu. 


I can add a new mode for spectating when that's ready, but for the current problem above, let me know what option you wanna pursue.

Chrometurtle: I was viewing this PR based on the title. The main focus was to prevent players from prematurely ending the cycle and to let the host set that pace.

How we want to handle pausing and text hooks is a much more involved process. Trying to address both of those here will balloon. 

We might not want to actually use the games "pause" menu b/c of game timer behaviors. There's another question on who owns the room after it's previous owner dies. If ownership is maintained, then this will probably further complicate things

6fears7: Fair enough. Will do #2. Just a text prompt update. Host sets pace. We address any future changes in future PR.

PR #211, opened 2024-08-25
modloader view modlist

6fears7:
- [x] Fixed loading new mods
- [x] Fixed unloading old mods
- [x] Fixed initial mod in list activating
- [x] Prompts for user to continue or cancel first


6fears7:
image

PR #212, opened 2024-08-26
Alter arena node logic

6fears7:


6fears7:
Additionally, 

`[Error  :RainMeadow] 21:29:33|410495|RPCManager.Process:Error handing RPC Arena_NextLevelCall System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.NullReferenceException: Object reference not set to an instance of an object`


Host is moving on too quickly before clients can RPC to the next level. Fix that.

PR #213, opened 2024-08-29
Story: allow host to die

youbitchoc:
base logic has been tested to work but more testing is needed to make sure there's no savestate serialization errors

WIN:
host dies > respawn at a winning shelter
players die > respawn with host

LOSE (everyone dies): respawn at original shelters

### DIFF
--- a/Game/RainMeadow.PlayerHooks.cs
+++ b/Game/RainMeadow.PlayerHooks.cs
@@ -16,6 +16,7 @@ public partial class RainMeadow
         On.Player.Grabability += PlayerOnGrabability;
         On.Player.AddFood += Player_AddFood;
         On.Player.AddQuarterFood += Player_AddQuarterFood;
+        On.Player.FoodInRoom_bool += Player_FoodInRoom;
         On.Mushroom.BitByPlayer += Mushroom_BitByPlayer;
         On.KarmaFlower.BitByPlayer += KarmaFlower_BitByPlayer;
         On.PlayerGraphics.DrawSprites += PlayerGraphics_DrawSprites1;
@@ -93,6 +94,16 @@ public partial class RainMeadow
                 OnlineManager.lobby.owner.InvokeRPC(RPCs.AddFood, (short)add);
             }
         }
+    }
+
+    private int Player_FoodInRoom(On.Player.orig_FoodInRoom_bool orig, Player self, bool eatAndDestroy)
+    {
+        if (OnlineManager.lobby != null) {
+            if (self.dead) {
+                return self.FoodInStomach;
+            }
+        }
+        return orig(self, eatAndDestroy);
     }
 
     private void Mushroom_BitByPlayer(On.Mushroom.orig_BitByPlayer orig, Mushroom self, Creature.Grasp grasp, bool eu)
--- a/Game/RainMeadow.PlayerHooks.cs
+++ b/Game/RainMeadow.PlayerHooks.cs
@@ -98,7 +98,7 @@ public partial class RainMeadow
 
     private int Player_FoodInRoom(On.Player.orig_FoodInRoom_bool orig, Player self, bool eatAndDestroy)
     {
-        if (OnlineManager.lobby != null) {
+        if (OnlineManager.lobby != null  && OnlineManager.lobby.gameMode is StoryGameMode) {
             if (self.dead) {
                 return self.FoodInStomach;
             }
--- a/Online/RPCs.cs
+++ b/Online/RPCs.cs
@@ -185,7 +185,8 @@ namespace RainMeadow
                 return;
             }
 
-            if (OnlineManager.lobby.isOwner) {
+            if (OnlineManager.lobby.isOwner)
+            {
                 if (!malnourished && !game.rainWorld.saveBackedUp)
                 {
                     game.rainWorld.saveBackedUp = true;
@@ -198,7 +199,7 @@ namespace RainMeadow
                 var storyClientSettings = OnlineManager.lobby.gameMode.clientSettings as StoryClientSettings;
                 if (storyClientSettings.isDead)
                 {
-                    storyClientSettings.myLastDenPos = null;
+                    storyClientSettings.myLastDenPos = denPos;
                 }
             }
 
--- a/Online/RPCs.cs
+++ b/Online/RPCs.cs
@@ -134,14 +134,7 @@ namespace RainMeadow
             RainMeadow.Debug($"({malnourished}, {denPos})");
 
             var game = (RWCustom.Custom.rainWorld.processManager.currentMainLoop as RainWorldGame);
-            if (game == null || game.manager.upcomingProcess != null)
-            {
-                if (game == null)
-                    RainMeadow.Debug("game is null");
-                else
-                    RainMeadow.Debug($"game has upcoming process {game.manager.upcomingProcess}");
-                return;
-            }
+            if (game == null || game.manager.upcomingProcess != null) return;
 
             if (OnlineManager.lobby.playerAvatars.TryGetValue(OnlineManager.mePlayer, out var playerAvatar))
             {
@@ -158,17 +151,6 @@ namespace RainMeadow
             {
                 game.GetStorySession.saveState.denPosition = denPos;
                 (OnlineManager.lobby.gameMode as StoryGameMode).defaultDenPos = denPos;
-            }
-
-            if (OnlineManager.lobby.isOwner)
-            {
-                if (!malnourished && !game.rainWorld.saveBackedUp)
-                {
-                    game.rainWorld.saveBackedUp = true;
-                    game.rainWorld.progression.BackUpSave("_Backup");
-                }
-                // TODO: handle null realized players (removed from game e.g. fall out of world / creature den)
-                game.GetStorySession.saveState.SessionEnded(game, true, malnourished);
             }
 
             foreach (OnlinePlayer player in OnlineManager.players)
@@ -198,7 +180,15 @@ namespace RainMeadow
                 return;
             }
 
-            if (!OnlineManager.lobby.isOwner)
+            if (OnlineManager.lobby.isOwner)
+            {
+                if (!malnourished && !game.rainWorld.saveBackedUp)
+                {
+                    game.rainWorld.saveBackedUp = true;
+                    game.rainWorld.progression.BackUpSave("_Backup");
+                }
+            }
+            else
             {
                 var storyClientSettings = OnlineManager.lobby.gameMode.clientSettings as StoryClientSettings;
                 if (storyClientSettings.isDead)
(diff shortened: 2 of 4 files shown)


Game/RainMeadow.PlayerHooks.cs
```diff
@@ -95,6 +96,16 @@ private void Player_AddFood(On.Player.orig_AddFood orig, Player self, int add)
         }
     }
 
+    private int Player_FoodInRoom(On.Player.orig_FoodInRoom_bool orig, Player self, bool eatAndDestroy)
+    {
+        if (OnlineManager.lobby != null) {
```
Chrometurtle, line 101: Check for storymode only. I don't think it'd effect other things, but just to be safe

Online/RPCs.cs
```diff
@@ -121,49 +121,86 @@ public static void GoToDeathScreen()
...
             }
 
-            //TODO: Having soft-win on makes this very difficult. For now I'm (Turtle) locking it down so that you can only win if the host is alive
             if (OnlineManager.lobby.isOwner) {
+                if (!malnourished && !game.rainWorld.saveBackedUp)
+                {
+                    game.rainWorld.saveBackedUp = true;
+                    game.rainWorld.progression.BackUpSave("_Backup");
+                }
                 game.GetStorySession.saveState.SessionEnded(game, true, malnourished);
```
Chrometurtle, line 194: So this guy is the thing you'll want to "sync". The food pips graphical error occurs because client's don't run this segment of code. So when a client does "go to win screen" it doesn't do the whole cycle end properly

youbitchoc: can we get away with hooking the save functions to not write to disk?

Chrometurtle: Short answer, no. 

Longer answer, the issue comes down to a null check the game itself makes. Local co-op gets around this by making all players have to hibernate in the same area. So they can just assume the hibernation shelter is X and continue with the saving logic

Online/RPCs.cs
```diff
@@ -121,49 +121,83 @@ public static void GoToDeathScreen()
                 game.GoToRedsGameOver();
                 return;
             }
-            game.GetStorySession.saveState.SessionEnded(game, false, false);
+            if (OnlineManager.lobby.isOwner) {
+                game.GetStorySession.saveState.SessionEnded(game, false, false);
```
Chrometurtle, line 125: Add a todo to investigate client side savestate syncing

Online/RPCs.cs
```diff
@@ -121,49 +121,83 @@ public static void GoToDeathScreen()
...
-            var denPos = (OnlineManager.lobby.gameMode as StoryGameMode).defaultDenPos;
-            if (denPos == null) {
-                AbstractCreature firstAlivePlayer = game.FirstAlivePlayer;
-                denPos = game.world.GetAbstractRoom(firstAlivePlayer.pos).name;
+            else
+            {
+                var storyClientSettings = OnlineManager.lobby.gameMode.clientSettings as StoryClientSettings;
+                if (storyClientSettings.isDead)
+                {
+                    storyClientSettings.myLastDenPos = denPos;
```
Chrometurtle, line 196: Might be mistaken, but are we moving dead players to the hosts hibernation den?

Story/RainMeadow.StoryHooks.cs
```diff
@@ -326,16 +329,28 @@ private void RainWorldGame_GoToDeathScreen(On.RainWorldGame.orig_GoToDeathScreen
...
-                if (!OnlineManager.lobby.isOwner)
+                foreach (var playerAvatar in OnlineManager.lobby.playerAvatars.Values)
                 {
-                    if (!OnlineManager.lobby.owner.OutgoingEvents.Any(e => e is RPCEvent rpc && rpc.IsIdentical(RPCs.MovePlayersToDeathScreen))) {
-                        OnlineManager.lobby.owner.InvokeRPC(RPCs.MovePlayersToDeathScreen);
+                    if (playerAvatar.type == (byte)OnlineEntity.EntityId.IdType.none) continue; // not in game
+                    if (playerAvatar.FindEntity(true) is OnlinePhysicalObject opo && opo.apo is AbstractCreature ac)
+                    {
+                        if (ac.state.alive)
+                            RainMeadow.Debug($"still alive: {playerAvatar}");
```
Chrometurtle, line 338: Do we still need these debug statements?

Story/RainMeadow.StoryHooks.cs
```diff
@@ -326,16 +329,28 @@ private void RainWorldGame_GoToDeathScreen(On.RainWorldGame.orig_GoToDeathScreen
...
                     }
                 }
-                else
+
+                if (OnlineManager.lobby.isOwner)
```
Chrometurtle, line 342: Wouldn't be bad to check for outgoing rpc, you can prob ignore this comment though because I added the check to prove that this wasn't causing the starving issue

Story/RainMeadow.StoryHooks.cs
Chrometurtle, line 1: Out of curiosity, do we know how the game behaves if the dead slugcats body is stored in the host shelter? Or even a different shelter?

Story/RainMeadow.StoryHooks.cs
```diff
@@ -404,25 +407,44 @@ private SaveState PlayerProgression_GetOrInitiateSaveState(On.PlayerProgression.
...
                 var storyClientSettings = gameMode.clientSettings as StoryClientSettings;
-                if (storyClientSettings.myLastDenPos != null)
+                RainMeadow.Debug($"origDenPos: {origSaveState.denPosition}");
+                RainMeadow.Debug($"lastDenPos: {storyClientSettings.myLastDenPos}");
+                RainMeadow.Debug($"defaultDenPos: {gameMode.defaultDenPos}");
+
+                if (OnlineManager.lobby.isOwner)
                 {
-                    origSaveState.denPosition = storyClientSettings.myLastDenPos;
+                    RainMeadow.Debug($"setting defaultDenPos to {origSaveState.denPosition}");
```
Chrometurtle, line 416: Cleanup if you think these debug statements aren't needed anymore

Chrometurtle (commented):
I'll need to look into the bringuptodate stuff, but other than some small nitpicks lgtm

Online/RPCs.cs
Chrometurtle: Might not need this, didn't see the save to disk stuff when I made the comment

Online/RPCs.cs
youbitchoc: yup, host passes their den position thru GoToWinScreen

Story/RainMeadow.StoryHooks.cs
youbitchoc: probably not

during playtesting there was a really weird situation where host died and clients sheltered but something in SessionEnded failed (seems like it couldn't find the host's realized creature?) so it went on to call GoToDeathScreen instead

haven't managed to reproduce even with getting snagged by wormgrass and falling out of the world but i suppose this debug statement wouldn't help much anyways

Chrometurtle: Yeah that's the one thing I was afraid of. I'm pretty sure it's possible for dead players realized creatures are deleted

Story/RainMeadow.StoryHooks.cs
youbitchoc: oh i changed it such that clients can't invoke `MovePlayersToDeathScreen` on the host, only the host can trigger it themselves

will likely have to be changed anyways for the spectator mode PR

Chrometurtle: Right, it should be impossible for the host to call moveplayerstodeathscreen multiple times now. So you can safely ignore the check

Online/RPCs.cs
youbitchoc: having clients save to memory still doesn't fix stomach item desync so i think we have to do more

Story/RainMeadow.StoryHooks.cs
youbitchoc: currently all dead slugcats just respawn with the host, but i like the idea of stealing dead slugcats to respawn with

6fears7: Dragging slugs is 'technically' possible with effort, but we'll need to rework grasp logic so that their weight is normalized and not tugging against the player. 

All that to say, when you test, stay near the shelter.

Online/RPCs.cs
```diff
@@ -122,48 +122,81 @@ public static void GoToDeathScreen()
...
+                }
             }
 
-            //TODO: Having soft-win on makes this very difficult. For now I'm (Turtle) locking it down so that you can only win if the host is alive
-            if (OnlineManager.lobby.isOwner) {
-                game.GetStorySession.saveState.SessionEnded(game, true, malnourished);
-            }
+            // TODO: investigate client save desync (e.g. swallowed items)
+            game.GetStorySession.saveState.SessionEnded(game, true, malnourished);
```
Chrometurtle, line 199: I'm not sure how this isn't crashing. Line 196 in Savestate. If a player is dead and unrealized, this should crash. and a dead player can be unrealized. 

PlayerSessionRecord needs to be synced as well.

Might need to just hook into sessionEnded and redo the logic to work with online versions of things

Chrometurtle (changes requested):

Game/RainMeadow.PlayerHooks.cs
```diff
@@ -95,6 +96,16 @@ private void Player_AddFood(On.Player.orig_AddFood orig, Player self, int add)
         }
     }
 
+    private int Player_FoodInRoom(On.Player.orig_FoodInRoom_bool orig, Player self, bool eatAndDestroy)
+    {
+        if (OnlineManager.lobby != null  && OnlineManager.lobby.gameMode is StoryGameMode) {
```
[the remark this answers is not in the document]
youbitchoc: if the host is dead, they probably aren't in a shelter

ideally we should do this for every player but im not sure how to do that

Story/StoryHooks.cs
```diff
@@ -447,25 +431,40 @@ private SaveState PlayerProgression_GetOrInitiateSaveState(On.PlayerProgression.
...
+        {
+            if (OnlineManager.lobby != null && !OnlineManager.lobby.isOwner) return false;
+            return orig(self, saveCurrentState, saveMaps, saveMiscProg);
+        }
+
+        private void SaveState_BringUpToDate(On.SaveState.orig_BringUpToDate orig, SaveState self, RainWorldGame game)
+        {
+            if (isStoryMode(out var gameMode)) {
+                var denPos = self.denPosition;
+                orig(self, game);
```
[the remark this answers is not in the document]
youbitchoc: this fixes den position being set to the room that the host died in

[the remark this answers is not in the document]
Chrometurtle: Ah I misread, big dumb

Online/RPCs.cs
youbitchoc: can we just remove unrealized players from the player list? not sure how this works though

Chrometurtle: add a null check to the realized player

chockery: done in 19f7c0c1fd0ef9f64a18069bc2795d243dae9e38

youbitchoc: done in 19f7c0c1fd0ef9f64a18069bc2795d243dae9e38

Chrometurtle (approved):

PR #218, opened 2024-09-01
story mode: save slugcat custom colour (in remix settings)

youbitchoc:
unsure if we should be calling `_SaveConfigFile()` directly

forgive the misc style cleanups, i can undo if needed


henpemaz:
Good stuff

PR #221, opened 2024-09-02
Lobby Menu fixes

SaddestUnderscore:
- Fix on a chance of having 3 digits of max players when you hold any number
- Fix on password popup being slightly off center on different aspect ratios


6fears7:
100+ players in Arena let's go!

PR #222, opened 2024-09-02
story mode: client continue button to wait for host

youbitchoc:
allows clients to press continue before the host is ready, then auto-join once the host has continued

would like the client's continue button to be toggleable (and also renamed to ready?) but not sure how to do that ui-wise

UPDATE: implemented both

[video demo here](https://discord.com/channels/1094716194180841602/1094730322911952917/1280088742530387969)


henpemaz:
Good stuff!

PR #225, opened 2024-09-03
story mode: room settings set current campaign

youbitchoc:
sets the campaign context in which rooms are loaded

fixes connections between rooms as well as unexpected rot, seems to fix scav toll spawns, may fix other creature spawns too

should this go in `Story/StoryHooks`?

### DIFF
--- a/Game/RainMeadow.GameHooks.cs
+++ b/Game/RainMeadow.GameHooks.cs
@@ -30,6 +30,8 @@ namespace RainMeadow
             On.Room.Loaded += Room_LoadedCheck;
             On.Room.PlaceQuantifiedCreaturesInRoom += Room_PlaceQuantifiedCreaturesInRoom;
 
+            On.RoomSettings.ctor += RoomSettings_ctor;
+
             On.FliesWorldAI.AddFlyToSwarmRoom += FliesWorldAI_AddFlyToSwarmRoom;
 
             // can't pause it's online mom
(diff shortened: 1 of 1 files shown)


youbitchoc:
room [SS_I02](https://henpemaz.github.io/Rain-World-Interactive-Map/map.html?region=SS&room=SS_I02) as survivor

BEFORE: seems to load monk's map (pipe blocked by rot)
image

AFTER: loads survivor's map (pipe is clear)
image

Game/RainMeadow.GameHooks.cs
```diff
@@ -180,6 +182,16 @@ private void Room_LoadedCheck(On.Room.orig_Loaded orig, Room self)
             }
         }
 
+        private void RoomSettings_ctor(On.RoomSettings.orig_ctor orig, RoomSettings self, string name, Region region, bool template, bool firstTemplate, SlugcatStats.Name playerChar)
+        {
+            if (isStoryMode(out var storyGameMode))
+            {
+                playerChar = storyGameMode.currentCampaign;
```
6fears7, line 189: In case Turtle wants also wants to take a look at this, I'll spare him some decompiling time:

```
        if (template)
        {
            filePath = AssetManager.ResolveFilePath("World" + Path.DirectorySeparatorChar + region.name + Path.DirectorySeparatorChar + name + ".txt");
        }
        else if (!File.Exists(filePath = ((playerChar == null) ? "" : WorldLoader.FindRoomFile(name, includeRootDirectory: false, "_settings-" + playerChar.value + ".txt"))))
        {
            string path = ((!ModManager.MSC || !name.EndsWith("-2")) ? WorldLoader.FindRoomFile(name, includeRootDirectory: false, "_settings.txt") : WorldLoader.FindRoomFile(name.Substring(0, name.Length - 2), includeRootDirectory: false, "-2_settings.txt"));
            if (File.Exists(path))
            {
                filePath = path;
            }
            else if (name.EndsWith("-2"))
            {
                filePath = WorldLoader.FindRoomFile(name.Substring(0, name.Length - 2), includeRootDirectory: false, "_settings.txt");
            }
            else
            {
                filePath = path;
            }
        }
```

Also of interest, this class uses Save() and Load() methods that take in a filepath that's assigned during the ctor. 

@youbitchoc the screenshot looks good. If you can verify that on death / sleep the world is loaded correctly that would be great.

6fears7: as for placement of code, GameHooks would be more appropriate imo.

Chrometurtle (approved):

Game/RainMeadow.GameHooks.cs
youbitchoc: tested to work across cycles and deaths as well as having clients load the room

Issue #195, opened 2024-08-18
Story / Meadow: Gravity


6fears7:
Players traversing across rooms with transitional gravity (high -> low, low-> high) had matching level of gravity applied instead of having that specific room's gravity. 

Players reported the client's room appeared to dictate the gravity for both players (2 player mode).

Additionally, gravity on/off timer is not synced.

- [ ] Correct room gravity to player 
- [ ] Sync gravity timers like in 5P

henpemaz:
the default behavior of the object that handles the gravity in those special rooms is to update "the entire rooms gravity" based on where the player 1 is, works well enough in singleplayer. There might be some special handling built into jollycoop, there was also special handling iirc in Monkland that we could reference

youbitchoc:
for the gravity timers, would it be preferable to have the room owner sync room gravity constantly, or should clients calculate it on their end?

im kinda leaning towards the latter because it can maintain consistency when room owners change

youbitchoc:
oops

antigravity in rooms is still broken

6fears7:
is the I/O gravity timer data specific to a room? That, in my mind, would be the deciding course to take for ownership.

youbitchoc:
yup, AntiGravity is per room

the way the game calculates this is very silly, i think we can get by with syncing just `bool on` and `int counter` then letting clients do the calculations on their end (cyclemin/max are initialized once on room load so we dont have to worry about syncing those)

for MSC support we'll need another bool to handle whatever's going on in RM

youbitchoc:
oh HECK it's not, it's a "global" in RainCycle
the code in AntiGravity ctor just initializes it if it's null

PR #227, opened 2024-09-04
fix per-player gradient gravity in SS_E08

youbitchoc:
sets it clientside based on the player position

should consider using IL hook to add on to the `is Player` check, i don't feel comfortable duplicating the gravity calculation

addresses transitional gravity in #195


youbitchoc:
managed to convert it to IL

youbitchoc:
tested to work

PR #229, opened 2024-09-04
correct game mode checks for arena

6fears7:


6fears7:
fixes online arena logic bleeding into local multiplayer

PR #234, opened 2024-09-09
Story: use campaign savestatenumber, intercept in save/load

youbitchoc:
game uses savestatenum to check player character, so only use custom meadow onlineplayer enum when loading/saving

- [ ] make wrapping less ugly
- [ ] find a better home for GetSaveStateNumber (maybe as a new func to OnlineGameMode? or is this too story-specific)


youbitchoc:
closing in favour of #240

PR #235, opened 2024-09-09
spectator mode

6fears7:


6fears7:
- [x] Fix bad rebase
- [ ] Handle weird map bug

PR #236, opened 2024-09-10
update string, update sleepy scugs

6fears7:


Story/StoryMenu.cs
```diff
@@ -398,7 +398,7 @@ private void CustomSlugcatSetup()
                 var btn = new SimplerButton(this, mainPage, SlugcatStats.getSlugcatName(slug), new Vector2(394, 515) - i * new Vector2(0, 38), new Vector2(110, 30));
                 btn.toggled = false;
                 mainPage.subObjects.Add(btn);
-
+                customSelectedSlugcat = slugList[0]; // default as Survivor so no sleep walking scugs
```
youbitchoc, line 401: must this be in the loop? could we set the default value at initialization?

youbitchoc (changes requested):

Story/StoryMenu.cs
6fears7: Sure could. Won't be back at it for another 36 hours or so, so if you wanna add those changes feel free. Else I'll get it then

youbitchoc: not too happy with hardcoding it to White but the lobby menu relies on this anyway (arena hardcodes it too) so it should be fine

youbitchoc:
lgtm
(there's some extra newlines here and there but am just being pedantic)

PR #247, opened 2024-09-14
sync abstractspear state

youbitchoc:
should fix spear lifetime for good

TODO: figure out how to hook this up to the other stuff


youbitchoc:
spear stuckInWall and vertical/horizontal are now synced, spearmaster needles still appear as spears but i don't think it's urgent

6fears7:
LGTM

PR #249, opened 2024-09-15
Story: refactor denpos logic

youbitchoc:
don't search for players, use denPos from ShelterDoor.Close

### DIFF
--- a/Game/RainMeadow.GameplayHooks.cs
+++ b/Game/RainMeadow.GameplayHooks.cs
@@ -226,7 +226,8 @@ namespace RainMeadow
                 return;
             }
 
-            if (OnlineManager.lobby.gameMode is StoryGameMode storyGameMode)
+            var storyGameMode = OnlineManager.lobby.gameMode as StoryGameMode;
+            if (storyGameMode != null)
             {
                 var playerIDs = OnlineManager.lobby.participants.Select(p => p.inLobbyId).ToList();
                 var readyWinPlayers = storyGameMode.readyForWinPlayers.ToList();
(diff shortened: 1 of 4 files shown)


youbitchoc:
can we trust that `!isDead` means host has successfully sheltered? if so then hostDenPos can be refactored out

6fears7:
Assuming we eventually fix grasps, what about if players are dragged to shelters?

youbitchoc:
dead scugs in shelters should still call ShelterDoor.Close, I'll rework the logic to not rollback if the player is in a shelter

youbitchoc:
done, also fixed myLastDenPos being set if the player enters a broken shelter

henpemaz:
> Assuming we eventually fix grasps, what about if players are dragged to shelters?

Grasps should be working already on realized AND abstract (stuckobjects). Player.dangergrasp field isn't synced but for realized players should likely have the right value.

Game/RainMeadow.GameplayHooks.cs
```diff
@@ -251,6 +245,14 @@ private void ShelterDoorOnClose(On.ShelterDoor.orig_Close orig, ShelterDoor self
...
             }
             orig(self);
+            if (!self.IsClosing) return;
+
+            if (storyGameMode != null)
+            {
+                var storyClientSettings = storyGameMode.clientSettings as StoryClientSettings;
+                storyClientSettings.myLastDenPos = self.room.abstractRoom.name;
+                storyClientSettings.hasSheltered = true;
+            }
```
henpemaz, line 255: I'll always frown a little bit at early-returns outside of a single-responsibility method. Makes mistakes harder to catch. 

Need another couple lines on the same hook? Didn't spot the early return? Blam.

6fears7:
By "grasps working" I really mean the gameplay effect equivalent to ownership being transferred where bodies are no longer resisting being dragged from their current pos. 

One easy test case for this impact is removing the sSpawning static Boolean and moving a dead slug. When the ownership is transferable the slug moves with ease.

Game/RainMeadow.GameplayHooks.cs
youbitchoc: hm, I decided to use the early return because I don't think we'd want to do anything if the shelter is broken

6fears7: If we really want to check if the door is closing without using an early return, could we just wrap 249 and below with:

`if (self.IsClosing)`?

youbitchoc: yup, done

PR #252, opened 2024-09-17
Add pipe color

6fears7:


Game/RainMeadow.PlayerHooks.cs
youbitchoc, line 1: could we refactor out the extra orig(self) calls?

youbitchoc (changes requested):

Game/RainMeadow.PlayerHooks.cs
youbitchoc: aight edited it on mobile, do undo if broken

6fears7:
Tested all game modes, works.

PR #253, opened 2024-09-18
handle player stomach items

youbitchoc:
needs more testing

- [ ] when player swallows item, swallow seems to fail for a moment before swallowed state is synced
  - can handle by faking the swallow clientside (~~too much hassle imo~~ EDIT: actually pretty noticeable, might be worth fixing)
- [ ] no swallow sound
  - play sound manually (ugly)
  - IL hook swallow function to skip item stuff and play sound (worth the hassle?)
- [ ] sheltering? dying?
  - should respawn in shelter then players can "claim" them while spawning in
- [ ] remix item tracking


youbitchoc:
closing in favour of #263

PR #254, opened 2024-09-18
Story: fix client slugcat classes

youbitchoc:
custom slugcats per client yippee


6fears7:
Curious. Were we having cases of null refs? (MSC?)

6fears7:
image

Works.

PR #256, opened 2024-09-18
updated spectator update to check input in Draw()

6fears7:


6fears7:
Improves get keydown impact

PR #263, opened 2024-09-21
sync stomach objects

youbitchoc:
- [X] test sheltering and dying
  - items carry over cycles
- [ ] fix host stomach item duping
- [ ] fix client stomach items spawning outside stomach
  - players should "claim" them while spawning in
- [x] fix swallowed creatures realizing out of stomach when entering room

~~- [ ] test remix item tracking~~ not too urgent imo


Online/State/PhysicalObjectEntityState.cs
```diff
@@ -65,6 +65,12 @@ public override void ReadTo(OnlineEntity onlineEntity)
             var apo = onlineObject.apo;
             RainMeadow.Trace($"{onlineEntity} received realized state? {realizedObjectState != null} entity realized?{onlineObject.realized}");
 
+            if (!realized && !onlineObject.realized && !onlineEntity.isTransferable)
```
youbitchoc, line 68: this assumes that if an entity is locked to one owner and it isn't realized in the owner's world then other players shouldn't be able to realize it

less hacky solution would be to have a `shouldRealize` bool then guard all the joinimpl/realize stuff

henpemaz: This is a bit pointless because once an item is swallowed, the game stops updating it and moving it, so we'll receive a non-moving position anyways.

youbitchoc: without this check, clients will try to realize it in the room where it was originally swallowed

Game/RainMeadow.PlayerHooks.cs
```diff
@@ -301,6 +304,65 @@ private Player.ObjectGrabability PlayerOnGrabability(On.Player.orig_Grabability
...
+                i => i.MatchStfld<Player>("objectInStomach")
+                );
+            c.Emit(OpCodes.Ldarg_0);
+            c.EmitDelegate((Player self) =>
+            {
+                if (OnlineManager.lobby != null && OnlinePhysicalObject.map.TryGetValue(self.abstractPhysicalObject, out var oe))
+                {
+                    if (!oe.isMine) return false;
+                    if (OnlinePhysicalObject.map.TryGetValue(self.objectInStomach, out var oeInStomach))
+                        oeInStomach.isTransferable = false; // don't release ownership
```
henpemaz, line 325: This field of OnlineEntity isn't synced, the player that sets it will have a unique perspective about it.

Might still work for the sake of controlling whether to release or to accept requests. However "region owner" might still transfer it if asked directly. Clients ask the region owner if the entity is in a resource the direct owner isn't it (assumed abandoned).

Online/State/PhysicalObjectEntityState.cs
henpemaz: I think they still will if the entity is already found there in abstractroom on room-load, regardless of the move() call here

henpemaz: Realize will only be called if ChangeRooms is called, which will only be called if pos1.room != pos2.room

If we are feeding the same position after the removal, which we should, this here isn't what's causing it to realize.

Game/RainMeadow.PlayerHooks.cs
youbitchoc: huuh.
should we want to sync isTransferable too? or use that shouldRealize bool
(I'm not too happy with the original "solution" of realizing first then checking if it's a stomach object)

youbitchoc:
closing in favour of #272

PR #264, opened 2024-09-21
Arena fix classes and msc levels

6fears7:
Now to determine where I introduced the saving bug


6fears7:
Introduced the bug at my first commit: fc69bc2fa8bcf70d53a8ef6cb1950548b160db0a

PR #266, opened 2024-09-21
UI add initial flash for player arrow

6fears7:
shows your username and arrow on initial spawn briefly to help player orient


6fears7:
works in story, still need to determine how to do arena

6fears7:
This int is from MSC. Need to find something else to capture time since spawn

6fears7:
Can count back from cycle timer for Story, will investigate Arena.

6fears7:
- [ ] Test story
- [ ] Test arena
- [ ] Test gates
- [ ] Test next cycle

PR #270, opened 2024-09-22
Story: fix start cutscenes

youbitchoc:
fixes hunter host not running start script on respawn
fixes spearmaster client spawning in wrong region


henpemaz:
Doesn't this cause the fruits on survivor/monk to dupe due to the room fully loading on each client?

youbitchoc:
no, it only affects roomspecificscripts, not actual room loading

youbitchoc:
first client in arti campaign can control fine, beyond that seems to be unable to move?

youbitchoc:
fixed

PR #273, opened 2024-09-24
Banhammer

6fears7:
- [x] Update HUD when user is kicked / removed during Update()
- [x] Add MeadowPlayerID to lobby list
- [x] Actually prevent client from joining


6fears7:
Just need to figure out why the list is not updating for clients in the ReadTo then we'll be in business

6fears7:
had henpe look over in disc.

PR #275, opened 2024-09-25
sync gravity cycle

youbitchoc:
fixes gravity cycle desync in 5P unfortunate development (SS_S03 is a good shelter for testing)

on worldsession ownership transfer, counter before next change will be initialized to minimum value (320 by default). we could handle this more correctly i.e. randomize the duration, but imo it's not too important

tested:
- [x] story
- [x] meadow


youbitchoc:
closes #195

PR #276, opened 2024-09-26
don't set abscreature destination if unrealized

youbitchoc:
only set destination if creature is realized

minimizes creatures trying (and failing) to pathfind to rooms that are not loaded on the client


youbitchoc:
some wacky creature ai, might be due to this? not sure, gotta test

youbitchoc:
needs re-investigation, i remember lagspikes when entering moon's superstructure

PR #278, opened 2024-09-26
simpler one-off RPCs with InvokeOnceRPC

youbitchoc:
slower than directly calling the original function but i think the simpler code is worth it

also fixed mushroom consumption multiplying per player

a note about `explodeSporePuff`: ideally we should be ignoring the colour when checking for outgoing but the original version didn't work anyway (comparison didn't provide colour params so length diff => counted as not equal) and colour shouldn't change so it should still work fine


6fears7:
This only impacts story and arena. Lgtm

PR #279, opened 2024-09-26
sync visited echoes

youbitchoc:
- [x] client doesn't always see echoes but host can (even if spectating)

### DIFF
--- a/GameModes/StoryGameMode.cs
+++ b/GameModes/StoryGameMode.cs
@@ -13,6 +13,7 @@ namespace RainMeadow
         public bool didStartCycle = false;
         public string? defaultDenPos;
         public SlugcatStats.Name currentCampaign;
+        public Dictionary<string, int> ghostsTalkedTo;
         public Dictionary<string, bool> storyBoolRemixSettings;
         public Dictionary<string, float> storyFloatRemixSettings;
         public Dictionary<string, int> storyIntRemixSettings;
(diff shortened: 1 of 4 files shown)


GameModes/StoryGameMode.cs
```diff
@@ -97,6 +98,23 @@ internal override void ResourceAvailable(OnlineResource onlineResource)
...
 
+        internal override void ResourceActive(OnlineResource onlineResource)
+        {
+            base.ResourceActive(onlineResource);
+            if (onlineResource is WorldSession ws)
+            {
+                var regionState = ws.world.regionState;
+                if (this.lobby.isOwner)
+                {
+                    ghostsTalkedTo = regionState.saveState.deathPersistentSaveData.ghostsTalkedTo.ToDictionary(kvp => kvp.Key.value, kvp => kvp.Value);
```
6fears7, line 109: Still want the debug here or was this just for testing?

youbitchoc: oops, forgot to remove

Story/StoryHooks.cs
```diff
@@ -809,5 +819,16 @@ private bool PlayersStandingStill(On.RegionGate.orig_PlayersStandingStill orig,
...
             return orig(self);
         }
+
+        private void GhostHunch_Update(On.GhostHunch.orig_Update orig, GhostHunch self, bool eu)
+        {
+            orig(self, eu);
+            if (OnlineManager.lobby == null) return;
+            if (!OnlineManager.lobby.isOwner && OnlineManager.lobby.gameMode is StoryGameMode)
+            {
+                if (self.ghostNumber != null && (self.room.game.session as StoryGameSession).saveState.deathPersistentSaveData.ghostsTalkedTo[self.ghostNumber] > 0)
```
6fears7, line 829: Just curious: Is this method ran every frame in the game and this is trying to optimize so it only RPCs when a change occurs or is there another reason it's set to > 0 to prevent early RPC in the interaction?

Story/StoryHooks.cs
```diff
@@ -809,5 +819,16 @@ private bool PlayersStandingStill(On.RegionGate.orig_PlayersStandingStill orig,
             }
             return orig(self);
         }
+
+        private void GhostHunch_Update(On.GhostHunch.orig_Update orig, GhostHunch self, bool eu)
+        {
+            orig(self, eu);
+            if (OnlineManager.lobby == null) return;
```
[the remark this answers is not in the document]
youbitchoc: ah shoot yea imma fix that

Story/StoryHooks.cs
youbitchoc: yup it runs every frame so this just waits until the echo has been activated before sending the rpc

PR #282, opened 2024-09-26
cleanup scav bomb code, sync singularity bombs

youbitchoc:
depends on #278

needs testing


youbitchoc:
lightning still persists sometimes

6fears7:
I have some concerns on the isMine check for objectHit. I don't think we can guarantee that all objects hit will belong to us.

youbitchoc:
yup reverted, wasn't the cause of spontaneous combustion (fixed by #290)

PR #285, opened 2024-09-26
added sync for sing bomb

6fears7:


6fears7:
might need more just fyi

youbitchoc:
thanks, incorporated into #282

PR #289, opened 2024-09-28
Story: sync savestate string

youbitchoc:
fixes echoes and consumables

savestate string is ~15kb, can be compressed to ~3.7kb (3.4kb if we convert separators to ascii control chars)
only has to be synced on worldsession join

~~may cause duped creature spawns? though this seems to be a symptom of a bigger issue since clients shouldn't be allowed to spawn creatures in the first place~~ _EDIT: no longer seems to be an issue_

explicitly syncing all the relevant data (consumables, echoes, passages?, no need for creature spawns) thru onlinefields would be more efficient, did some progress in #267 (consumables) and #279 (echoes WIP)

but imo this is safer

### DIFF
--- a/GameModes/StoryGameMode.cs
+++ b/GameModes/StoryGameMode.cs
@@ -14,8 +14,6 @@ namespace RainMeadow
         public string? defaultDenPos;
         public string? region = null;
         public SlugcatStats.Name currentCampaign;
-        public Dictionary<string, int> ghostsTalkedTo;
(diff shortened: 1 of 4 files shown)


6fears7:
Can this be closed or still pondering?

youbitchoc:
echo visits didn't call SessionEnded, fixed by syncing later at GetOrInitiateSaveState

youbitchoc:
need a playtest

youbitchoc:
eh

PR #290, opened 2024-09-29
disable throwing for remote players

youbitchoc:
fixes grenades and singularity bombs in offhand getting triggered by clients and going off

player throws item (with bomb in offhand)
-> throw input pressed -(1)
-> item is out of hand -(2)
if client sees (2) before (1) they will think player is throwing with their offhand (which is the bomb)

### DIFF
--- a/Game/RainMeadow.PlayerHooks.cs
+++ b/Game/RainMeadow.PlayerHooks.cs
@@ -21,6 +21,7 @@ public partial class RainMeadow
         On.Player.SwallowObject += Player_SwallowObject;
         On.Player.Regurgitate += Player_Regurgitate;
         On.Player.SpitUpCraftedObject += Player_SpitUpCraftedObject;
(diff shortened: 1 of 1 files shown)


6fears7:
An Aggressive hook, so I have a few questions:

- [ ] While it shouldn't, does this impact animations we see of other slugs throwing things?

- [ ] RealizedWeaponState (bomb, sing bomb, spear, etc..) should just be broadcasting their properties we listed for our sync to occur. We shouldn't lose functionally on that by disabling remote throws, but would like to confirm throwable behavior (and if necessary, request transfers like on spear attacks) function as well. 

I will admit this is a bit of a weird fix for this, but I also can't think of real negatives from this approach.

youbitchoc:
tested online in artificer campaign
throwing animation looks fine, spears and rocks work ok too

6fears7:
amazing

PR #294, opened 2024-09-30
sync weapon thrownBy

youbitchoc:
don't sync tailPos and lastRotation (calculated clientside)

will need to test if this affects creature relations

### DIFF
--- a/Online/State/RealizedWeaponState.cs
+++ b/Online/State/RealizedWeaponState.cs
@@ -9,27 +9,24 @@ namespace RainMeadow
         // is all this data really necessary?
         [OnlineField]
         protected byte mode;
-        [OnlineField]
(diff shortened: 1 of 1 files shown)


youbitchoc:
this should fix friendly fire

6fears7:
Tested creature relations. Does not address changes alone. Might help with the hook we need though.

youbitchoc:
do we need to sync HitSomethingWithoutStopping for that? or should we sync killtags

youbitchoc:
i think we can deal with creature relations in another pr, this one is more for fixing friendly fire

6fears7:
oh yeah creature relations is gonna be a bear. Outside this PR scope

LGTM

PR #301, opened 2024-10-01
check if WorldSession owned, drop changedRegions

youbitchoc:
more robust fix for clients joining in the wrong region

needs testing


youbitchoc:
game logic for determining spawn region is very ugly and very hard to reuse for this (it loads the world)

6fears7:
Think this has been addressed in a new build but will wait  for confirm.

youbitchoc:
? which commit/pr?

6fears7:
Could've sworn you dropped a build for enabling clients to join after crossing karma gate and then denning.

youbitchoc:
better solution might be to spawn the player in then abort if the worldsession is different

PR #306, opened 2024-10-02
fix clients calling out hits

youbitchoc:
should fix explosives getting multi-exploded

### DIFF
--- a/Game/RainMeadow.GameplayHooks.cs
+++ b/Game/RainMeadow.GameplayHooks.cs
@@ -187,7 +187,8 @@ namespace RainMeadow
             if (!room.isOwner)
             {
                 OnlinePhysicalObject.map.TryGetValue(self.abstractPhysicalObject, out var objectHit);
-                if (objectHit != null)
+                OnlinePhysicalObject.map.TryGetValue(explosion?.sourceObject.abstractPhysicalObject, out var explosionSource);
+                if (objectHit != null && (objectHit.isMine || (explosionSource != null && explosionSource.isMine)))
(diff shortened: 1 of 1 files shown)


6fears7 (commented):
1. me  owner has a bomb
2. me throw it
3. me hit an object that is not mine
4. if the source (bomb) is mine, notify room owner.

Order looks right, but have some concerns:

in Violence:

```
if (suspect is Explosion explosion) trueVillain = explosion.sourceObject;
                else if (suspect is PhysicalObject villainObject) trueVillain = villainObject;
                if (trueVillain != null)
                {
                    if (!OnlinePhysicalObject.map.TryGetValue(trueVillain.abstractPhysicalObject, out var onlineTrueVillain))
                    {
                        if (trueVillain.abstractPhysicalObject.type == AbstractPhysicalObject.AbstractObjectType.ScavengerBomb
                            || trueVillain.abstractPhysicalObject.type == MoreSlugcats.MoreSlugcatsEnums.AbstractObjectType.SingularityBomb)
                        {
                            // bombs exit quickly, and that's ok.
                            OnlinePhysicalObject onlineVillain = null;
                            (onlineVictim as OnlineCreature).RPCCreatureViolence(onlineVillain, hitchunk?.index, hitappendage, directionandmomentum, type, damage, stunbonus);
                            return;
                        }
```
we might have issues where the explosion source owner is null because the resource exited.

youbitchoc:
the object that was hit can still call it out, or the room owner could also just simulate it on their end

PR #313, opened 2024-10-03
handle duplicate room names

youbitchoc:
appends a `.$number` suffix to duplicate room names in worldsession dictionary (could this clash with modded regions?)

fixes red room bug in gilded sanctuary, may fix outer expanse too

we can't skip the rooms because they're still counted in World

```C#
for (var i = 0; roomSessions.Keys.Contains((name = room.name + "." + i)); i++);
```
not too proud of this code, feel free to refactor it

### DIFF
--- a/Online/Resource/WorldSession.cs
+++ b/Online/Resource/WorldSession.cs
@@ -36,7 +36,18 @@ namespace RainMeadow
             foreach (var room in world.abstractRooms)
             {
                 var rs = new RoomSession(this, room);
-                roomSessions.Add(room.name, rs);
+                try
+                {
+                    roomSessions.Add(room.name, rs);
+                }
+                catch (Exception)
+                {
+                    RainMeadow.Error($"duplicate room {room.name} for rs {rs}");
+                    var name = "";
+                    for (var i = 0; roomSessions.Keys.Contains((name = room.name + "." + i)); i++);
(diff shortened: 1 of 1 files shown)


6fears7:
You know that feeling when you get hit with pain in areas you didn't know existed? This is that. Where in the world are we seeing duplicate room names? MSC?

youbitchoc:
JHSFHDNB
apparently it's a bug in gilded sanctuary, the duplicated room (GS_M09_S) [doesn't actually do anything](https://discord.com/channels/291184728944410624/431534164932689921/1291337339859374203)

but rain world handles it fine so i guess we gotta too

youbitchoc:
HSDJKBN

duplicate room in question is GS_M09_S in gilded sanctuary, [seems like the duplicate is completely useless](https://discord.com/channels/291184728944410624/431534164932689921/1291337339859374203)
unsure if we'll encounter more of these in the future but if rainworld handles it just fine then i guess we gotta too

6fears7:
LGTM, this shouldn't impact arena if a player replays the same level twice because world session gets destroyed and recreated but I'll keep an eye out if something weird happens.

youbitchoc:
outer expanse works fine without so looks like it's just gilded sanctuary

PR #327, opened 2024-10-04
Story: fix sheltering, refactor readyForWin

youbitchoc:


Story/StoryHooks.cs
```diff
@@ -612,37 +612,9 @@ private void KarmaLadderScreen_Singal(On.Menu.KarmaLadderScreen.orig_Singal orig
 
         private void Player_Update(On.Player.orig_Update orig, Player self, bool eu)
         {
+            if (isStoryMode(out var gameMode) && OnlinePhysicalObject.map.TryGetValue(self.abstractCreature, out var oe) && oe.isMine)
+                gameMode.storyClientSettings.readyForWin = false;
```
6fears7, line 616: would we have issues where player_update would update our local readyForWin to false and DenClose would update to true back and forth?

6fears7:
After reviewing comment, if changes are not required you are G2G. Tested with players and worked. Had one scenario where it seemed like shelter wouldn't close but could not replicate the next 5 attempts.

Story/StoryHooks.cs
youbitchoc: this back and forth happens within player.update, readyForWin should be consistent outside of this function

PR #330, opened 2024-10-05
HUD: fix alive icon after falling out of world

youbitchoc:
enabling the fall out of world prevention for story mode could have unexpected side effects


6fears7:
LGTM

PR #339, opened 2024-10-07
BanHammer: kick to lobby select, not main menu

youbitchoc:
don't call ExitToMenu, may break in arena mode

### DIFF
--- a/Game/RainMeadow.GameHooks.cs
+++ b/Game/RainMeadow.GameHooks.cs
@@ -311,7 +311,8 @@ namespace RainMeadow
 
                 if (ws.isActive) ws.Deactivate();
                 ws.NotNeeded();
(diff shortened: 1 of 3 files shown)


6fears7:
Can't decompile atm, can you expand your statement? Why would ExitToMenu break arena?

youbitchoc:
arena exits to multiplayermenu rather than mainmenu

youbitchoc:
banned player gets stuck trying to release resources and blackscreens, not sure why

henpemaz:
If you're kicking someone from the lobby, they shouldn't be allowed to stay on the lobby.

Did you mean lobbyselect?

That lobbymenu that you modified is the abandoned menu for FreeRoam.

The infinite load is because in a hook to rainworldgame.shutdownprocess we specifically check for the next process in order to decide whether to spinlock the resource release or not.

youbitchoc:
ah. whoops.

youbitchoc:
might be better to just redo this

PR #341, opened 2024-10-07
Hud pointing

6fears7:
A few things:

1. This input logic lets you:
A. If using controller, use a button and still allow movement while pointing
B. If using MK, use mouse to orient hand and allow movement

This only works so far if the player has a specific choice in the Options menu for the input. Probably need to add a default behavior for standing still and allowing vector to control input. 

We could also support double hand raising for more complex emotions if desired.

If you're *really* wanting us to have pointing only when still then I'll add an override for arena since stillness is death in that mode and I have future plans for pointing features there

https://github.com/user-attachments/assets/9ad7c8bf-556f-4ed4-bd2c-b8536ab56929

### DIFF
--- a/Arena/ArenaHooks.cs
+++ b/Arena/ArenaHooks.cs
@@ -515,7 +515,7 @@ namespace RainMeadow
             if (isArenaMode(out var _))
             {
                 self.AddPart(new TextPrompt(self));
-                self.AddPart(new Pointing(self));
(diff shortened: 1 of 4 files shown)


6fears7:
image

Default key will be primary mouse click (not pictured) unless you think there's a better button, @youbitchoc

PR #342, opened 2024-10-08
OnlineCreature: handle swallowed creatures

youbitchoc:
swallowed creatures are denoted by room -1, handle it properly

returning a null abstractcreature seems to freeze the game


6fears7:
LGTM

PR #345, opened 2024-10-09
Sync pointing

6fears7:
draft for finishing up pointing with tail


6fears7:
TODO: Pointer is desynced between slugs.

6fears7:
fixed desync, still need to work out general kinks for spear pointing. That SLerp in jolly is a doozy

6fears7:
- [ ] Could IL Hook graphics update for cleaner insertion
- [ ] Need to make either spearDir or spear spearRotation synced due to graphical issues when pointing across the wire

6fears7:
Fixed the .ANY input type. Currently still banging my head as to what I need in Player Graphics to sync the spear's new hand position.

Story/OnlineUIComponents/Pointing.cs
```diff
@@ -22,9 +23,9 @@ public override void Draw(float timeStacker)
             base.Draw(timeStacker);
 
             if (!Input.GetKey(RainMeadow.rainMeadowOptions.PointingKey.Value))
-                return;
-
-            if (OnlineManager.lobby.gameMode.avatars[0] is OnlinePhysicalObject opo && opo.apo is AbstractCreature ac)
```
6fears7, line 27: note to self to test this

6fears7:
So close. Just need to find why hand's position is still snapping back to spear spot when holding spear

Online/State/RealizedPlayerState.cs
```diff
@@ -35,6 +35,16 @@ public class RealizedPlayerState : RealizedCreatureState
...
         [OnlineField(group = "tongue", nullable = true)]
         public BodyChunkRef tongueAttachedChunk;
+        [OnlineField]
+        public bool reachingForObject;
+        [OnlineField]
+        public Vector2 absoluteHuntPos;
+        [OnlineField]
+        public Vector2 tailPos3;
+        [OnlineField]
+        public Vector2 tailPos2;
```
henpemaz, line 45: this is a LOT of floats, totals 24 bytes.

Could we maybe use `[OnlineFieldHalf]` so they're half-encoded?

Halfs have ~1 increments in 1k~2k range so for rooms bigger than that that could be a problem (that one shoreline room is 6k pixels wide, on the far right side you'd have an error of up to 4px on the half-encoded positions. A way around that is to encode these relative to mainbodychunk.pos instead so they're operating on a small range.

Online/State/RealizedPlayerState.cs
```diff
@@ -110,6 +130,20 @@ public override void ReadTo(OnlineEntity onlineEntity)
                 //if (pl.slugOnBack != null)
                 //    pl.slugOnBack.slugcat = (slugOnBack?.FindEntity() as OnlinePhysicalObject)?.apo?.realizedObject as Player;
 
+
+                if ((pl.graphicsModule as PlayerGraphics).tail != null)
+                {
+                    (pl.graphicsModule as PlayerGraphics).tail[3].pos = tailPos3;
+                    (pl.graphicsModule as PlayerGraphics).tail[3].lastPos = tailPos3;
```
henpemaz, line 137: typically setting lastpos causes animation to look more jittery than it needs to. lastpos is used to interpolate the draw-position of stuff since the game runs physics at 40fps but graphics at 60. forcing lastpos effectively causes stuff to move at 40fps while the rest of things animate at 60

Online/State/RealizedPlayerState.cs
```diff
@@ -58,6 +68,16 @@ public RealizedPlayerState(OnlineCreature onlineEntity) : base(onlineEntity)
                 tongueRequestedLength = tongue.requestedRopeLength;
                 tongueAttachedChunk = BodyChunkRef.FromBodyChunk(tongue.attachedChunk);
             }
+
+            if ((p.graphicsModule as PlayerGraphics).tail != null)
```
henpemaz, line 72: I have a gist this might nullref when a player enters-from-offscreen and spawns in the pipes system.

Having the whole thing under `if (p.graphicsModule is PlayerGraphics playerGraphics)` lets you be safe and reuse the cast.

Online/State/RealizedPlayerState.cs
6fears7: We can actually just close this whole PR. I'm going to remove the half baked implementation at the moment because there's more important items to work on.

PR #346, opened 2024-10-09
Spectate: testing ownership transfers

6fears7:
Thinking we can check the abstract creature owner and then deny ownership of resource.


6fears7:
Note to self: LocalEntityTransfers should also be denied

6fears7:
- [ ] Test spectating while in den. Wonder if it breaks when den door trying to close due to unrealized player

6fears7:
General update on this:

Seems to work. In doing so, however, discovered the issue with dupes will still be problematic when two players enter and load the room but haven't determined who's owner yet. 

Really the two outcomes are either: take away duplicates, or prevent load until ownership is determined. Dupes are a safer and more user friendly route. 

We could perform an APO.Destroy() on the non room owner's objects unless they're in grasps. 

And if room owner changed we don't need to do it again so we could try and do it on the initial load. 

Spectator can get away with it because they are not truly "there" so everything they want is ignorable.

PR #348, opened 2024-10-11
fix beehive crash

youbitchoc:


youbitchoc:
do we _really_ need to sync bees?

6fears7:
Relevant: https://go.dev/blog/errors-are-values

Individual bees position / netting? Probably not. So long as gameplay performance isn't impacted by the visual disparity (if someone doesn't get tripped up by "invisible" sticky spots).

youbitchoc:
FromStick is used in two eldritch horrors of linq collectors, I concluded that the best place to catch the error was within FromStick itself

sticky spots cannot affect the player if invisible (because they simply don't exist). the worst case is that a player joins the room right after paincone triggers and won't see the bees

6fears7:
Won't walking into bees trigger them to attack and net you? If a player cannot see them, it'd be an invisible trap.

youbitchoc:
no, because bees are clientside only

remote players cannot trigger bees for the client

we just sync the paincone popping, that's all

6fears7:
beautiful

youbitchoc:
`[Error  :RainMeadow] System.NotImplementedException: DartMaggot+DartMaggotStick`
this should fix dart maggots crashing the game too

6fears7:
LGTM

PR #352, opened 2024-10-12
refactor extensions

youbitchoc:
CanMove: whether apo is allowed to move
IsLocal: whether apo is offline or local (basically syntax sugar around `apo.GetOnlineObject(out var oe) && oe.isMine`)


youbitchoc:
snuck this into a playtesting build and nothing exploded

henpemaz:
Not going to scrutinize for logic changes but looks good at a glance

PR #354, opened 2024-10-12
fix dressmyslugcat for local players

youbitchoc:
DMS uses playerNumber, set remote players to beyond that of local


6fears7:
need this wrapped in a story mode check please

youbitchoc:
oh does this break arena?

youbitchoc:
faking playerNumber breaks lots of stuff like social events, would be easier to make DMS online-aware (needed for skin sync anyway)

PR #358, opened 2024-10-14
fix creature violence

youbitchoc:
syncs lizard violence

also flushes online state on owner transfer but very hacky


youbitchoc:
tested locally with 2s latency and 50% packet loss

6fears7:
Henpe fixed this, closing.

PR #360, opened 2024-10-14
fix spectate unrealize

youbitchoc:


youbitchoc:
spectator spawning stuff with beastmaster seems to be broken but i'm not too worried about fixing that

6fears7:
`spectator spawning stuff with beastmaster seems to be broken 
`
where do people come up with this stuff

Story/OnlineUIComponents/SpectatorOverlay.cs
```diff
@@ -218,9 +218,23 @@ public override void Update()
 
             }
 
-
+            { // manually realize room of spectated player since we detach it from roomrealizer
```
6fears7, line 221: Thanks for fixing this

youbitchoc:
> `spectator spawning stuff with beastmaster seems to be broken ` where do people come up with this stuff

i was testing it LOL

youbitchoc:
superceded by #365

PR #361, opened 2024-10-14
sync shelter close

youbitchoc:
new sheltering bool

also attempt to fix shelter close for spectators


youbitchoc:
spectating shelter close will need more work but i cba rn

PR #363, opened 2024-10-16
Story/Arena Chat feature

SaddestUnderscore:
- [x] fix typing being finicky
- [x] possibly disable other processes while chat is activate (dev tools toggle, map, etc)
- [x] steamworks implementation
- [x] chat log implementation

### DIFF
--- a/Arena/ArenaHooks.cs
+++ b/Arena/ArenaHooks.cs
@@ -826,6 +826,8 @@ namespace RainMeadow
                 self.AddPart(new TextPrompt(self));
                 self.AddPart(new Pointing(self));
                 self.AddPart(new ChatHud(self, session.game.cameras[0]));
+                self.AddPart(new SpectatorHud(self, session.game.cameras[0]));
+
 
             }
             else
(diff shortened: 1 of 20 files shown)


SaddestUnderscore:
I'll mark the second task as complete for now, this will be good to merge when disabling inputs when chat is active/pause menu is implemented

6fears7:
The only other thing I would like to see added is a timer that ticks to return the label of the username back to their username and not the username + chat message. 

Haven't tested your newest build yet so apologies if I missed that in the code. But thinking:

1. When a message is sent, start timer. 
2. Increment timer (the value I used for the initial hud fadeaway on down could be used)
3. When that time is reached, reset label back to owner.id.name. 
4. Reset timer. 

5. Bonus points is if it has a cool fadeaway effect bit not necessary. 


If you want me to work on that lmk. Otherwise great work on this!

6fears7:
- [ ] Lift out InitChat into a new Page() and new HUD element so users can optionally keep up the chatlog the whole time. 
- [ ] Find a way to disable dev tools / other tools opening on key input
- [ ] Test Story (chat overlay active while trying to sleep / death); 
- [ ] Test Arena (chat overlay active while player result menu comes up, NextLevel)
- [ ] Celebrate

PR #366, opened 2024-10-17
sync region gates

youbitchoc:
get rid of ugly gate logic duplication, sync gate transition explicitly

water flow isn't synced but not a big deal

consider syncing gate mode?


6fears7:
Is there any concern on performance running the check every lobby tick?

youbitchoc:
made water flow clientside

PR #372, opened 2024-10-19
sync creature teleports

youbitchoc:
fixes vulture, deer, leviathan, mirosbird, mirosvulture

hook CreatureTakeFlight and RPC like we do SuckedIntoShortCut

### DIFF
--- a/Game/RainMeadow.ShortcutHooks.cs
+++ b/Game/RainMeadow.ShortcutHooks.cs
@@ -14,6 +14,7 @@ namespace RainMeadow
             IL.ShortcutHandler.Update += ShortcutHandler_Update; // cleanup of entities in shortcut system
             On.ShortcutHandler.VesselAllowedInRoom += ShortcutHandlerOnVesselAllowedInRoom; // Prevent creatures from entering a room if their online counterpart has not yet entered!
 
+            On.ShortcutHandler.CreatureTakeFlight += ShortcutHandler_CreatureTakeFlight;
             On.Creature.SuckedIntoShortCut += CreatureSuckedIntoShortCut;
         }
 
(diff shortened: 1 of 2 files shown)


Online/Entity/OnlineCreature.cs
```diff
@@ -277,6 +277,51 @@ public void SuckedIntoShortCut(IntVector2 entrancePos, bool carriedByOther)
             enteringShortCut = false;
         }
 
+        public void BroadcastTookFlight(AbstractRoomNode.Type type, WorldCoordinate start, WorldCoordinate dest)
+        {
+            if (currentlyJoinedResource is RoomSession room)
+            {
+                RainMeadow.Debug(this);
+                if (id.type == 0) throw new InvalidProgrammerException("here");
```
6fears7, line 285: Intentional error message in the exception or residual from testing?

Game/RainMeadow.ShortcutHooks.cs
```diff
@@ -157,6 +158,39 @@ private bool ShortcutHandlerOnVesselAllowedInRoom(On.ShortcutHandler.orig_Vessel
...
+                RainMeadow.Debug($"{onlineCreature} took flight");
+                onlineCreature.BroadcastTookFlight(type, start, dest);
+            }
+            else if (onlineCreature.enteringShortCut) // If this call was from a processing event
+            {
+                onlineCreature.enteringShortCut = false;
+            }
+            else
+            {
+                return;
```
6fears7, line 189: Maybe add a log here just so it's not a completely silent return

6fears7 (commented):
Tested this with miros birds yesterday. Wasn't able to force a scenario to test the desired conditions but it also didn't do anything super weird. If there's a spawn point with vultures for easy testing that would be ideal I think

Online/Entity/OnlineCreature.cs
youbitchoc: copypasted from suckedintoshortcut

PR #373, opened 2024-10-20
Add filters and search bar to lobby menu

### LINKED ISSUE #94: Lobby List - Filter by Game Mode
From the Trello

T1mbits:
Added filter options and a search bar to better navigate lobbies. Also fixed some overflowing text from the meadow and story mode descriptions as well as greyed out the "about", "stats", and "unlocks" buttons.

I think this fixes issue #94 

https://github.com/user-attachments/assets/13e6a172-34a6-4c47-adb8-6d04a9d9465a

### DIFF
--- a/Menu/LobbySelectMenu.cs
+++ b/Menu/LobbySelectMenu.cs
@@ -13,542 +13,542 @@ using UnityEngine;
 
 namespace RainMeadow
 {
-	public class LobbySelectMenu : SmartMenu, SelectOneButton.SelectOneButtonOwner, CheckBox.IOwnCheckBox
-	{
-		private List<FSprite> sprites;
-		private EventfulSelectOneButton[] lobbyButtons;
-		private LobbyInfo[] lobbies;
-		private LobbyInfo[] displayLobbies;
-		private LobbyFilter filter;
-		private OpComboBox2 filterModeDropDown;
-		private OpComboBox2 filterPasswordDropDown;
-		private OpTextBox filterLobbyLimit;
-		private OpTextBox lobbySearchInputBox;
-		private float scroll;
-		private float scrollTo;
-		private int currentlySelectedCard;
-		private OpComboBox2 visibilityDropDown;
-		private OpTextBox lobbyLimitNumberTextBox;
-		private SimplerButton playButton;
-		private SimplerButton refreshButton;
-		private OpComboBox2 modeDropDown;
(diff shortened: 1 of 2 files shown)


T1mbits:
phanie suggested that I use checkboxes instead for the gamemode filter, but if custom game modes are going to be supported that will probably cause clutter so maybe I'll make a multi-selection dropdown menu

also I think the mode filter will need to be rewritten entirely if custom game modes are going to be supported

Menu/LobbySelectMenu.cs
```diff
@@ -260,6 +309,70 @@ public override void Update()
...
+            public enum ModeFilter
+            {
+                All,
+                Meadow,
+                Story,
+                ArenaCompetitive
+            };
+            public enum PasswordFilter
+            {
+                [Description("Don't care")]
```
6fears7, line 358: Update these please so we can have the game _attempt_ to benefit translating them like pictured below:

image

T1mbits: image

like so correct?

henpemaz:
Really good first PR. I personally don't like the placement but that can be figured out later, good to go.

T1mbits:
do you have any suggestions  on where I can  move it to? I'm thinking about making a multi-selection dropdown for the game modes already so I could move the filter stuff  wherever you want while I'm doing that

6fears7:
Personally, I'd like it more centered with the  fields above. And id like to see a horizontal separator with titles so I can clearly delineate between "Create Lobby" and "Filter Lobby", since right now on that page we have both.

6fears7:
Id also want 

1. the larger Create button removed from the lobby selection list. 

2. Clicking a lobby joins it, instead of click to select and click separate button to join 

3. Create / join button becomes "create" and is moved to be with the creation fields 

^ these might give the space some breathing room

T1mbits:
Do the "about", "stats", and "unlocks" buttons have any planned purpose or should I remove those too? I know in meadow mode that there are unlocks but I'm not quite sure what it would do for other game modes. Stats seems a bit weird  unless you're showing playtime/deaths/cycles in your save and about should be covered by the description underneath the mode selection

6fears7:
I won't speak for Henpe but I don't see the usefulness in those buttons outside of a particular game mode. I would anticipate modded modes to have descriptions inside their process menus and not lobby select

PR #374, opened 2024-10-20
Update spectate ownership transfer

6fears7:


Story/OnlineUIComponents/SpectatorHud.cs
```diff
@@ -53,6 +53,7 @@ public override void Update()
                     RainMeadow.Debug("Shutting down spectator overlay due to another process request");
                     spectatorOverlay.ShutDownProcess();
                     spectatorOverlay = null;
+                    OnlineManager.mePlayer.isActuallySpectating = false;
```
youbitchoc, line 56: we don't restore camera when spectator hud closes, currently player has to spectate themself

better to check if ac being spectated is ours

youbitchoc (changes requested):

PR #378, opened 2024-10-23
Fix some lobby menu bugs

T1mbits:
Synced background to user's chosen submenu to fix my OCD.

Changed LobbyInfoCard to SimplerButtons that immediately allow you to connect and changed the create/join button to a dedicated create button since I removed the lobby create card. This should also fix a bug where if you selected a lobby card but didn't join you wouldn't be able to create a lobby until you reentered the menu because there was no create card to click on.

Still working on a mockup for a full redesign, it's just that I don't think I'll be done before the next release and these were some simple bugs to fix in my previous meddling of the menu. Honestly, there's a high likelihood this broke something else but I didn't notice anything wrong.

### DIFF
--- a/Menu/LobbySelectMenu.cs
+++ b/Menu/LobbySelectMenu.cs
@@ -13,10 +13,10 @@ using UnityEngine;
 
 namespace RainMeadow
 {
-    public class LobbySelectMenu : SmartMenu, SelectOneButton.SelectOneButtonOwner, CheckBox.IOwnCheckBox
+    public class LobbySelectMenu : SmartMenu, CheckBox.IOwnCheckBox
     {
         private List<FSprite> sprites;
-        private EventfulSelectOneButton[] lobbyButtons;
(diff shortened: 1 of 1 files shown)


T1mbits:
maybe I'm hallucinating but I could've sworn I removed the create lobby card last time but I guess not. Well, it's gone now and it probably won't cause any problems???

6fears7:
code looks good. Do you mind posting a video of the changes?

T1mbits:
After thinking about it, I think I'll just keep the lobby create card to minimize confusion, but I'll keep the background fix. Or maybe it's not worth it and I'll just close the PR and just fix everything in the redesign PR

6fears7:
Just got through testing this. I like it a lot. Do you still want to close this PR / update it or are you good for a merge?

T1mbits:
If you want to merge it then go ahead, I don’t have anything against it as long as it works properly

6fears7:
Tested, works.

PR #380, opened 2024-10-24
Destroy spear if stuckObject is null to prevent state read failure

6fears7:
- [ ] Test


6fears7:
Deletes all spears. Do not do this.

PR #398, opened 2024-10-28
Story: fix lobby freeze when owner leaves

youbitchoc:


youbitchoc:
we should kick players out but for now just allow them to leave without freezing

PR #400, opened 2024-10-29
Story: make use of gamestart hooks

youbitchoc:
should work, needs testing


6fears7:
LGTM to me as well. If you don't test tomorrow I'll drop a build for it. 

Things to test:

- [ ] Host dies, then spectate client. Enter den. Creatures should be present next round.
- [ ] Dens correctly view player state
- [ ] Clients can join after progressing to a new region.
- [ ] Clients cannot join until after den -> next cycle is complete.

PR #403, opened 2024-10-30
6fears7 arena block kicking players midgame

6fears7:


6fears7:
closing in favor of #405

PR #405, opened 2024-10-30
6fears7 arena v74

6fears7:
- Shows readied up players in list overflow so you can know for certain who you're waiting on
- Allows kicking players + kicking players in overflow list
- Blocks kicking players in-game from using spectate

- [x] Investigate JoinImpl and fix before this is merged

- [x] Test
- [x] Fix war sound playing every 5s


6fears7:
Fix war drum playing every 5s

PR #408, opened 2024-10-31
JoinImpl: Handle Undefined Node & Undefined Tile

6fears7:
So the if/else in JoinImpl is assuming that if the Tile is not defined, the node must be defined. However, in cases like the arena map Multi_GodOfEyes (and sometimes I assume in story when invis), the` abstractNode is -1`. And NodeDefined is explicity if `abstractNode > -1`. 

So  I added a force realization to the room` if Node && Tile == undefined`. That enabled Slugcats joining who were hitting index out of range errors (because` NodeDefined = -1)` to have a way to get here.

### DIFF
--- a/Online/Entity/OnlinePhysicalObject.cs
+++ b/Online/Entity/OnlinePhysicalObject.cs
@@ -261,6 +261,8 @@ namespace RainMeadow
                 else if (inResource is RoomSession newRoom)
                 {
                     RainMeadow.Debug($"room join");
+                    RainMeadow.Debug($"topos Tile defined? {topos.TileDefined}");
+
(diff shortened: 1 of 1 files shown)


6fears7:
- [ ] Test in story
- [ ] Test in longer arena sessions (worked when navigating between multi_rooms and single rooms but want to just find edge cases)

6fears7:
From Henpe:

For Arena:

`when a client spawns the abstractcreature of the player, realizes it and shoves it into the shortcut system, make sure it has a proper .pos set with the node of the shortcut that it's being used`

shouldn't need to force realize if we do it right.

Will close this PR and then delete branch when this issue is resolved.

PR #410, opened 2024-11-01
Story: fix swallow tracker

youbitchoc:
remove persistent tracker for remote swallowed items

was breaking map and sleepanddeathscreen


youbitchoc:
remix key item tracking will need more stuff:
specifically, finding the player that puked it up

we can't just check ownership or grasps, looping thru playerstates could be racy

not worth the effort imo

PR #415, opened 2024-11-02
make slugcatStats per-player

youbitchoc:
cwt of player:slugcatstats

fixes slight issue with clientside desync
we should consider syncing slugcatstats state (malnourished, pup, ...)


youbitchoc:
also move unserialized playerstate init to deserialization procedure

6fears7:
GOAT

PR #416, opened 2024-11-02
fix double-grabbing items

youbitchoc:


youbitchoc:
tested to work in arena

6fears7:
LGTM, merge at discretion

PR #426, opened 2024-11-05
fix creature extraData serialization

youbitchoc:
fixes mismatched centipede lengths and shelter failure spawns


youbitchoc:
also change serialization world check to use singleworldroom rather than assuming from arena gamemode

PR #432, opened 2024-11-06
don't release swallowed items

youbitchoc:
needs testing

technically belongs to region but we should keep ownership


youbitchoc:
superseded by #437

PR #443, opened 2024-11-11
Arena: Remixes & Fixes

6fears7:
setup RPCs for remix options, fixed ready up not notifying clients, fixed timer pausing when someone joins lobby, fixes slug class

- [x] Test


6fears7:
Tested, works.

PR #446, opened 2024-11-13
Spectate: fix crash when spectatee unrealized

youbitchoc:
happens when snagged by vulture and dragged out of room


6fears7:
LGTM

PR #448, opened 2024-11-13
6fears7 story prevent self kill with spear

6fears7:
This prevents killing yourself with a normal spear. Are there any conditions in base game where this kind of behavior would be expected?

Behavior now is that players are temporarily stunned from the physics of the spear impact. 

https://github.com/user-attachments/assets/af5fd18e-e6f3-4016-af62-1f75e92e7049

(worse) Alternatives:

1. Change collision layer for spear after it's collided (messier code)
2. Change collision layer for player after an object is thrown in those rooms
3. Remove change in player collision in rooms

### DIFF
--- a/Game/RainMeadow.PlayerHooks.cs
+++ b/Game/RainMeadow.PlayerHooks.cs
@@ -50,6 +50,7 @@ public partial class RainMeadow
     }
 
 
+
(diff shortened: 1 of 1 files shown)


6fears7:
https://github.com/user-attachments/assets/053b3dbd-d3ea-45a7-a059-28a5d76216e7

youbitchoc:
lgtm

PR #450, opened 2024-11-13
fix modapplier

youbitchoc:


youbitchoc:
needs a bigger refactor, closing for now

PR #453, opened 2024-11-15
refactor modapplier, add msc in check

youbitchoc:


youbitchoc:
subresource mismatch is kicking us out before we can apply mods
and anyways we should be putting modlist in metadata

PR #456, opened 2024-11-17
Add story and arena to vscode build tasks

xamionex:
title


youbitchoc:
ideally we should make local games use the selected gamemode instead

henpemaz (approved):
Haven't tested but this seems to mimic well enough the vs setup, good to go.

PR #457, opened 2024-11-17
Slugpup players

xamionex:
- Mustfix:
  - [ ] Fix saint being bald
  - [ ] Confirm whether or not tails are full length for some slugcats (and if they are, fix it)
  - [x] Fix stats for custom cats
  - [ ] Change the way we get playerIsSlugpup (right now it changes the bool in PlayerHooks from StoryMenu)

- Not sure if this is because of this implementation, will look into these
  - [ ] Fix name buggyness when client leaves
  - [ ] Fix text below name disappearing too fast

### DIFF
--- a/Game/RainMeadow.PlayerHooks.cs
+++ b/Game/RainMeadow.PlayerHooks.cs
@@ -8,6 +8,7 @@ using UnityEngine;
 using MonoMod.RuntimeDetour;
 using System.Runtime.CompilerServices;
 using MoreSlugcats;
+using MonoMod;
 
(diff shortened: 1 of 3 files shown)


xamionex:
- [ ] Check for MSC before adding checkbox

henpemaz:
So we've talked a bit and seems like something like this will be better as its own separate mod. The sheer amount of IL hooking there scares me a bit and I'd rather keep to a minimum how much base Meadow interacts with DLC stuff so we don't have any dependency headaches.

PR #460, opened 2024-11-18
hardset bodychunk pos

youbitchoc:
should fix rubberbanding centipedes


henpemaz:
This causes the animation of anything moving at more than 1px/tick to become very jittery. There's probably a way to solve the centipedes problem in a more specific way.

PR #463, opened 2024-11-19
add isPup sync

xamionex:
Support for [Rain Meadow Pupifier](https://github.com/xamionex/RainMeadowPupifier) (and additionally other mods that want to change isPup)

Basically just syncs the playerstate.isPup between players
It doesn't do anything functionally other than telling the game how to visually render us
So for people without pup mods, we look like pups, but it doesn't do anything more than that
But for people with pup mods, it enables syncing if they're a pup and the mods will function normally

- Tested with [RMP](https://github.com/xamionex/RainMeadowPupifier) and it worked fine
- Tested without RMP and it worked fine
- Tested with/without mod in gourmand ending, worked both ways
- It works when we have host and client discrepancy (aka host has mod/client doesnt, host has/client doesnt)

Comment anything you want me to test here

### DIFF
--- a/Online/State/RealizedPlayerState.cs
+++ b/Online/State/RealizedPlayerState.cs
@@ -14,6 +14,8 @@ namespace RainMeadow
         private bool standing;
         [OnlineField]
         private bool glowing;
+        [OnlineField]
+        private bool isPup;
         [OnlineField(nullable = true)]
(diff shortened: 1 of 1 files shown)


6fears7:
Works for me. @youbitchoc any other thoughts?

PR #465, opened 2024-11-21
add blacklist

Persondotexe0:


henpemaz:
Your commit is either changing line endings or tabulation or something like that because it picks all lines as changed, can you fix that?

Persondotexe0:
> Your commit is either changing line endings or tabulation or something like that because it picks all lines as changed, can you fix that?

Yes, can in a few hours.

PR #466, opened 2024-11-21
fix null entity readto

youbitchoc:
var accepts null


henpemaz:
I had to test it to believe it, now I just want to fill a complaint against the .net guys

youbitchoc:
was hoping for a more idiomatic way, but I think
```C#
if (def.entityId.FindEntity() is not null and var ent)
```
is the best we can do

PR #468, opened 2024-11-21
6fears7 storymenu rewrite

6fears7:
### New Game
- Friendly Fire toggled
- Host color swatch shown
- Client gets "New Game" label 
image

### Game With Save
- Client gets Host's current region from their save
image

### Notables

- Moved a lot of the Story Menu items to a helper file to clean up the Story Menu presentation itself
- Doesn't have tertiary color support applied to online slugs
- StoryLobbyData: `region` string property added to notify clients of where the current save will be loading into
- Had to add some null reff handling in `Player_checkInput` with our custom hook due to how Hunter campaign loads seemingly slower with the real StoryMenu
- Adds @xamionex 's FriendlyFire checkbox

### Client Nuance
When a client chooses a custom slugcat (as most do):

If you want to have the color palette for each individual slugcat on click, we have to set the slugcat page index to the client's chosen slugcat. Not a problem. 

However, when you do that, StoryMenu wants to retrieve the savestate for the new page index. If you've never played that slugcat before, your save data is null and the menu null reffs because it's unclear if you are on a NewPage or a ContinuePage.

This can be avoided by filtering out the client's custom slugcat choices by checking that each slugcat save file if its null, but reduces their options if they're new to the game. Playing as a given slugcat at least once adds it to the list if that data was null.

Additionally, the tertiary color is not synced or applied yet so I could really go either way on this one.


Story/StoryMenuRedux.cs
```diff
@@ -0,0 +1,226 @@
...
+using Steamworks;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using UnityEngine;
+using HUD;
+namespace RainMeadow
+{
+    public class StoryMenuRedux : SlugcatSelectMenu, SelectOneButton.SelectOneButtonOwner
```
youbitchoc, line 12: I'd prefer calling it StoryCustomMenu

Story/StoryMenuHelpers.cs
```diff
@@ -0,0 +1,410 @@
...
+
+                }
+            }
+
+        }
+
+        #endregion
+
+        #region SlugcatSetup
+        private static List<SlugcatStats.Name> AllSlugcats()
```
youbitchoc, line 235: can't we just reuse vanilla?

Story/StoryMenuRedux.cs
```diff
@@ -0,0 +1,226 @@
...
+        internal CheckBox clientWantsToOverwriteSave;
+        internal CheckBox friendlyFire;
+        internal EventfulSelectOneButton[] playerButtons = new EventfulSelectOneButton[0];
+        internal EventfulHoldButton hostStartButton;
+        internal EventfulHoldButton clientWaitingButton;
+        internal SlugcatCustomization personaSettings;
+        private PlayerInfo[] players;
+        internal bool resetSave;
+        private bool updateDefaultColors;
+        internal StoryGameMode storyModeOnline;
```
youbitchoc, line 24: `storyGameMode`

Story/StoryMenuHelpers.cs
```diff
@@ -0,0 +1,410 @@
...
+
+
+        }
+
+        public static void GetRegionAndCampaignNameForClient(StoryMenuRedux storyMenu, StoryGameMode storyModeOnline)
+        {
+            if (!OnlineManager.lobby.isOwner && storyMenu.onlineDifficultyLabel != null)
+            {
+
+                storyMenu.onlineDifficultyLabel.text = storyModeOnline.region != "" ? GetCurrentCampaignName(storyModeOnline) + $" - {storyModeOnline.region}" : GetCurrentCampaignName(storyModeOnline) + storyMenu.Translate(" - New Game");
```
youbitchoc, line 384: can be simplified

Story/StoryMenuRedux.cs
```diff
@@ -0,0 +1,226 @@
...
+            else //I'm the host
+            {
+
+
+                RainMeadow.Debug("CURRENT CAMPAIGN: " + StoryMenuHelpers.GetCurrentCampaignName(storyModeOnline));
+                personaSettings.playingAs = storyModeOnline.currentCampaign;
+            }
+
+            manager.arenaSitting = null;
+            if (restartChecked)
```
youbitchoc, line 96: pretty sure story menu already had this functionality, can we reuse that?

youbitchoc (changes requested):

Story/StoryMenuRedux.cs
6fears7: nw

Story/StoryMenuRedux.cs
6fears7: - [ ] Rename Class
- [ ] storyModeOnline => storyGameMode
- [ ] Retrieve built-in sluglist
- [ ] simplify onlineDifficultyLabel
- [ ] Cast client slugcat page depending on save state presence

6fears7:
- [ ] Rename Class
- [ ] storyModeOnline => storyGameMode
- [ ] Retrieve built-in sluglist
- [ ] simplify onlineDifficultyLabel
- [ ] Cast client slugcat page depending on save state presence

6fears7:
Tested slugcats with empty saves and present saves, selected fine. 
Tested modded slugcat selection + color palette. 
Tested ensuring all buttons were deselected to inherit campaign slug choice

PR #469, opened 2024-11-21
fix SS_E08 gradient gravity again

youbitchoc:
needs testing


youbitchoc:
previous was working like
```C#
if (room.physicalObjects[i][j] is Player || self.room.physicalObjects[i][j].abstractPhysicalObject.IsLocal())
```
rather than `&&`

6fears7 (approved):
I have no idea if you’re waiting for approvals or just committing to pull in later but LGTM once proven tested

PR #481, opened 2024-11-26
fix and refactor onlinehud

youbitchoc:


Story/OnlineUIComponents/SpectatorHud.cs
```diff
@@ -46,23 +46,9 @@ public override void Update()
...
-                    }
-
-
-                }
-
-                if (RainMeadow.isArenaMode(out var _))
-                {
-
-                    if (game.arenaOverlay != null || game.pauseMenu != null || game.manager.upcomingProcess != null)
+                    if (game.pauseMenu != null || game.arenaOverlay != null || camera.hud.map.visible || game.manager.upcomingProcess != null)
```
6fears7, line 51: heads up, camera hud map check will break arena

youbitchoc: changed it to `camera.hud?.map?.visible is true`

Story/OnlineUIComponents/PlayerSpecificOnlineHud.cs
```diff
@@ -123,20 +123,19 @@ public override void Update()
...
             }
 
-            if (!clientSettings.inGame) return;
+            this.found = false;
+
             if (camera.room == null || !camera.room.shortCutsReady) return;
-            if (abstractPlayer == null)
+            if (!clientSettings.inGame) return;
+            if (clientSettings.avatars.Count == 0) return;
+            if (clientSettings.avatars[0]?.FindEntity(true) is OnlineCreature oc) // TODO: support multiple avatars
```
youbitchoc, line 131: I don't like this but
1. negligible performance cost
2. not enough spoons to refactor everything

Story/OnlineUIComponents/SpectatorHud.cs
```diff
@@ -46,23 +46,9 @@ public override void Update()
...
-                    }
-
-
-                }
-
-                if (RainMeadow.isArenaMode(out var _))
-                {
-
-                    if (game.arenaOverlay != null || game.pauseMenu != null || game.manager.upcomingProcess != null)
+                    if (game.pauseMenu != null || game.arenaOverlay != null || camera.hud?.map?.visible is true || game.manager.upcomingProcess != null)
```
6fears7, line 51: Assuming you jumped into arena just to sanity verify this won't create issues, refactor lgtm

6fears7 (commented):
Also much better bookean "mutedPlayers"

Story/OnlineUIComponents/SpectatorHud.cs
youbitchoc: yup, tested spectate to work just fine in arena

6fears7 (approved):

PR #482, opened 2024-11-26
fix spearmaster spear sync

youbitchoc:
needs testing


youbitchoc:
not too proud of that IL hook but meh

PR #487, opened 2024-11-28
fix story menu breaking when remix disabled

youbitchoc:
also clean up a bunch of other stuff


Story/StoryMenuHelpers.cs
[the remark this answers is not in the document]
youbitchoc: > It’s possible you are updating the client page index elsewhere

yup, see SetCurrentlySelectedOfSeries

Story/StoryOnlineMenu.cs
```diff
@@ -134,90 +116,168 @@ public override void ShutDownProcess()
...
+
+        public void SetCurrentlySelectedOfSeries(string series, int to)
+        {
+            switch (series)
             {
-                var playerbtn = playerButtons[i];
-                playerbtn.RemoveSprites();
-                this.pages[0].RemoveSubObject(playerbtn);
+                case "scugButtons":
+                    slugcatPageIndex = to;
```
6fears7, line 157: I was expecting to see RemoveExcessStoryObjects called here again for when the page update triggers a new slugcat, but I only see it called when the client first joins the lobby. Can you verify?

youbitchoc: I changed RemoveExcessStoryObjects to remove for all pages

6fears7: Awesome

PR #494, opened 2024-11-30
Sync dropwigs

youbitchoc:


Game/RainMeadow.CreatureHooks.cs
```diff
@@ -157,6 +160,18 @@ private void GarbageWormAI_Update(On.GarbageWormAI.orig_Update orig, GarbageWorm
...
+            if (!self.AI.creature.IsLocal()) return;
+            orig(self);
+        }
+
+        private void DropBugAI_CeilingSitModule_JumpFromCeiling(On.DropBugAI.CeilingSitModule.orig_JumpFromCeiling orig, DropBugAI.CeilingSitModule self, BodyChunk targetChunk, Vector2 attackDir)
+        {
+            if (!self.AI.creature.IsLocal()) return;
+            orig(self, targetChunk, attackDir);
+        }
+
```
henpemaz, line 174: Not a blocker for now but more of a though, never running these on client means we don't get the drop sfx to play when it activates, right?

henpemaz (approved):
gnuff!

Game/RainMeadow.CreatureHooks.cs
youbitchoc: yup, no sound

jumpfromceiling plays the sound, we could rpc that
(dislodge just calls jumpfromceiling)

PR #496, opened 2024-11-30
sync deer antlers grab

phanerka:
🥴

### DIFF
--- a/Online/State/RealizedPhysicalObjectState.cs
+++ b/Online/State/RealizedPhysicalObjectState.cs
@@ -8,7 +8,7 @@ namespace RainMeadow
     [DeltaSupport(level = StateHandler.DeltaSupport.NullableDelta)]
     public class RealizedPhysicalObjectState : OnlineState
     {
-        public float chunkPosThreshold = 0f;
+        public float chunkPosLeniency = 0f;
 
         [OnlineField] // todo field that does sequenceequals or type that handles these better
         private ChunkState[] chunkStates; // pretty sure these get sent every time because the array comparison doesn't do sequenceequal but does referenceequal instead
@@ -28,12 +28,13 @@ namespace RainMeadow
             var po = (onlineEntity as OnlinePhysicalObject).apo.realizedObject;
             for (int i = 0; i < chunkStates.Length; i++) //sync bodychunk positions
             {
-                chunkStates[i].ReadTo(po.bodyChunks[i], chunkPosThreshold);
+                chunkStates[i].ReadTo(po.bodyChunks[i], chunkPosLeniency);
             }
             if (po.collisionLayer != collisionLayer)
             {
                 po.ChangeCollisionLayer(collisionLayer);
             }
+            this.chunkPosLeniency = 0f;  // restore to default in case this object is reused
         }
     }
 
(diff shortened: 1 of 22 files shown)


Online/State/RealizedPhysicalObjectState.cs
```diff
@@ -8,6 +8,8 @@ namespace RainMeadow
     [DeltaSupport(level = StateHandler.DeltaSupport.NullableDelta)]
     public class RealizedPhysicalObjectState : OnlineState
     {
+        public float chunkPosThreshold = 0f;
+
```
henpemaz, line 12: whatever value you put here is at mercy of the incoming flow of state-objects and the state of the delta-encoding stream.

if the incoming object is in full, ie a new room owner or new object owner, then it will be overwritten, if the incoming is delta, then the value will be copied from a baseline.

basically all of this is to say, it's not safe to put it here. this should be in onlinephysicalobject instead

henpemaz:
So two things, first of, @phanerka how much of did did you write yourself? Are the commits accurate? I just want to have an idea.

henpemaz:
And then second thing is, I don't like this mechanism (chunkPosThreshold) much, I think I'd rather it be one of the following:

1. cheap solution where realizedplayerstate.readto doesn't call base depending on incoming animation state or deer state

2. more complex solution where the sender sends a boolean flag or a similar way for the sender to specify "physics driven position"

If there's a problem where a grasp is interrupted in the client because the distance became too great, when re-assigning that grasp it's probably safe to move mainbodychunk to a non-interrupting distance

Online/State/RealizedPhysicalObjectState.cs
youbitchoc: we don't rely on persistence, it's just a convenient way to tell the RPOstate chunk reader whether to be more lenient
I considered refactoring readto to take it as an argument but that'd require forwarding it through the chain of base.readto's which is very ugly

youbitchoc:
> more complex solution where the sender sends a boolean flag or a similar way for the sender to specify "physics driven position"

imo we shouldn't make chunkstate nullable because new participants should still need to see state, physics interpolation won't work well if the physics objects are screens away from each other

bool in realizedphysicalobject sounds good to me, but if we wanna handle grasps between different owners we should keep a reference to the owner object

perhaps this should go into the grasps system instead?

Online/State/RealizedPhysicalObjectState.cs
henpemaz: you'll need a mechanism to set it back to default consistently then, either before of after use, don't rely on the constructor default because the object might be copied from a baseline

Online/State/RealizedPlayerState.cs
```diff
@@ -95,10 +164,11 @@ public Player.InputPackage GetInput()
             i.analogueDir.y = analogInputY;
             return i;
         }
-
+        
         public override void ReadTo(OnlineEntity onlineEntity)
         {
             RainMeadow.Trace(this + " - " + onlineEntity);
+            if (playerInAntlersState != null) this.chunkPosThreshold = 4f;
```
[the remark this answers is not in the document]
youbitchoc: I have a followup PR at https://github.com/phanerka/Rain-Meadow/pull/2 that fixes this

Online/State/RealizedPhysicalObjectState.cs
youbitchoc: oops didn't realize we could restore to default after, I'll update the PR

henpemaz (approved):
aight gnuff

PR #503, opened 2024-12-03
v75 hotfixes

6fears7:
- Adds back early rain exit in Arena
- Fixes issue where clients couldnt join Story menu if their computers were slower


henpemaz (approved):
lgtm

PR #508, opened 2024-12-04
modloader: disable load order requirement for now

youbitchoc:


youbitchoc:
fix modloader instead

PR #509, opened 2024-12-05
sync player grabbing vines

youbitchoc:
doesn't sync the position of the vine itself, which can get noticeably desynced

also change chunkPosLeniency to a bool in opo that skips chunk sync entirely rather than relying on a threshold, as suggested by henpe

(the diff looks weird because i unindented some stuff)

### DIFF
--- a/Online/Entity/OnlinePhysicalObject.cs
+++ b/Online/Entity/OnlinePhysicalObject.cs
@@ -92,6 +92,7 @@ namespace RainMeadow
 
         public readonly AbstractPhysicalObject apo;
         public bool realized;
+        public bool lenientPos;
 
         public bool beingMoved;
         public static ConditionalWeakTable<AbstractPhysicalObject, OnlinePhysicalObject> map = new();
--- a/Online/State/RealizedPhysicalObjectState.cs
+++ b/Online/State/RealizedPhysicalObjectState.cs
@@ -8,8 +8,6 @@ namespace RainMeadow
     [DeltaSupport(level = StateHandler.DeltaSupport.NullableDelta)]
     public class RealizedPhysicalObjectState : OnlineState
     {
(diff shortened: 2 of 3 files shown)


youbitchoc:
we should probably sync untethered climbables too (specifically those coral neuron thingies in pebbles) but that'll require a new onlineentity? so probably a separate PR

Extensions.cs
```diff
@@ -216,5 +217,14 @@ public static Color SafeColorRange(this Color valuecolor)
...
         }
+
+        public static ILCursor GotoNextOutEnd(this ILCursor cursor, out ILLabel end, params Func<Mono.Cecil.Cil.Instruction, bool>[] predicates)
+        {
+            var ret = cursor.GotoNext(moveType: MoveType.AfterLabel, predicates);
+            var c = cursor.Clone();
+            c.Index += predicates.Length;
+            end = c.MarkLabel();
+            return ret;
+        }
```
henpemaz, line 228: 🤨 

I see the attempt at making a pattern out of these types of matches, but 1. it's not a lot of code; 2. they'll all have small nuances specially with how if/else blocks might be organized; and lastly 3. I wouldn't encourage matching the whole body of an if, matches are ideally as short as possible so to not be affected by other mods' edits

henpemaz (approved):
Haven't checked the whole diff but the stuff thats new looks alright, couple notes:

as is oc.lenientpos might be a tick late (inc. stuff sets it to true but base already ran), not sure if intented, not sure if might have any meaningful side-effects

that `oc.lenientPos = false;` might get moved around later when more things start using this system, but no worries for now

youbitchoc:
> oc.lenientpos might be a tick late

split it out into a separate function so hopefully base.readto won't be too cluttered

henpemaz (approved):

PR #514, opened 2024-12-06
Story: sync passages

youbitchoc:
host decides

### DIFF
--- a/Story/StoryHooks.cs
+++ b/Story/StoryHooks.cs
@@ -38,6 +38,7 @@ namespace RainMeadow
             On.Menu.FastTravelScreen.StepRegion += FastTravelScreen_StepRegion;
             On.Menu.FastTravelScreen.InitiateRegionSwitch += FastTravelScreen_InitiateRegionSwitch;
             On.Menu.FastTravelScreen.Update += FastTravelScreen_Update;
+            On.Menu.FastTravelScreen.Singal += FastTravelScreen_Singal_ClientLoadGameNormally;
 
             On.HUD.HUD.InitSinglePlayerHud += HUD_InitSinglePlayerHud;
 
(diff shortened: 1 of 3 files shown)


youbitchoc:
are there any worldloading race conditions i should look out for?

Story/StoryHooks.cs
```diff
@@ -802,6 +809,95 @@ private void RainWorldGame_GoToDeathScreen(On.RainWorldGame.orig_GoToDeathScreen
...
+            orig(self, manager, ID);
+            if (isStoryMode(out _) && !OnlineManager.lobby.isOwner)
+            {
+                StoryMenuHelpers.RemoveMenuObjects(self.mapButtonPrompt, self.prevButton, self.nextButton, self.chooseButton, self.buttonInstruction);
+            }
+        }
+
+        private void FastTravelScreen_ctor_ClientDontFilterRegions(ILContext il)
+        {
+            try
```
6fears7, line 842: Nitpick: commenting the changes to the IL. I can infer from the method name though. Your call.

6fears7 (commented):
I realize this got converted back to a draft but I'll just post the comment for now.

Story/StoryHooks.cs
youbitchoc: it's just
```diff
  bool flag = false;
+ if (isStoryMode(out _) && !OnlineManager.lobby.isOwner) flag = true;
```

and to understand what flag does you'll have to look at the IL anyway

henpemaz:
> are there any worldloading race conditions i should look out for?

The passage-screen performs world-loading to get room data and map data but it might already have been addressed that it doesn't require/requests the resource.

Then finally using a passage also performs a quick world-loading where it loads the game in the now-modified save position, saves and immediately exits out into a new rainworldgame instance to re-start the cycle. This is iirc detectable with the manager.menusettings.newgamecondition.IsFastTravel, see where and how that's used by the game.

Imo, this "passage handling" cycle-load is probably best to have be performed only in the host and have clients be waiting in menus until the real-deal cycle starts.

youbitchoc:
> The passage-screen performs world-loading to get room data and map data but it might already have been addressed that it doesn't require/requests the resource.

yup, fast travel screen uses a null game so it gets ignored

> Imo, this "passage handling" cycle-load is probably best to have be performed only in the host

aight

Story/StoryHooks.cs
```diff
@@ -1,4 +1,5 @@
 using HUD;
+using IL.Menu;
```
henpemaz, line 2: this looks like it could cause trouble

henpemaz:
other than that seems alright at a glance, I didn't check the specifics of the new hooks

Story/StoryHooks.cs
youbitchoc: whoops, fixed

youbitchoc:
the passage ones are for stopping client from changing region and syncing host selected region

other significant change is using default denpos from synced savestring rather than syncing separately (now it's just used as an indicator for passage / custom ending denpos, not happy with it but it works well enough)

6fears7 (approved):

PR #517, opened 2024-12-06
Icon adjustments

Zed9876:
Some adjustments to the shelter and gate icons.

### DIFF
--- a/Story/OnlineUIComponents/OnlineHud.cs
+++ b/Story/OnlineUIComponents/OnlineHud.cs
@@ -1,6 +1,7 @@
 ﻿using HarmonyLib;
 using HUD;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using UnityEngine;
 
@@ -12,6 +13,8 @@ namespace RainMeadow
 
         private RoomCamera camera;
         private readonly OnlineGameMode onlineGameMode;
+
+        public int hudCounter;
 
         public OnlineHUD(HUD.HUD hud, RoomCamera camera, OnlineGameMode onlineGameMode) : base(hud)
         {
@@ -62,6 +65,7 @@ namespace RainMeadow
         {
             base.Update();
             UpdatePlayers();
+            hudCounter++;
         }
     }
 }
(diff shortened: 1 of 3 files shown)


Story/OnlineUIComponents/PlayerSpecificOnlineHud.cs
```diff
@@ -85,6 +85,14 @@ public bool PlayerInShelter
             }
         }
 
+        public bool PlayerInAncientShelter
+        {
+            get
+            {
```
6fears7, line 91: May want to do either do a safety check in game to ensure non-MSC won't break here or wrap this in ModManager.MSC

6fears7 (commented):
If you could also submit a video of it in action that would be awesome. I trust you, I just don't trust myself to catch every piece of code

Story/OnlineUIComponents/OnlineHud.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using HarmonyLib;
 using HUD;
 using System.Collections.Generic;
+using System.Diagnostics;
```
youbitchoc, line 4: not needed?

Story/OnlineUIComponents/PlayerSpecificOnlineHud.cs
youbitchoc: imo it's safe as is, it's a bool that defaults to false

6fears7: Works for me

Story/OnlineUIComponents/OnlineHud.cs
6fears7: Probably just added when testing different methods to find the right one

6fears7: @Zed9876 remove the newly added library and then merge. I'll pre approve

6fears7 (approved):
Remove added diagnostics lib and will merge

Zed9876:
Video of Icon adjustments:

https://www.youtube.com/watch?v=IRHGuOHZu34&ab_channel=JacobPreece

6fears7 (approved):

PR #518, opened 2024-12-07
Plop and Music Update

Intikus:
Meadow Music is now synced in between players.
Vibe rooms/Hubs now play Plopmachine, a proc-gen music engine.
Meadow ambience music and Plopmachine volume will cross fade.
Vanilla songs can play in meadow mode. 
Meadow songs no longer lag when loading.
13 new meadow songs! 

Wobbly meadow pause buttons!


henpemaz (approved):
Looks good, I noted down that there might be a detail regarding the entitydata readout that impacts both this and tagmod, I'll get to it soon

PR #528, opened 2024-12-09
client don't place quantified creatures

youbitchoc:


youbitchoc:
may fix weird stuff happening with leeches

henpemaz:
sounds about right, affects small-spiders too (the shaded darkness ones)

PR #529, opened 2024-12-09
change some noisy error loggers to trace instead

youbitchoc:
palette is spammy during nighttime, and canmove has a really chunky stacktrace


henpemaz:
Defo agree with the color one, it's no longer needed, but remote entities being moved by the client are still sources of problem and I'd rather keep visibility on that

youbitchoc:
> remote entities being moved by the client

it's mostly from seeing remote creatures enter shortcuts, i suppose we could mute shortcut canmove checks?

PR #530, opened 2024-12-09
init for custom blacklist

6fears7:
Still need to address resizble width of the remix container. If users keep typing they will eventually be blocked from adding more mods due to width.

### DIFF
--- a/Online/RainMeadowModManager.cs
+++ b/Online/RainMeadowModManager.cs
@@ -8,54 +8,49 @@ namespace RainMeadow
     public class RainMeadowModManager
     {
         public static string[] GetActiveMods()
-        { 
+        {
             return ModManager.ActiveMods
                 .Where(mod => Directory.Exists(Path.Combine(mod.path, "modify", "world"))
-                    || highImpactMods.Contains(mod.id)
-                    || cheatMods.Contains(mod.id))
+                    || blacklistMods.Contains(mod.id))
                 .Select(mod => mod.id)
                 .ToArray();
         }
-        
-        public static readonly string[] highImpactMods = {
-            "devtools",
-            "rwremix",
-            "moreslugcats",
-            "keepthatawayfromme",  // needs extra syncing to work
-            "no-damage-rng",
-        };
 
-        public static readonly string[] cheatMods = {
-            "maxi-mol.mousedrag",
-            "fyre.BeastMaster",
-            "slime-cubed.devconsole",
-            "zrydnoob.UnityExplorer",
-            "warp",
-            "presstopup",
-            "CandleSign.debugvisualizer",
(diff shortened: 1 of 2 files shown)


Online/RainMeadowModManager.cs
```diff
@@ -8,54 +8,49 @@ namespace RainMeadow
     public class RainMeadowModManager
     {
         public static string[] GetActiveMods()
-        { 
+        {
             return ModManager.ActiveMods
                 .Where(mod => Directory.Exists(Path.Combine(mod.path, "modify", "world"))
-                    || highImpactMods.Contains(mod.id)
-                    || cheatMods.Contains(mod.id))
+                    || blacklistMods.Contains(mod.id))
```
youbitchoc, line 14: this won't work

you have to sync the banlist and the require-list separately

Online/RainMeadowModManager.cs
```diff
@@ -8,54 +8,49 @@ namespace RainMeadow
...
+            return mods.Intersect(blacklistMods).ToArray();
+
+
+
         }
         internal static void CheckMods(string[] lobbyMods, string[] localMods)
         {
-            var lobbyCheatMods = GetCheatMods(lobbyMods);
-            var localCheatMods = GetCheatMods(localMods);
+            var lobbyCheatMods = GetBlackList(lobbyMods);
```
youbitchoc, line 52: this logic will have to be changed too, also rename the vars to make more sense

Online/RainMeadowModManager.cs
youbitchoc: handle overlaps by removing from banlist?

youbitchoc (changes requested):
logic won't work as-is, need separate banlist and require-list

Online/RainMeadowModManager.cs
6fears7: Tested by having host type in "rwremix" into blacklist. Client had remix enabled and was prevented from joining. Not sure what I'm missing from what you're saying.

Online/RainMeadowModManager.cs
6fears7: thanks

Online/RainMeadowModManager.cs
youbitchoc: the banlist is for mods that clients should disable
the requirelist is for mods that clients should enable

these two should not be lumped together

youbitchoc: yes, and this will require redoing a lot of logic

youbitchoc: > require (beyond "modify", "region")

world and enum-changing ones, order matters

PR #532, opened 2024-12-09
fix item dupe for eggbug, vulture, mother spider

youbitchoc:
hacky, should use IL hooks to disable clientside spawns

### DIFF
--- a/Game/RainMeadow.CreatureHooks.cs
+++ b/Game/RainMeadow.CreatureHooks.cs
@@ -32,6 +32,10 @@ namespace RainMeadow
 
             On.EggBugGraphics.Update += EggBugGraphics_Update;
             On.BigSpiderGraphics.Update += BigSpiderGraphics_Update;
+
+            On.EggBug.DropEggs += EggBug_DropEggs;
+            On.Vulture.DropMask += Vulture_DropMask;
+            On.BigSpider.BabyPuff += BigSpider_BabyPuff;
         }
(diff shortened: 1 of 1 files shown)


Game/RainMeadow.CreatureHooks.cs
```diff
@@ -202,5 +206,36 @@ private void TentaclePlantAI_Update(On.TentaclePlantAI.orig_Update orig, Tentacl
             orig(self);
             if (!self.creature.IsLocal()) self.mostInterestingItem = mostInterestingItem;
         }
+
+        // HACK: doesn't play sounds, we should IL hook to disable just the eggs
+        private void EggBug_DropEggs(On.EggBug.orig_DropEggs orig, EggBug self)
+        {
+            if (!self.IsLocal())
```
6fears7, line 213: Because we're checking OPOs won't we need a game mode check?

youbitchoc: if it's a remote creature then remote will be in charge of simulating it, gamemode doesn't change this

6fears7: Is it worth wrapping it in a game mode check (or at least it online mode is not null) so we don't run operations period in single player story?

youbitchoc: IsLocal returns true if offline

PR #536, opened 2024-12-09
gives host extra control over when the cycle should restart

6fears7:
https://github.com/user-attachments/assets/4aca7bfc-1897-4cb0-9948-0bf3a411e608

### DIFF
--- a/GameModes/StoryGameMode.cs
+++ b/GameModes/StoryGameMode.cs
@@ -26,6 +26,7 @@ namespace RainMeadow
 
         public string? myLastDenPos = null;
         public bool hasSheltered = false;
+        public bool wantsToRestartCycle = false;
 
         public void Sanitize()
         {
@@ -39,6 +40,7 @@ namespace RainMeadow
             region = null;
             saveStateString = null;
             storyClientData?.Sanitize();
+            wantsToRestartCycle = false; 
         }
 
         public bool saveToDisk = false;
--- a/GameModes/StoryGameMode.cs
+++ b/GameModes/StoryGameMode.cs
@@ -26,7 +26,6 @@ namespace RainMeadow
 
         public string? myLastDenPos = null;
         public bool hasSheltered = false;
-        public bool wantsToRestartCycle = false;
 
         public void Sanitize()
         {
@@ -40,7 +39,7 @@ namespace RainMeadow
             region = null;
             saveStateString = null;
             storyClientData?.Sanitize();
-            wantsToRestartCycle = false; 
+
         }
 
         public bool saveToDisk = false;
--- a/GameModes/StoryGameMode.cs
+++ b/GameModes/StoryGameMode.cs
@@ -39,7 +39,6 @@ namespace RainMeadow
             region = null;
             saveStateString = null;
             storyClientData?.Sanitize();
-
         }
 
         public bool saveToDisk = false;
--- a/Story/StoryHooks.cs
+++ b/Story/StoryHooks.cs
@@ -26,6 +26,8 @@ namespace RainMeadow
             return false;
         }
 
+        public static bool showRestartButton = false;
+
         private void StoryHooks()
         {
             On.PlayerProgression.GetOrInitiateSaveState += PlayerProgression_GetOrInitiateSaveState;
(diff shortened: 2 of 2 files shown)


Story/StoryHooks.cs
```diff
@@ -89,8 +89,32 @@ private void StoryHooks()
...
+                }
+            }
+        }
+
+        private void PauseMenu_SpawnExitContinueButtons(On.Menu.PauseMenu.orig_SpawnExitContinueButtons orig, Menu.PauseMenu self)
+        {
+            orig(self);
+            if (isStoryMode(out var story) && story.wantsToRestartCycle)
+            {
+                var restartCycleButton = new Menu.SimpleButton(self, self.pages[0], self.Translate("Restart Cycle"), "ONLINERESTARTCYCLE", new Vector2(self.exitButton.pos.x - (self.continueButton.pos.x - self.exitButton.pos.x) - self.moveLeft - self.manager.rainWorld.options.SafeScreenOffset.x, Mathf.Max(self.manager.rainWorld.options.SafeScreenOffset.y, 15f)), new Vector2(110f, 30f));
```
youbitchoc, line 114: we can use a SimplerButton callback instead and avoid hooking PauseMenu_Singal

Story/StoryHooks.cs
```diff
@@ -249,6 +272,10 @@ private void TextPrompt_Update(On.HUD.TextPrompt.orig_Update orig, TextPrompt se
                     if (touchedInput)
                     {
                         self.gameOverMode = false;
+                        if (OnlineManager.lobby.isOwner)
+                        {
+                            story.wantsToRestartCycle = true;
```
youbitchoc, line 277: also this variable name is misleading, imo `showRestartButton` would be a better name

Story/StoryHooks.cs
youbitchoc: also name the button something shorter like "Restart"

GameModes/StoryGameMode.cs
```diff
@@ -26,6 +26,7 @@ public class StoryGameMode : OnlineGameMode
 
         public string? myLastDenPos = null;
         public bool hasSheltered = false;
+        public bool wantsToRestartCycle = false;
```
youbitchoc, line 29: do we _need_ to put this here? why not a static bool in StoryHooks?

youbitchoc (changes requested):

Story/StoryHooks.cs
```diff
@@ -89,9 +91,26 @@ private void StoryHooks()
...
+
+        private void PauseMenu_SpawnExitContinueButtons(On.Menu.PauseMenu.orig_SpawnExitContinueButtons orig, Menu.PauseMenu self)
+        {
+            orig(self);
+            if (isStoryMode(out var story) && showRestartButton)
+            {
+                var restartCycle = new SimplerButton(self, self.pages[0], self.Translate("Restart"), new Vector2(self.exitButton.pos.x - (self.continueButton.pos.x - self.exitButton.pos.x) - self.moveLeft - self.manager.rainWorld.options.SafeScreenOffset.x, Mathf.Max(self.manager.rainWorld.options.SafeScreenOffset.y, 15f)), new Vector2(110f, 30f));
+                restartCycle.OnClick += (_) =>
+                {
+                    showRestartButton = false;
```
youbitchoc, line 107: can't rely on it getting unset from clicking, we'll have to unset it at game start or something like that

youbitchoc: ... that won't get unset if you're revived though

imo fixing that is way too much hassle, thoughts on always adding the restart button for host? then we can get rid of `showRestartButton`

Story/StoryHooks.cs
```diff
@@ -89,9 +91,26 @@ private void StoryHooks()
...
         }
 
 
+
+        private void PauseMenu_SpawnExitContinueButtons(On.Menu.PauseMenu.orig_SpawnExitContinueButtons orig, Menu.PauseMenu self)
+        {
+            orig(self);
+            if (isStoryMode(out var story) && showRestartButton)
+            {
+                var restartCycle = new SimplerButton(self, self.pages[0], self.Translate("Restart"), new Vector2(self.exitButton.pos.x - (self.continueButton.pos.x - self.exitButton.pos.x) - self.moveLeft - self.manager.rainWorld.options.SafeScreenOffset.x, Mathf.Max(self.manager.rainWorld.options.SafeScreenOffset.y, 15f)), new Vector2(110f, 30f));
```
youbitchoc, line 104: also this variable should be named restartButton for consistency

youbitchoc (changes requested):

Story/StoryHooks.cs
6fears7: You don't have to twist my arm.

GameModes/StoryGameMode.cs
```diff
@@ -39,7 +39,6 @@ public void Sanitize()
             region = null;
             saveStateString = null;
             storyClientData?.Sanitize();
-
```
youbitchoc: 🙏

youbitchoc (dismissed):
lgtm

youbitchoc:
feel free to merge when tested

6fears7:
Tested, updated string to capitalization for consistency in button texts. Merging.

PR #542, opened 2024-12-10
Host controls voidspawn AddToRoom

6fears7:
- [x]  Needs testing


6fears7:
VoidSpawn infinite null reffs on BodyChunks when present:

[Client]

```
[Error  :RainMeadow] 23:33:08|68999|OnlineResource.State.ReadTo:Failed to read state to entity in Room SHSH_B10 : VoidSpawn #2697:apo:0001 from local:8720
[Error  :RainMeadow] 23:33:08|68999|OnlineResource.State.ReadTo:System.IndexOutOfRangeException: Index was outside the bounds of the array.
  at RainMeadow.RealizedPhysicalObjectState.ReadTo (RainMeadow.OnlineEntity onlineEntity) [0x00039] in C:\Users\6fear\source\repos\Rain-Meadow\Online\State\RealizedPhysicalObjectState.cs:32
```


### Relevant code:

https://github.com/henpemaz/Rain-Meadow/blob/main/Online/State/RealizedPhysicalObjectState.cs#L32


Issue is an out-of-range index error every packet that clogs network. VoidSpawns specifically have differing length of their bodyChunks compared to their chunkState, so I added a equivalency check to ensure the index mismatch doesn't happen. 

### Replicate Error
1. Go to Moon and get hasGlow
2. Go to SH_S07 or any room with VoidSpawns present, really. 
3. Observe client logs

Please send thoughts.

6fears7:
this will result in a "desync" of sorts with the void spawns. local player won't have what remote player has, but neither would overlap either.

youbitchoc:
```cs
public void GenerateBody()
{
	int num = Random.Range(3, Random.Range(3, 16));
	for (int i = 0; i < num; i++)
	{
		val.Add(new BodyChunk(this, num2, default(Vector2), num7, num7 * 0.1f));
	}
	base.bodyChunks = val.ToArray();
}
```

random state isn't synced so they can have different body lengths

can we just not sync the void spawn? add them to the shouldsyncapoin(room|world) blocklist

6fears7:
Oh sweet easy enough

Story/StoryHooks.cs
```diff
@@ -97,9 +97,22 @@ private void StoryHooks()
...
             On.Menu.SlugcatSelectMenu.GetChecked += SlugcatSelectMenu_GetChecked;
             On.Menu.PauseMenu.SpawnExitContinueButtons += PauseMenu_SpawnExitContinueButtons;
+
+            On.VoidSpawnKeeper.AddOneSpawn += VoidSpawnKeeper_AddOneSpawn;
         }
 
 
+        private void VoidSpawnKeeper_AddOneSpawn(On.VoidSpawnKeeper.orig_AddOneSpawn orig, VoidSpawnKeeper self)
+        {
+            if (isStoryMode(out var _) && !OnlineManager.lobby.isOwner)
```
youbitchoc, line 107: feels a bit silly to only allow host to spawn them

also this desynced bodychunk issue _will_ affect other gamemodes too so it doesn't make sense to gate it behind story

Online/State/RealizedPhysicalObjectState.cs
```diff
@@ -25,11 +25,14 @@ public virtual void ReadTo(OnlineEntity onlineEntity)
             if (onlineEntity.isPending) { RainMeadow.Trace($"not syncing {onlineEntity} because pending"); return; };
             var opo = onlineEntity as OnlinePhysicalObject;
             var po = opo.apo.realizedObject;
-            if (!opo.lenientPos)
+            if (!opo.lenientPos && chunkStates != null)
```
youbitchoc, line 28: why do we need a nullcheck?

Online/State/RealizedPhysicalObjectState.cs
```diff
@@ -25,11 +25,14 @@ public virtual void ReadTo(OnlineEntity onlineEntity)
...
             {
                 for (int i = 0; i < chunkStates.Length; i++) //sync bodychunk positions
                 {
-                    chunkStates[i].ReadTo(po.bodyChunks[i]);
+                    if (po.bodyChunks.Length == chunkStates.Length)
```
youbitchoc, line 32: do we *have* to check this every loop?

youbitchoc (changes requested):
imo we simply shouldn't sync voidspawn

6fears7:
I agree I'll update PR tomorrow

PR #556, opened 2024-12-13
Arena: Refactor PlayerResultsBox_ctor

6fears7:
Probably merge this after I get back and can test users quitting to ensure the results box functions correctly still.

- [ ] Needs Testing when player quits mid-game


6fears7:
Will re-open at a later date.

PR #558, opened 2024-12-14
Story: undo spectate shelter close

youbitchoc:
should fix bug with win deathscreen when host dead

just a visual nicety that doesn't even work consistently, we can do without it


6fears7 (approved):
While not convinced, approving in the event no other explanation can be determined due to 

1. Impact of change (visual)
2. proximity to release.

youbitchoc:
we can look into this next time

PR #562, opened 2024-12-15
OnlineResource: pendingEntities for early joiners

youbitchoc:
make EnterResource queue entities if resource is not yet active


henpemaz:
at a glance:

there's no point in queuing up entities entering rooms that are miles away from the player and might not be loaded ever, there's no mechanism for leaving the queue either so you'll be attempting to enter with entities that went through that room and left half an hour ago

PR #565, opened 2024-12-16
Extensions: CanMove fix faulty within-room check

youbitchoc:
I am so sorry


youbitchoc:
this affected apo.changerooms and apo/ac.move

henpemaz (approved):

PR #569, opened 2024-12-17
improve modlist checker

youbitchoc:
gonna squash (or at least rename) once i fix the loading popup to show


youbitchoc:
we can fix the loader itself later

youbitchoc:
final stretch

PR #570, opened 2024-12-17
Void Avatar removal

Zed9876:
Added player avatar removal on void fluid submersion, along with sfx and vfx.


6fears7 (commented):
And what happens when we go swimming? Does a player only see themselves and no hud arrows pointing to other players?

6fears7 (approved):
Approving conditionally on the basis you test the bejonkers out of this before merging

Zed9876:
Yes it removes the HUD icons for any players that step foot in the drink. (except for isMe)

PR #579, opened 2024-12-18
OnlineGameModeHelpers: reinitialize on modapply

youbitchoc:


henpemaz:
can we maybe do something this simple the normal way?

PR #586, opened 2024-12-19
Chat: cleanup

youbitchoc:
fixes crash when leaving lobby then joining lobby

also cleanup


henpemaz (approved):
as far as Online/Matchmaking/SteamMatchmakingManager.cs goes good to go, no clue about the rest

Game/RainMeadow.PlayerHooks.cs
6fears7, line 1: IIRC if there’s not null checks here Hunter game start will null ref sometimes.

6fears7 (approved):
Just added the comment for awareness prior to merging.

PR #590, opened 2024-12-20
ModLoader: enable

youbitchoc:
apparently it just works now


6fears7 (commented):
I guess we're just gonna roll the dice with the MSC enablement at this point in time?

6fears7:
@youbitchoc dunno what you wanna do about this PR

PR #591, opened 2024-12-20
Chat: readd StopPlayerMovement in Player.checkInput

youbitchoc:
whoops, turns out this was needed


6fears7 (approved):
Conditionally approving if Hunter campaign start is checked first. It needed some null checks previously and I'm not sure if just the hud / camera null checking is enough.

PR #596, opened 2024-12-21
Player & lobby count on Lobby menu

Quaxly:
A pretty basic player count that appears above the version number. There might be a better way of getting statistics that I don't know about so do lmk if thats the case.

image


6fears7:
LGTM, gonna pass to Henpe for review

6fears7 (approved):

PR #616, opened 2024-12-28
Story: don't place sporeplants, don't sync bees

youbitchoc:
fixes crash

also don't place hanging pearlstrings for now

### DIFF
--- a/GameModes/StoryGameMode.cs
+++ b/GameModes/StoryGameMode.cs
@@ -55,8 +55,15 @@ namespace RainMeadow
             return RainMeadow.Ext_ProcessID.StoryMenu;
         }
 
+        public HashSet<PlacedObject.Type> disallowedPlacedObjects = new()
+        {
+            PlacedObject.Type.SporePlant,  // crashes the game, ask Turtle
+            PlacedObject.Type.HangingPearls,  // duplicates and needs to be synced, ask choc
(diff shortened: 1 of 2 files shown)


GameModes/StoryGameMode.cs
```diff
@@ -55,8 +55,15 @@ public override ProcessManager.ProcessID MenuProcessId()
...
 
+        public HashSet<PlacedObject.Type> disallowedPlacedObjects = new()
+        {
+            PlacedObject.Type.SporePlant,  // crashes the game, ask Turtle
+            PlacedObject.Type.HangingPearls,  // duplicates and needs to be synced, ask choc
+        };
+
         public override bool AllowedInMode(PlacedObject item)
         {
+            if (disallowedPlacedObjects.Contains(item.type)) return false;
```
6fears7, line 66: I tried this in arena too, didn't seem to work. Just so you're aware. Was gonna investigate more.

youbitchoc: as in, it still places the object? or it doesn't fix the crash?

6fears7: For arena it still placed the object. I'm going to go test again rq just to verify tho.

6fears7 (approved):

PR #625, opened 2024-12-31
Add Local multiplayer

### LINKED ISSUE #599: Non Steam version
First of all, the mod looks amazing and is somewhat a dream come true.

Nevertheless, I admit I don't understand why steam is required for this to work (as described, it uses p2p, so I suppose steam is only used to relieve the "matchmaking" part). As installing mod in RW is as simple as dropping a folder, wouldn't be possible to support non-steam version (with the GOG version). It would be nice to not have this mod under the walled garden of Steam

Once again, amazing mod :)
have a nice day

invalidunits:
- [x]  Abstractify NetIO, Create a SteamNetIO class
- [x] Remove references to steam in non steam classes
- [x] reImplement LocalNetIO as a Base class for SteamNetIO (for local play on steam)
- [x] Add a Direct IP connection menu.
- [x] Add a proper lan matchmaker

Closes #599


henpemaz:
Feel free to reengineer the matchmaker interface if it gets in the way. It doesn't have much intent behind it I was just trying to get steam and the local-udp builds to cohexist.

At the time felt like inheritance over ifdefs would help since it keeps symbols visible for refactoring renaming etc. But then for the reads/sends I went the lazy route. Still not sure which is better.

6fears7:
Mark as ready for review when you think you are good to go. We will definitely want to get some playtesting for this as well.

invalidunits:
Since it finally works (through direct connection play), I'm opening the pull request for review. 
Though it can't be merged until some obvious bugs are addressed.

- [ ] Fix players not being able to join lobbies 
- [ ] Fix matchmaker so that lobbies appear.

invalidunits:
After doing some research on UDP hole punching. It seems like UDP would require the host to relay packets from clients to clients. As a result, If the host is disconnected the connection between everyone ends. 

Because of this, I would much rather have a different "Matchmaking Domain" for this process and limit the "Local domain" to only running supporting local network clients.

henpemaz (dismissed):
Looks great!

6fears7 (approved):

henpemaz:
@invalidunits @6fears7 merge checklist

- [ ] Steam play working
- [ ] LAN play working
- [ ] Local development (single pc multiple game instances) still possible and convenient to do

henpemaz (approved):

6fears7 (approved):
What a journey

bruglet:
I've been following this pull request pretty closely! Very exciting that we're getting close to it being merged! 🎉 

Once it is merged will you guys post compiled release somewhere in addition to the Steam workshop releases? (For people who own on GOG)

6fears7:
@bruglet yes our discord will have it available

henpemaz (approved):

6fears7:
@invalidunits  SEND IT

PR #635, opened 2025-01-06
handle spears in chunk on Destroy

6fears7:


Online/State/BodyChunkRef.cs
```diff
@@ -20,7 +20,16 @@ public BodyChunkRef(OnlinePhysicalObject owner, int index)
         {
             if (bodyChunk is null) return null;
             if (!OnlinePhysicalObject.map.TryGetValue(bodyChunk.owner.abstractPhysicalObject, out var oe))
-                throw new InvalidProgrammerException("body chunk owner doesn't exist in online space! " + bodyChunk.owner.abstractPhysicalObject);
+            {
+                for (int num = 0; num < bodyChunk.owner.abstractPhysicalObject.stuckObjects.Count; num++)
+                {
+                    if (bodyChunk.owner.abstractPhysicalObject.stuckObjects[num] is AbstractPhysicalObject.AbstractSpearStick && bodyChunk.owner.abstractPhysicalObject.stuckObjects[num].A.type == AbstractPhysicalObject.AbstractObjectType.Spear && bodyChunk.owner.abstractPhysicalObject.stuckObjects[num].A.realizedObject != null)
+                    {
+                        (bodyChunk.owner.abstractPhysicalObject.stuckObjects[num].A.realizedObject as Spear).ChangeMode(Weapon.Mode.Free);
```
youbitchoc, line 28: lillypuck can be stuck as well, consider casting as Weapon instead of Spear and removing the above objecttype checks

also consider using linq instead, performance impact is negligible (this isn't a hot loop)

youbitchoc (changes requested):

Online/State/BodyChunkRef.cs
```diff
@@ -21,13 +22,12 @@ public BodyChunkRef(OnlinePhysicalObject owner, int index)
...
-                {
-                    if (bodyChunk.owner.abstractPhysicalObject.stuckObjects[num] is AbstractPhysicalObject.AbstractSpearStick && bodyChunk.owner.abstractPhysicalObject.stuckObjects[num].A.type == AbstractPhysicalObject.AbstractObjectType.Spear && bodyChunk.owner.abstractPhysicalObject.stuckObjects[num].A.realizedObject != null)
-                    {
-                        (bodyChunk.owner.abstractPhysicalObject.stuckObjects[num].A.realizedObject as Spear).ChangeMode(Weapon.Mode.Free);
-                    }
-                }
+                bodyChunk.owner.abstractPhysicalObject.stuckObjects
+                    .OfType<AbstractPhysicalObject.AbstractObjectStick>()
+                    .Where(stick => stick.A.realizedObject != null)
+                    .ToList()
```
youbitchoc, line 28: not necessary, it's already an enumerable

Online/State/BodyChunkRef.cs
```diff
@@ -21,13 +22,12 @@ public BodyChunkRef(OnlinePhysicalObject owner, int index)
...
+                    .ForEach(stick => (stick.A.realizedObject as Weapon)?.ChangeMode(Weapon.Mode.Free));
```
youbitchoc, line 29: was thinking more of a normal `foreach (...) { }` but this works too

youbitchoc:
C# really needs a SelectNotNull

youbitchoc (dismissed):

henpemaz:
What's the reasoning behind this one?

On `opo.leaveimpl` there's a call to `apo.LoseAllStuckObjects();` which will clear out `apo.stuckobjects` so the list you're iterating on in your handling will always be empty.

Maybe this handling could be there in `leaveimpl` instead?

6fears7:
> What's the reasoning behind this one?
> 
> On `opo.leaveimpl` there's a call to `apo.LoseAllStuckObjects();` which will clear out `apo.stuckobjects` so the list you're iterating on in your handling will always be empty.
> 
> Maybe this handling could be there in `leaveimpl` instead?

When a body chunk is stuck with an object, I believe that the bodychunk ref state is running first and because the chunk is no longer present we throw exceptions. Spears will enter an errored state because they are still trying to stick into that null chunk, so setting the mode alleviates the issue. 

I'll test these changes in leaveimpl

6fears7:
Updated.

henpemaz (approved):

PR #639, opened 2025-01-11
Arena: stop RPCs from null reffing if not in game, add more UI elements

6fears7:
- [x] Test Locally
- [x] Test Online


6fears7:
image

PR #642, opened 2025-01-14
Plop, Music and UI Update

Intikus:
Plopmachine volume is now set by rain world's music volume instead of sound effect volume 
Adjusted the volume of custom songs and plopmachine to not be overly loud.
Songs no longer jolt on start
Can now play region bump songs
Will now update song data of clients without music players
Songs no longer start playing (audibly) in hubs.
No longer spams join noise
Being in a group no longer stops playing new music 
Hastened and polished meadows pause menu animations
Added flashing animation for new emotes unlocked
Unstuck now teleports you onto the closest shortcut

### DIFF
--- a/Online/Resource/OnlineResource.cs
+++ b/Online/Resource/OnlineResource.cs
@@ -99,6 +99,8 @@ namespace RainMeadow
             isWaitingForState = true;
         }
 
+        public static event Action<OnlineResource>? OnAvailable;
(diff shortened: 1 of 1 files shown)


Intikus:
Dammit I just wanted to save it. Ah, whatever. Still got a two three things i wanna do before actually pushing

PR #645, opened 2025-01-18
Add rainmeadow.json system

forthfora:
Allows other mods to define a `rainmeadow.json`file in their root directory, to tell meadow important information about the mod and it's requirements

At the moment, there are 2 keys:
`sync_required_mods`: list of mod ids that are required to be enabled on all connected clients if they are enabled on the host
`banned_online_mods`: list of mod ids that are banned from meadow game modes

Also allows a user defined `rainmeadow.json` to be created in StreamingAssets, functions the same as if the file was defined in a mod

The file is loaded from each mod and the results merged at the end without duplicates, at the moment this occurs in `ModManager.RefreshModLists`


henpemaz:
Give me a chance to look at these closer later today but at a glance might make sense to unify both the nomenclature and behavior with "high impact mods" (required to match host) and "cheat mods" (I think?) (can only be present on host or if host has it)

ModManager/RainMeadowModInfoManager.cs
```diff
@@ -0,0 +1,173 @@
...
+            if (rainMeadowModInfo is null)
+            {
+                throw new Exception("Deserializer returned null.");
+            }
+
+            return rainMeadowModInfo;
+        }
+        catch (Exception e)
+        {
+            RainMeadow.Debug($"Error loading Meadow mod info:\n{e}\n{e.StackTrace}"); // TODO: use a proper debug error log
```
forthfora, line 136: please lmk the proper error log method, just went for the first thing I found

ModManager/RainMeadowModInfoManager.cs
```diff
@@ -0,0 +1,173 @@
...
+
+            ModInfos[mod.id] = generatedModInfo;
+        }
+
+        LoadUserDefinedModInfo();
+
+        RefreshMergedModInfo();
+    }
+
+    internal static void RefreshUserDefinedModInfo() // TODO: call it when a lobby is loaded maybe
```
forthfora, line 58: let me know, a good place this should be called, if it's needed at all

(mainly here so when a user wants to change their settings they don't have to restart the game every time)

ModManager/RainMeadowModManager.cs
```diff
@@ -0,0 +1,99 @@
...
+using System.IO;
+using System.Linq;
+
+namespace RainMeadow;
+
+public static class RainMeadowModManager
+{
+    public static string[] GetRequiredMods()
+    {
+        var requiredMods = RainMeadowModInfoManager.MergedModInfo.ClientRequiredMods;
```
forthfora, line 11: most of the file is unchanged, only here

ModManager/RainMeadowModManager.cs
```diff
@@ -0,0 +1,99 @@
...
+        return ModManager.ActiveMods
+            .Where(mod => requiredMods.Contains(mod.id)
+                          || Directory.Exists(Path.Combine(mod.path, "modify", "world")))
+            .Select(mod => mod.id)
+            .ToArray();
+    }
+
+    public static string[] GetBannedMods()
+    {
+        var highImpactMods = RainMeadowModInfoManager.MergedModInfo.ClientRequiredMods;
```
forthfora, line 22: and here

ModManager/RainMeadowModManager.cs
```diff
@@ -0,0 +1,99 @@
...
+        var bannedMods = RainMeadowModInfoManager.MergedModInfo.OnlineBannedMods;
```
forthfora, line 23: and these 2 lines here

Online/RainMeadowModManager.cs
youbitchoc, line 1: git mv?

.gitignore
```diff
@@ -396,6 +396,7 @@ FodyWeavers.xsd
 
 # JetBrains Rider
 *.sln.iml
+.idea/
```
forthfora, line 399: gets rid of Rider's meta file

ModManager/RainMeadowModInfoManager.cs
youbitchoc: we can put it in the remix settings, though not too important for this PR

Online/RainMeadowModManager.cs
forthfora: ah, wasn't aware I could do that, let me fix it up

youbitchoc:
> unify "high impact mods" and "cheat mods"

agreed, the split was a relic from past design choices

although would be nice to separate them with a newline in the rainmeadow.json for users trying to edit manually

ModManager/RainMeadowModInfoManager.cs
forthfora: nvm, just noticed RainMeadow.instance so I'll use RainMeadow.instance.LogError

youbitchoc: RainMeadow.Error

forthfora:
name change suggestions are welcome for both `online_banned_mods` and `client_required_mods`, was trying to make them self descriptive without being too verbose

ModManager/RainMeadowModInfoManager.cs
forthfora: ty, fixing now

youbitchoc:
i'm happy with anything but "required", would be weird to have cheat mods in a "required" list

HighImpact sounds good

ModManager/RainMeadowModInfoManager.cs
forthfora: got it, will add a call there

ModManager/RainMeadowModInfoManager.cs
```diff
@@ -0,0 +1,173 @@
...
+    // Generates a new mod info class, automatically adding any mandatory tags (e.g. if it modifies regions, then it needs a mod info with itself added to the required_mods list)
+    private static RainMeadowModInfo GetGeneratedModInfo(ModManager.Mod mod)
+    {
+        var modInfo = new RainMeadowModInfo();
+
+        if (mod.modifiesRegions)
+        {
+            modInfo.ClientRequiredMods.Add(mod.id);
+        }
+
```
forthfora, line 110: ExtEnum detection can go here if it ever gets added

forthfora:
> EDIT: referring to unifying both lists
> 
> i'm happy with anything but "required", would be weird to have cheat mods in a "required" list
> 
> HighImpact sounds good

my only gripe with `high_impact` is that it's not immediately clear what it means exactly
to me it sounded like it was referring to something to do with performance at first

maybe something along the lines of `if_server_has_client_must_have` (though less verbose)
but yeah it's a minor issue, I'll wait for more discussion

henpemaz:
> if_server_has_client_must_have

It's more than that, if server doesn't have, clients aren't allowed to have either, it's basically "these mods must be synced"

Because the serialization uses the indexes "that stuff ends up at" for lots and lots of stuff, anything that adds regions rooms enums onlinestates rpcs etc needs to end up in that list so it's matched on all sides.

forthfora:
ah, I understand now, maybe something like `sync_required_mods` or `sync_enforced_mods` then

forthfora:
just noting this down, what would we think of another 2 custom matching fields like `sync_required_mods_override` and `banned_online_mods_override`

they would remove those mods from their respective fields, mainly for use in the userdefined meadow.json in case someone wants to allow mods that would normally be banned e.g. dev console

6fears7:
Wondering if we can also include "banner_users" array as well so when we do our mod checks / user ban checks we can also support perma banned players by steam ID

forthfora:
> Wondering if we can also include "banned_users" array as well so when we do our mod checks / user ban checks we can also support perma banned players by steam ID

sounds good, the user defined json would also let people add their own list of banned users technically
(probably would want a more user-friendly way to do that in the future though)

ModManager/RainMeadowModInfo.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// Users that are banned from joining lobbies hosted on this device.
+    /// </summary>
+    [JsonProperty("banned_users")]
+    public List<string> BannedUsers { get; set; } = new();
+
+    /// <summary>
+    /// Users that are banned from joining lobbies hosted on this device.
+    /// </summary>
+    [JsonProperty("chat_filtered_words")]
+    public List<string> ChatFilteredWords { get; set; } = new();
```
henpemaz, line 42: I don't think these belong here.

From what I get you're trying to unify this system so it can be used by Meadow to set a baseline of mod rules, for mods to add more modrules, and as user-defined.

User-defined should not be a JSON unless it can be 100% edited in-game, most people struggle with it. Imo user-defined stuff should live in a separate txt file so that anyone can edit it.

henpemaz:
Imo the txt-based way for users to configure required/banned mods that is currently implemented is better than expecting average-joe to edit a json file, I don't think we should get rid of it until there's a more user-friendly way of editing those lists.

ModManager/RainMeadowModManager.cs
```diff
@@ -1,70 +1,48 @@
...
+
+            var requiredMods = modInfo.SyncRequiredMods.Except(modInfo.SyncRequiredModsOverride).ToList();
+
+            requiredMods = UpdateFromOrWriteToFile(SyncRequiredModsFileName, requiredMods);
 
             return ModManager.ActiveMods
-                .Where(mod => highImpactMods.Contains(mod.id)
-                    || Directory.Exists(Path.Combine(mod.path, "modify", "world")))
+                .Where(mod => requiredMods.Contains(mod.id)
+                              || Directory.Exists(Path.Combine(mod.path, "modify", "world"))) // TODO: check if this is still necessary, see GetGeneratedModInfo()
```
forthfora, line 29: (checking mod.modifiesRegion should technically be sufficient, I think)

ModManager/RainMeadowModManager.cs
```diff
@@ -1,70 +1,48 @@
...
-            if (File.Exists(path))
-            {
-                lines = File.ReadAllLines(path);
-            }
-            else
-            {
-                File.WriteAllLines(path, lines);
-            }
-        }
+        // TODO: possibly rename these
```
forthfora, line 10: With a syntax rework they could be renamed considering people would have to update their user defined files anyways

forthfora:
dropping the summary of the rework here, for reference:

the way it would work is: load and merge all the modinfo from the `rainmeadow.json` across all mods, then create and write to `meadow-bannedmods.txt` and `meadow-highimpactmods.txt` if they don't exist

if they do exist, read from them, write new entries from the modinfos that don't exist, and use a syntax to determine whether a mod has been removed from the list on purpose

for instance in `meadow-bannedmods.txt`:
```
maxi-mol.mousedrag
fyre.BeastMaster
//slime-cubed.devconsole
zrydnoob.UnityExplorer
//warp
```
would indicate warp and dev console are excluded, not sure what the best way to indicate this to the average user would be
something like `warp - true` would also work

it's not as intuitive as just removing it from the file, but it's not an option in this implementation considering the contents from the modinfos need to be added and somehow check that they aren't being disabled on purpose

forthfora:
Added comments to the user-defined file parsing system - leading comments indicate a mod is disabled specifically, but you may also add comments after mod ids or on their own line which will be ignored when parsing

6fears7:
@youbitchoc do you have any additional thoughts on this?

forthfora:
another thought: can auto generate a comment at the top of each file to explain how it works and how to use it

henpemaz:
Hi, can you limit yourself to just adding rainmeadow.json for mods and not reworking the systems that are already in place in this PR? I want to get it through we can discuss further changes later

forthfora:
Stashed all of the reworks, this is just the isolated rainmeadow.json system now, should have no effect on how meadow currently behaves

6fears7 (approved):

6fears7:
Post stash LGTM, just needs the other approvals

henpemaz (approved):

PR #646, opened 2025-01-18
SlugcatCustomization: sync tertiary colors

youbitchoc:
sync colours as array, closer to how miscProgressionData.colorChoices handles it

potentially improves slugbase mod compat

### DIFF
--- a/Story/StoryOnlineMenu.cs
+++ b/Story/StoryOnlineMenu.cs
@@ -3,6 +3,7 @@ using Menu;
 using Steamworks;
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using UnityEngine;
 
@@ -43,6 +44,9 @@ namespace RainMeadow
                 SetupClientOptions();
             }
 
+            // HACK: force-register MMFEnums.SliderID because god is dead and we have killed them
+            MoreSlugcats.MMFEnums.SliderID.RegisterValues();
+
             SetupOnlineCustomization();
 
             SetupOnlineMenuItems();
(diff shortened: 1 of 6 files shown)


6fears7 (approved):
This whole thing fills me with agony. Approved, assuming tested:

- Remix off (no null refs)
- standard (looked good from dev post)
- arena

youbitchoc:
remix works fine even when hot-enabling and disabling

as for arena i added body/eyeColor compat so _should_ be fine but yet to test

6fears7:
I'll test quickly tomorrow

6fears7:
Errors in Arena:

```
[Info   :RainMeadow] Creating player graphics! 0
[Error  :RainMeadow] System.NullReferenceException: Object reference not set to an instance of an object
  at (wrapper dynamic-method) PlayerGraphics.DMD<PlayerGraphics::CustomColorSafety>(int)
  at (wrapper dynamic-method) MonoMod.Utils.DynamicMethodDefinition.Trampoline<PlayerGraphics::CustomColorSafety>?298349568(int)
  at (wrapper delegate-invoke) <Module>.invoke_Color_int(int)
  at RainMeadow.RainMeadow.PlayerGraphics_CustomColorSafety_SlugcatCustomization (On.PlayerGraphics+orig_CustomColorSafety orig, System.Int32 staticColorIndex) [0x00035] in C:\Users\6fear\source\repos\Rain-Meadow\Game\RainMeadow.CustomizationHooks.cs:137
```

6fears7 (dismissed):
Tested offline/online across modes

Story/StoryOnlineMenu.cs
```diff
@@ -43,6 +44,9 @@ public StoryOnlineMenu(ProcessManager manager) : base(manager)
                 SetupClientOptions();
             }
 
+            // HACK: force-register MMFEnums.SliderID because god is dead and we have killed them
+            MoreSlugcats.MMFEnums.SliderID.RegisterValues();
+
```
henpemaz, line 49: not ideal. Imo just make Remix (mmf) a requirement to have custom colors, which I think is already the case no?

henpemaz:
a bit conflicted about the whole bypassing remix-custom-colors thing, but I do get it from the UX perspective. Having no colors online sucks, and requiring the user to go tweak a setting for something that's borderline required is bad UX.

how hard would it be to just pop your own OPTinyColorPicker in there instead if no custom-colors available?

henpemaz (approved):
Approved as far as onlinestate stuff goes for sure.

Still the above concerns, but feel free to roll with it.

6fears7:
> a bit conflicted about the whole bypassing remix-custom-colors thing, but I do get it from the UX perspective. Having no colors online sucks, and requiring the user to go tweak a setting for something that's borderline required is bad UX.
> 
> 
> 
> how hard would it be to just pop your own OPTinyColorPicker in there instead if no custom-colors available?

May not be feasible. OPTinyColorPicker depends on SmartMenu, which SlugcatSelectMenu is not. Would also be a strange UX to have two kinds of color pickers depending on remix presence. 

Overall, trade-offs any direction you look. 

@youbitchoc please review and merge or change at your discretion.

6fears7 (approved):

PR #649, opened 2025-01-19
custom colors in arena

6fears7:
convert list of custom color to string so remix can store the value and reuse across modes


youbitchoc (changes requested):

youbitchoc:
is the problem that arena expects customColors to contain tertiary colours?

PR #650, opened 2025-01-19
Arena: fix monk hits at range

6fears7:
- Fixes monk kills at a range
- Does not fix HitBy not running when close to another slug


henpemaz:
Please at least add the "spear" data group tag so it plays nice with how the delta system groups changes.

Ideally also either use OnlineFieldHalf for a 2-bytes-encoded float, or figure out a way to squeeze it even further into a single byte?

6fears7:
You got it

6fears7:
Updated

henpemaz (approved):

PR #640, opened 2025-01-12
Chat Updates

Quaxly:
Added experimental death messages. They currently display PVP & PVE related deaths and any death where Creature.Die() is called on a player.
Chat will now stay toggled across arena/lobby switches. 
System messages will show up in yellow for easier readability

RainWorld_WxyhZBUVW0

PR #651, opened 2025-01-20
6fears7 chat updates from Quaxly

6fears7:
- Suggesting to close #640  in favor of these changes:
- [x] Moved OnlineUI out of story folder
- [x] Added death events for hitting ground and swimming in acid/lava
- [x] Moved hook from comp to general arena
- [x] Used native GetOnlineObject func


OnlineUIComponents/DeathMessage.cs
```diff
@@ -0,0 +1,171 @@
...
+                    break;
+                case DeathType.Abyss:
+                    ChatLogManager.LogMessage("", $"{t} fell into the abyss.");
+                    break;
+                case DeathType.Drown:
+                    if (player.grabbedBy.Count > 0)
+                    {
+                        ChatLogManager.LogMessage("", $"{t} was drowned by {player.grabbedBy[0].grabber.Template.name}.");
+                    }
+                    ChatLogManager.LogMessage("", $"{t} drowned.");
```
SaddestUnderscore, line 33: wouldn't this cause 2 messages to fire if it met the condition

henpemaz (approved):
I'm the code-owner of the new folder by default, I'll update the codeowners file soon, blindly approved for now.

OnlineUIComponents/ChatHud.cs
```diff
@@ -13,8 +13,8 @@ public class ChatHud : HudPart
         private ChatLogOverlay? chatLogOverlay;
         private ChatInputOverlay? chatInputOverlay;
         public bool chatInputActive => chatInputOverlay is not null;
-        private bool showChatLog = false;
-        
+        public bool showChatLog = RainMeadow.rainMeadowOptions.ChatLogOnOff.Value;
```
youbitchoc, line 16: consider making this a property that sets the configurable Value directly so users can toggle persist in-game as well

henpemaz (approved):
lgtm

PR #653, opened 2025-01-23
Chat Enhancement 1.1

SaddestUnderscore:
- Brought back changing the enter key bind, original reason for the removal was it would mess stuff up if you had a different bind that's not enter, it's handled now with events
- Slight changes to chat tutorial, hopefully to get people to look down and read the text with the darker bool being set to true
- Brighten dark usernames for chat log
- Fixed message positioning with using labeltest getwidth as opposed to multiplying username length by five
- Added a cursor, currently it's just for visual purposes
- New parent object for ChatTextBox, allowing for text alignment being set to left instead of middle


SaddestUnderscore:
tested successfully

6fears7 (approved):

PR #658, opened 2025-01-24
Mod Filtering for Lobbies

TheLazyCowboy1:
**Adds a lobby filter that allows for three things:**
1. Filtering for lobbies with "Any" mods (no effect; allows all lobbies)
2. Filtering for lobbies with "All" of your currently installed highImpact mods (only reveals lobbies with an identical mod list and load order; although non-relevant mods are ignored)
3. Filtering for specific highImpact mods that you have installed

It works. It was easy to implement. The only objectionable things might be:
* ~~The less-than-beautiful code implementation (but honestly, it doesn't look much worse than the surrounding code)~~
    * Edit: Added functions that make the code look a little cleaner.
* The placement and appearance of the filter in the menu
    * Edit: I can't figure out how to get "Any" to be the first option. It's always "All".
* Does "Lobby Mods" need to be translated??? I don't know how to set up the translation data; and I certainly can't translate anything myself.
* ~~The way "All" works. Perhaps we prefer it to ignore load order?~~
    * Edit: Made "All" check for whether the client has all the lobby's required mods installed.
    * Edit: Added "Exact Order" that checks if the mod load order is the same.
* Objections to mod filters in general

**Why implement a mods filter?**
As a modder trying to make mods for this game, I want people to be able to easily use my mods in Rain Meadow, and I want to be able to play with my mods with others.
* Filtering mods makes it easier for players to find compatible lobbies (without having to join, leave, change mod list, rejoin, etc.).
    * Thus, mod filters will save time and ease the use of Rain Meadow.
* Filtering for specific mods allows players to find people to play specific mods with (other than arranging through Discord or something).
    * If I make a lobby with my Region Randomizer mod, several people will join it and have to leave, and people who DO want to play with the randomizer won't know that there's an available lobby for it!
* Filters will help players who do not have the Downpour DLC to find compatible lobbies.

...a "copy mod list" button would also be useful, but that's probably an over-kill feature.


6fears7:
@youbitchoc any thoughts on this?

henpemaz:
Interesting, I thought a "DLC required" indicator would precede this.

Here's the use-cases I see for mod-filtering:
- I want to see if there's any lobbies using mod X (and Y and Z) because I want to play them (other mods are fine)
- I want to see only lobbies that do not have X (nor Y nor Z) because I don't want to play them (or I don't own the DLC!!!) (other mods are fine)

It's honestly not a lot of use-cases, sure a complex filter covers them, but it might make sense to try and streamline the two above.

I'm on the road most of the day today but I can test what you got going later.

youbitchoc:
agree with henpe, not too sure about "Exact"

Online/RainMeadowModManager.cs
```diff
@@ -36,6 +37,26 @@ public static string[] GetRequiredMods()
...
+            return ModManager.ActiveMods
+                .Where(mod => highImpactMods.Contains(mod.id)
+                    || Directory.Exists(Path.Combine(mod.path, "modify", "world")))
+                .Select(mod => mod.name)
+                .ToArray();
+        }
+
+        public static string RequiredModsArrayToString(string[] requiredMods)
+        {
+            return string.Join(" ; ", requiredMods);
```
youbitchoc, line 53: I'd use newlines

youbitchoc: then regex won't be required and you can just string.split

Online/RainMeadowModManager.cs
```diff
@@ -36,6 +37,26 @@ public static string[] GetRequiredMods()
                 .ToArray();
         }
 
+        public static string[] GetRequiredModNames()
```
youbitchoc, line 40: please don't duplicate the requirement logic, I'd rather you look up the full name from the ID

youbitchoc:
previewing enabled/disabled mods would be nice to have but can be another PR

Menu/LobbySelectMenu.cs
```diff
@@ -123,6 +125,18 @@ public LobbySelectMenu(ProcessManager manager) : base(manager, RainMeadow.Ext_Pr
             filterLobbyLimit.OnChange += UpdateLobbyFilter;
             new UIelementWrapper(this.tabWrapper, filterLobbyLimit);
 
+            where.y -= 30;
+            var filterModsLabel = new ProperlyAlignedMenuLabel(this, mainPage, Translate("Lobby Mods"), where, new Vector2(200f, 20f), false);
+            mainPage.subObjects.Add(filterModsLabel);
+            where.y -= 27;
+            string[] requiredModIDs = (new string[3] { "Any", "All", "Exact" }).Concat(RainMeadowModManager.GetRequiredMods()).ToArray();
+            string[] requiredModNames = (new string[3] { "Any", "All", "Exact Order" }).Concat(RainMeadowModManager.GetRequiredModNames()).ToArray();
```
youbitchoc, line 133: consider [Enumerable.Zip](https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.zip?view=net-9.0)

youbitchoc: also I'd rather add the custom options separately to avoid confusion

so initialize requiredModsList with Any/All/Exact then append the required mods

Menu/Objects/LobbyCardsSelector.cs
```diff
@@ -255,13 +257,32 @@ public void FilterLobbies()
...
+
         foreach (var lobby in allLobbies)
         {
             if (filter.lobbyName != "" && !lobby.name.ToLower().Contains(filter.lobbyName)) continue;
             if (filter.enabled)
             {
                 if (filter.gameMode != "All" && lobby.mode != filter.gameMode) continue;
                 if (filter.publicLobby && lobby.hasPassword) continue;
+                //filter for required mods
+                if (filter.requiredMods != "Any") {
```
youbitchoc, line 271: consider switch case to make it clear that requiredMods can be the actual mod name

youbitchoc (dismissed):

TheLazyCowboy1:
> agree with henpe, not too sure about "Exact"

Yeah, I only added the "Exact Order" option because it was my original (poorly thought out) implementation of "All". And in descriptions, y'all make such an emphasis on "make sure the load order matches exactly" etc. But personally, I think "Exact Order" is just clutter.

TheLazyCowboy1:
One of my greatest concerns is not making the UI cluttered. And are the options understandable to most people? They probably need to be renamed.

And random note: it would be easy to add an indicator on lobby cards for whether your current mod list is compatible. Is this desireable?

TheLazyCowboy1:
Sorry that I keep yapping, but...

I'm concerned about the naming of "Any" and "All". Currently for filtering Game Modes, the term "All" is used to mean "the filter allows lobbies with any game mode; that is, the filter doesn't apply." (Personally, I think the term "Any" would be better suited for filtering Game Modes, but that's a pretty unimportant issue.) However, in my filter, "Any" means "the filter doesn't apply / the filter allows any lobby" whereas "All" literally means "All of the below".

Ideas for "All" renaming:
* "Exact Match" (if we remove the "Exact Order" option)
* "All of the below" (the most exact name but sounds really weird and wordy)
* "Compatible mods only" (??? I'm running out of ideas, lol)

Online/RainMeadowModManager.cs
TheLazyCowboy1: > I'd use newlines

Oh my goodness... how on earth did I forget about the existence of the newline character. I use it all the time for delimiters. Thanks!

Online/RainMeadowModManager.cs
TheLazyCowboy1: > please don't duplicate the requirement logic, I'd rather you look up the full name from the ID

But... but... that'll take like an extra half-a-millisecond of processing time!!! ;) Yeah you're right; duplicating the logic was sloppy of me.

Menu/LobbySelectMenu.cs
TheLazyCowboy1: > consider Enumerable.Zip

Perhaps I'm totally misunderstanding the modern uses of enumerables, but I thought that enumerables weren't intended for lists of varying length?

Regardless, I don't think the built-in options for converting enumerables to drop-downs would work because I need the display names to be separate from the mod IDs.

> initialize requiredModsList with Any/All/Exact then append the required mods

Okay, I guess that'll be slightly more readable or something?

Menu/Objects/LobbyCardsSelector.cs
TheLazyCowboy1: So, like...
```
case "Any": break;
case "All": if (...) continue;
default: if(...) continue;
```
...I rarely use switch statements, but it seems to me like calling "continue" from within a switch function might try to apply to the switch rather than to the loop.

TheLazyCowboy1:
@youbitchoc I think I have made all the changes you requested.

Current state of the lobby filter:

Screenshot (34)

I have removed "Exact Order" from the list; although I have kept its functionality in the FilterLobbies() code.
I also renamed "All" to "All of the below" just to see how it looks... Um... what do y'all think?
I also fixed the order of the list. Wow, it's almost like specifying indices can be useful for things like that ;)

6fears7:
> One of my greatest concerns is not making the UI cluttered. And are the options understandable to most people? They probably need to be renamed.
> 
> 
> 
> And random note: it would be easy to add an indicator on lobby cards for whether your current mod list is compatible. Is this desireable?

The only indicator I really care about is DLC usage outside of any filter. 

If a pre-filtering mechanism exists, then I don't see a reason for an icon if a modlist is compatible because the filter would do the same thing behind the scenes. And "compatible" could be misleading

6fears7:
> Sorry that I keep yapping, but...
> 
> 
> 
> I'm concerned about the naming of "Any" and "All". Currently for filtering Game Modes, the term "All" is used to mean "the filter allows lobbies with any game mode; that is, the filter doesn't apply." (Personally, I think the term "Any" would be better suited for filtering Game Modes, but that's a pretty unimportant issue.) However, in my filter, "Any" means "the filter doesn't apply / the filter allows any lobby" whereas "All" literally means "All of the below".
> 
> 
> 
> Ideas for "All" renaming:
> 
> * "Exact Match" (if we remove the "Exact Order" option)
> 
> * "All of the below" (the most exact name but sounds really weird and wordy)
> 
> * "Compatible mods only" (??? I'm running out of ideas, lol)

When it comes to the decision tree, start with all lobbies first. Then by game mode, then by any other filters. That is our current setup in main. 

So if a mod filter is active, it should be dependent on the returned lobbies based on the higher filter, "game mode."

TheLazyCowboy1:
...and I know it'd go in a separate PR, but... do we want to have some indications on the lobby cards for whether they're compatible?
Like having a mark in the top-right corner of the lobby card:
* Green checkmark if the load order is the same
* Yellow circle (or no mark whatsoever) if the mods are compatible
* Red X if the mods are incompatible.

Or they could all be circles...?

Here's an example of what that could look like: (20-second copy/paste edit job in MS Paint3D:)
LobbyModCompatiblityIndicators
The main problem for me is: How will anyone know what the marks mean?? It might be a nice convenience, but only if people understand what it's indicating. I'd be afraid that people would think the red cycle means the lobby is currently unavailable or disconnected or something.

It'd probably be an excessive feature. The reason why I'm asking about it is because I could make a PR for it, but I just don't want to spend an hour doing that unless it's actually wanted.
Okay, hopefully I'm finally done yapping for the day. I talk too much.

6fears7:
> @youbitchoc I think I have made all the changes you requested.
> 
> 
> 
> Current state of the lobby filter:
> 
> 
> 
> Screenshot (34)
> 
> 
> 
> I have removed "Exact Order" from the list; although I have kept its functionality in the FilterLobbies() code.
> 
> I also renamed "All" to "All of the below" just to see how it looks... Um... what do y'all think?
> 
> I also fixed the order of the list. Wow, it's almost like specifying indices can be useful for things like that ;)

I would also like to hear @T1mbits insight on this for the UI. I'm not a fan of a drop-down single-select but we may not have a lot of other options.

T1mbits:
Sorry for the terrible structure here, I’m on my phone rn.

I’ve had some visions for a proper filter settings UI page, but it would require the creation of new UI elements which I don’t think anyone wants to dedicate their time to. As of now the single select drop down is all we have, I’ve fiddled a bit with a multi-select drop down and it does not work very well. The indicator is definitely good, and I think a proper dedicated space for maybe a scrollbox containing checkboxes would be good? I’m also working on something that affects this menu so I guess I’ll put that off until later.

TL;DR
Single drop down is an alright band-aid fix until we decide on a proper UI for filtering, and the indicators for DLC are good.

I’m open to working on a proper filtering UI now that I have a bit more free time, if you guys want to discuss a bit more on Discord. The mod filter is great, but I’m not quite sold on the placement.

T1mbits:
actually looking at the indicators and your concerns that may be an issue, I don’t think the menu system supports tooltips so it may be an excessive feature that could just be replaced with a popup stating the incompatible mods on an attempted lobby join

TheLazyCowboy1:
This T1mbits fellow is thinking rightly about these things.

> I’m open to working on a proper filtering UI now that I have a bit more free time, if you guys want to discuss a bit more on Discord. The mod filter is great, but I’m not quite sold on the placement.

@T1mbits It seems that you have some charge over the UI design. What changes would you like me to make to the mod filter?
The current placement is simply the result of copy/pasting the filtering code for the other 3 filters. The mod filter can easily be repositioned and resized.

>  I think a proper dedicated space for maybe a scrollbox containing checkboxes would be good? I’m also working on something that affects this menu so I guess I’ll put that off until later.

I'm really bad at UI design (go check the config menus on my mods, lol). I could try to make a scrollbox with checkboxes, but I can't guarantee the result would be pretty.
It seems to me that the current conclusion is: "Current filter alright; will move/replace it when the filtering system gets reworked"??

T1mbits:
I'm currently working on rewriting and abstracting some components of the UI in preparation of an Arena menu UI overhaul, which involves changing the current filtering system. I was responsible for the last lobby select menu overhaul as well as the filtering system so I feel a bit bad for leaving it in such a mess.

If you have time we can come up with a proper filtering system + UI and then merge this PR afterwards, or we can merge the PR now and fix the filtering later. Both are fine with me, but trying to implement scrollboxes are a nightmare (I speak from experience), so what you have right now is fine IMO.

TL;DR
my bad forgot other people would look at the filtering code O_o. Current PR is fine.

TheLazyCowboy1:
> If you have time we can come up with a proper filtering system + UI and then merge this PR afterwards

I really just don't know how much time we have until the next update. My personal top priority is just to have a functioning mod filter in the next update (0.1.2 or whatever it'll be).

@T1mbits But I'll absolutely help you with filters to the best of my extremely limited ability if you want to attempt a filter overhaul within the next few days. I am semi-competent at UI implementation, just not design.

(Random note: The filtering code was extremely easy for me to understand. I like code like that; it actually makes sense. It's too bad it's considered unprofessional or something like that. Who cares about standardsssss...)

T1mbits:
If the top priority is to just have it functioning and in main branch then we should be good to merge I think, haven't looked at your  implementation yet but I'm assuming it's passed checks from henpemaz, 6fears, and notchoc. The filtering code isn't "bad", but I just don't really like how it's laid out. I'm glad to hear my code is readable however :D. Not quite sure how else I'd go about doing it tho, so maybe it's as good as it gets.

Menu/Objects/LobbyCardsSelector.cs
youbitchoc: `continue` only affects loops, you may be thinking of `break`

TheLazyCowboy1: Well it's showing that I never learned C#... but I solved the issue anyway with my latest commit. I just used a bool flag, which turned out to be necessary regardless, so there's no resulting clutter or inefficiency.

Menu/LobbySelectMenu.cs
youbitchoc: yeah, Select() would be better now that we're deriving the name from the id

so something like `requiredmodslist.AddRange(ids.Select(id => new ListItem(id, namefromid(id)).ToList())`

6fears7:
@T1mbits Is there an opportunity to overlap a sprite on top of the card so we can implement MSC indicator?

T1mbits:
yeah, since the card is just a simpler button and it has access to LobbyInfo you can probably just add a sprite to the card (idk if there's a circular one but that'd be the best for an indicator) and change its colour with data from the LobbyInfo (I'm assuming that's where the mod data of a lobby is stored?)

TheLazyCowboy1:
As far as I know, this would be quite easy to implement. The hardest part would probably be choosing the sprite/color!But what exactly would we have symbols for? Just DLC (MSC and maybe someday Watcher)? In which case, I would suggest having a small "MSC" or "DLC" symbol (literally those letters are the symbol), since we can't have text descriptions when hovering over it (without some extra work...),  and those are the only two things that would be pretty self-explanatory.From: Timbits ***@***.***>Sent: Monday, February 10, 2025 2:02:05 PMTo: henpemaz/Rain-Meadow ***@***.***>Cc: TheLazyCowboy1 ***@***.***>; Author ***@***.***>Subject: Re: [henpemaz/Rain-Meadow] Mod Filtering for Lobbies (PR #658)
yeah, since the card is just a simpler button and it has access to LobbyInfo you can probably just add a sprite to the card (idk if there's a circular one but that'd be the best for an indicator) and change its colour with data from the LobbyInfo (I'm assuming that's where the mod data of a lobby is stored?)

TheLazyCowboy1:
And uh... there seem to be a lot of merge conflicts... GitHub won't let me see what the conflicts actually are, so I have no clue how to fix them. Do I have to just redo my PR but based on the current main branch instead?

T1mbits:
if you’re on your own fork just sync the fork’s working branch and resolve the conflicts within github

TheLazyCowboy1:
...That was an extremely painful hour. Why is GitHub so horrible... oh well. I think I did it.

TheLazyCowboy1:
Well I guess this PR is outdated. The sole point of it was to get in a mod filter in the 0.1.2 update. I guess I'll have to redo it for the UI update.

TheLazyCowboy1:
Just writing here for my own archival purposes:

* I'm trying to think about how I want to implement the filter for the next update.
* The main issue mentioned about this PR is that the devs don't want to have to maintain it.
* ...the problem is: This PR is pretty maximally simple and maintainable.

If I create an entire new filtering menu with a complex UI for a complex filtering system... That's going to be
1. Way more work for me, *(I'm lazy)*
2. Wayyyyyyyyy more work for the devs to maintain *(think of how many bugs I might accidentally cause (and fail to catch)!!! And every update (especially UI updates), they'll have to ensure that the menu doesn't break!)*, and
3. Significantly less convenient to use for the vast majority of players. *(More advanced filters does not equal greater ease of use. I am accustomed to databases in which the filtering system by default works basically this way. The current system allows players to quickly check for compatible lobbies or for one specific mod they're interested in. Why would anyone need to check for two or more mods? Even if so, they could easily cross-reference the lists (there will probably only be like 1 or 2 lobbies that'll meet the requirement anyway). The only potential use case is specifically excluding certain mods, like MSC?)*

...So honestly I'm really struggling to find anything lacking about this current PR other than the fact that it'd be nice if the dropdown was positioned like 35 pixels higher...? Yeah, that's the best change I can come up with: Shift all the filters 30 pixels higher. But then I'd be touching other folks' UI stuff.

Yeah now I'm confused, lol.

TheLazyCowboy1:
That is exactly the current functionality.

Are you saying that you dislike the term "Any"? The filter of "Any" means that "any" mod is allowed; that is, the filter is ignored. This is the default behavior.

Although for use case 1, would you prefer that it matches the exact load order, or just the mod list? I have both implemented, though I currently have the "Exact Match" (matches load order) option hidden.

6fears7:
1. Yeah swap "Any" to "No filter" since users are going to request they don't "want a filter applied" so this will mentally help them.
2. Let's offer them parity (1-1) and specific (chosen mod). Having the right mods in the wrong order is still a red flag for a number of content mods and will increase confusion when they join thinking "I have the same mods" but have extenum ordering issues
3. Total list of options:

- No Filter
- Exact Remix Match
- <mod>

TheLazyCowboy1:
Here is what it looks like now:
Screenshot (44)

"Unfiltered" felt like more exact language to me. Feel free to nitpick that choice if you want; it's a 5-second change.

However, in the current state of Rain Meadow, the vast majority of mods used do not implement ExtEnums or complex region changes, so their load order doesn't matter. I feel like load order should be enforced when ***joining*** a lobby, not ***searching*** for one. So I hid "All of the below" at the bottom, and of course renamed it to "All of the above".
Because quite simply: If I'm looking for someone using, say, MSC, Remix, and a randomizer mod, then I can change my load order if it's problematic, but I primarily want to be able to use the filter to see if someone's using those mods (in any order).

Random notes:
(It might be nice to distinguish mods that require similar load orders and those that don't. That's obviously a separate PR, and it's probably more work than it's worth, but it'd be a slight convenience.)
(I think I'm going to go make a client-side mod to copy a lobby's mod list using the Mod Presets mod...)

6fears7:
Works for me. Thinking "Exact match" may still be confusing to players as to *what* they're matching. Maybe "Exact order", , "Exact Remix Match", "Matches Mod Order", or something of the like. 

Also might be useful to have a character denoting "this is a selection not a mod name" eg.

"- Unfiltered"
"- Exact match"
"Rain World Remix"

so a user can quickly identify, "Ok this is a selection I can make and this is an actual mod."

TheLazyCowboy1:
Excellent suggestions. However...
Screenshot (45)
I feel like that just looks slightly bad. None of the other default filter options have a "- " in front, so it seems weird.
I tried putting a "-" in front of the mods themselves, but that made me think it was indicating that I was filtering for lobbies that did **not** have specific mods.
*(Then I spent several minutes excitedly implementing an option to both require and exclude mods before realizing there are half a dozen reasons why it can't be that simple.)*

Instead, I landed with this:
Screenshot (46)
It seems good to me.

6fears7:
Cool. I'll pull fork and just give it a once over and should be good

6fears7:
Updated "All of the above" to "Any order". Let me know if this is a blocker for you

TheLazyCowboy1:
That's totally fine. I just want a functional mod filter for my own use, lol.
I'm currently working on a PR to fix the mod applier code, btw... Who on earth wrote all of it and then just gave up and commented it out??

Online/Matchmaking/SteamMatchmakingManager.cs
```diff
@@ -8,8 +8,8 @@ namespace RainMeadow
 {
     public class SteamLobbyInfo : LobbyInfo {
         public CSteamID iD;
-        public SteamLobbyInfo(CSteamID id, string name, string mode, int playerCount, bool hasPassword, int? maxPlayerCount) : 
-            base(name, mode, playerCount, hasPassword, maxPlayerCount) {
+        public SteamLobbyInfo(CSteamID id, string name, string mode, int playerCount, bool hasPassword, int? maxPlayerCount, string highImpactMods = "") : 
+            base(name, mode, playerCount, hasPassword, maxPlayerCount, highImpactMods) {
```
6fears7, line 12: Need to set this up with LAN too, assuming LAN has similar structure

6fears7 (changes requested):

6fears7:
ModManager sometimes crashes when interacting with MSC, that's why it's commented out

Online/Matchmaking/SteamMatchmakingManager.cs
TheLazyCowboy1: LAN has an annoyingly different structure, but as far as I can tell it works just fine now (with the most recent commit).
Now, trying to join a LAN lobby that has a different mod list results in a crash, but I don't think that's due to the filter.

TheLazyCowboy1:
I hate GitHub.
But the filter works for LAN lobbies now. There's probably no point in it doing so, but hey, it wasn't *too* difficult to implement!
(Except for the hours spent fiddling with branches in GitHub. But that's GitHub's fault.)

6fears7 (dismissed):

henpemaz (approved):
Good to go if tested

henpemaz:
Also, on a similar vein would be the idea of having a [DLC] badge on the lobby cards for lobbies with the DLC on, there's some frustration when trying to join lobbies as one who doesn't own the dlc. But might also be interesting to wait for the new DLC to see if we might need to implement support for multiple badges.

henpemaz:
Mind solving the conflicts for a merge? `Online/Matchmaking/SteamMatchmakingManager.cs`

TheLazyCowboy1:
I won't be at home for several hours... I'm trying to do it on my phone, but... that's proving incredibly difficult.

TheLazyCowboy1:
Why on it did it "dismiss the reviews"? GitHub is ridiculous...

henpemaz (approved):
Good stuff! Keep it up!

PR #664, opened 2025-01-26
AbstractCreature: only owner calls extractKarma

youbitchoc:
try to fix scav corpses decaying rapidly

needs testing


6fears7:
Tested, no dice. Karma is desynced between players. 

PreGate: Player1 (holds scav): Karma Max
PreGate: Player2: Karma min

PostGate: Player1: (holds same scav) Karma min
PostGate: Player2: (holds scav): Karma Max, Player 1 then gets Karma min + 1.

PR #665, opened 2025-01-26
fixes null reff crash on restarting save

6fears7:


Story/StoryOnlineMenu.cs
```diff
@@ -92,7 +92,7 @@ public StoryOnlineMenu(ProcessManager manager) : base(manager)
                 personaSettings.customColors = val;
             }
             manager.arenaSitting = null;
-            if (restartChecked)
+            if (restartCheckbox?.Checked)
```
youbitchoc, line 95: is this 100% analogous?

6fears7: image

image

Yes + I think using the actual checkbox value versus this bool is more appropriate considering the original wackiness impl of that bool.

PR #667, opened 2025-01-27
Better Death Detection

Quaxly:
Added DeathContextualizer

Credit to EtiTheSpirit for this method

DeathContextualizer uses HarmonyX patches to find occurrences of Creature.Die and insert a call function above it to do a proper death event. This method should be much easier, cleaner and better for future proofing overall.

Player deaths from rain, centipedes, zap coils, worm grass, five pebbles, wall rot and the player's tongue contacting an active zap coil are now handed properly and have unique death messages.

image


6fears7:
Good stuff

henpemaz:
Sorry isn't harmony like obsolete? I'd appreciate if we stuck with monomod for hooks and ILs reather then mix and match with harmony.

Quaxly:
Gotchu, I'll revisit this in a few days when I got time. Switching it to ILHooks shouldn't be too challenging.

henpemaz (approved):
Looks good

PR #675, opened 2025-01-29
Rename Private visibility to Invite Only

T1mbits:
Limit confusion as giving a public lobby a password causes it to appear as "private" in the lobby select, whereas the Private visibility (now Invite Only) was just an invite only lobby

image


6fears7 (approved):
My approval means nothing but here: have some dopamine.

TheLazyCowboy1 (approved):

TheLazyCowboy1:
Absolutely yes!

...my feedback obviously has no effect on whether this gets merged, but just know that some random guy named TheLazyCowboy1 approves...

henpemaz (approved):

PR #677, opened 2025-01-31
6fears7 lan arena rpcs

6fears7:
For @invalidunits  to test


youbitchoc (changes requested):
just the arena stuff, not looking at lan impl

PR #681, opened 2025-02-07
Online: Fix indefinite tinnitus

6fears7:
handle cases where tinnitus continues after player is dead


youbitchoc:
is this a vanilla bug?

6fears7:
happening because we're not fully calling APO Destroy to prevent world release

PR #682, opened 2025-02-08
SlugBase customColors fix

forthfora:
At the moment this throws cause SlugBase assumes customColors is not null if CustomColorsEnabled returns true

Might be a better place to set this, lmk if so

### DIFF
--- a/Game/RainMeadow.CustomizationHooks.cs
+++ b/Game/RainMeadow.CustomizationHooks.cs
@@ -124,6 +124,8 @@ namespace RainMeadow
         {
             if (hackySlugcatCustomization is not null)
             {
+                // SlugBase will attempt to access customColors, assuming it is not null as CustomColorsEnabled is true
+                PlayerGraphics.customColors = hackySlugcatCustomization.customColors;
                 return true;
             }
             return orig();


forthfora:
btw from a quick test: new color system does sync custom SlugBase colors without any extra work which is great, does seem to have some quirks though (at least in StoryP2P):

- after changing color in the menu, it's only applied after entering then exiting
- colors start off desynced when a client joins, fixes itself after going through a shortcut

6fears7:
I noticed first bullet point too. We do set the color when the slide bar changes and we used to read from the remix value, it's possible we don't know following that change. Will have to look

6fears7 (changes requested):
Changes may be needed. Setting this to unapproved as a reminder to check color changes and why we might be needing to enter the game instead of just writing to remix 

- [ ] Address color desync before second realization (ie. Going into a pipe)
- [ ] Address color selection requiring going in-game (we should snag color data to write to remix in the menu if we can

- [x] Wish for a better process to manage colors (maybe a standalone settings page button accessible in lobby)

6fears7 (dismissed):
Changed my mind. Will open an issue for the aforementioned color problems

Game/RainMeadow.CustomizationHooks.cs
```diff
@@ -124,6 +124,8 @@ private bool PlayerGraphics_CustomColorsEnabled_SlugcatCustomization(On.PlayerGr
         {
             if (hackySlugcatCustomization is not null)
             {
+                // SlugBase will attempt to access customColors, assuming it is not null as CustomColorsEnabled is true
+                PlayerGraphics.customColors = hackySlugcatCustomization.customColors;
```
6fears7, line 128: If a user changes the color after setting the checkbox, won't player graphics custom colors be referencing outdated color info?

forthfora: The idea behind this placement is if somewhere is checking `CustomColorsEnabled` then it's about to use PlayerGraphics.customColors (this is what SlugBase does anyways)

likely a better place to put this though yeah, would need to look closer at the Meadow implementation or ask choc

forthfora: @youbitchoc just to bring this to your attention

forthfora:
[#](https://github.com/henpemaz/Rain-Meadow/blob/e5e2d6789eb13fd56fa7cdd187394c77c850dd53/Game/RainMeadow.CustomizationHooks.cs#L93)

^ wondering whether this is safe btw, or would we need a CWT?

noticed that colors were switching back and forth between players in testing, dunno if that's a side-effect of using StoryP2P on the same machine though

Game/RainMeadow.CustomizationHooks.cs
6fears7: choc might respond, but he's chosen to go on a sabbatical from the project. 

If you see a better spot for this lmk, otherwise we can go with it.

forthfora:
> > https://github.com/henpemaz/Rain-Meadow/blob/e5e2d6789eb13fd56fa7cdd187394c77c850dd53/Game/RainMeadow.CustomizationHooks.cs#L93
> > 
> > ^ wondering whether this is safe btw, or would we need a CWT?
> > noticed that colors were switching back and forth between players in testing, dunno if that's a side-effect of using StoryP2P on the same machine though
> > EDIT: realised that's hard to do when they are static... might need to think of something else
> 
> You were testing Pearlcat ( a Slugbase Slugcat) or vanilla?

This was with Pearlcat yes, might be a SlugBase color exclusive issue then?

forthfora:
Believe it's fixed, needed a hook to InitiateSprites, also rewrote the code that grabs colors from savedata as I suspected it originally but not sure that really made any difference

Will do some more tests to be sure

6fears7:
Gonna see if we can squeeze this into release:

- [ ] Test remix off
- [ ] Test story menu color swap => Arena
- [ ] Test saint tongue (apparently its wacko)
- [ ] Test color change in story, quit, make new game

6fears7 (approved):

PR #683, opened 2025-02-08
Add MeadowOnline to HiddenOrUnplayableSlugcats

forthfora:
Wanted to suggest this change as I saw this popping up in the Arena selection screen for instance

Was wondering if there was a particular reason it's not being used already tbh, did a quick test with LocalP2P and seemed alright but yeah, let me know


henpemaz (approved):
didn't know this was a thing, 1.5 ui was opt-in not opt-out

PR #684, opened 2025-02-09
6fears7 arena attacker advantage saint

6fears7:
- Add burst x/y sync
- Add monk ascension sync (sometimes wouldn't proc)
- Grant Saint attacker advantage


henpemaz (dismissed):
good to go

if not too much trouble add the saint tag to the ascension flag as well

henpemaz (approved):

PR #685, opened 2025-02-10
Optimization for lenientPos

forthfora:
Don't need to update chunkStates if lenientPos is set to true, save sending some data (assuming meadow can recognise there's no change in the array)

Tested it with Raindeer as I saw it being used there, and with Pearlcat (which uses lenientPos for pearls) and seemed fine, let me know if you can see any issues with this though

### DIFF
--- a/Online/State/RealizedPhysicalObjectState.cs
+++ b/Online/State/RealizedPhysicalObjectState.cs
@@ -16,11 +16,7 @@ namespace RainMeadow
         public RealizedPhysicalObjectState() { }
         public RealizedPhysicalObjectState(OnlinePhysicalObject onlineEntity)
         {
-            if (chunkStates is null || !onlineEntity.lenientPos)
-            {
-                chunkStates = onlineEntity.apo.realizedObject.bodyChunks.Select(c => new ChunkState(c)).ToArray();
-            }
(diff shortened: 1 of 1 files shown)


henpemaz:
Right now lenientPos is resolved and tracked on the receiving side and might not be in sync between all clients. In a few instances also the received pos is used to snap back in place if the physics desync too much (player in deer antlers comes to mind). I don't think physics pos should be omitted unless some of that is reviewed. In general lenientpos is only used in a few niche cases so far (vine climbs and such) I wouldn't waste much time in it.

Also, that is null condition is causing the original code to run regardless of your optimization.

forthfora:
did some logging and I think the change should be effective now, let me know if you can see anything wrong with it still

I should ask though if anyone had better ideas on how I could improve performance?
this was just the first simple thing that came to mind and was easy to do on Meadow's end, but open to other suggestions, hard for me to judge what would have the biggest impact

henpemaz:
Closing this as per the reasoning in my first comment, it's not a good generalized behavior. If you need an optimization for Pearlcat's pearls, hook makestate optionally with your own custom pearl state and assign the chunkstate to empty there.

PR #693, opened 2025-02-12
Switch DeathMessages to use RPC

Quaxly:
The kill feed now gets sent via RPCMethods. When a message is sent the creature's online ID gets stored in the local PlayerSpecificHud and won't display any more messages pertaining to it until the next session. This should prevent duplicates and ensure everyone gets an accurate and consistent kill feed.

It'll definitely be worth community testing this at some point but it works much better than before from my testing over LAN.

### DIFF
--- a/OnlineUIComponents/DeathMessage.cs
+++ b/OnlineUIComponents/DeathMessage.cs
@@ -259,7 +259,7 @@ public static class DeathMessage
     {
         if (crit.killTag != null && crit.killTag.realizedCreature != null)
         {
-            if (crit.killTag.realizedCreature is Player && !RainMeadow.isArenaMode(out var _))
+            if (crit.killTag.realizedCreature is Player)
             {
                 PvPRPC(crit.killTag.realizedCreature as Player, crit, 0);
             }
--- a/OnlineUIComponents/DeathMessage.cs
+++ b/OnlineUIComponents/DeathMessage.cs
@@ -20,6 +20,7 @@ public static class DeathMessage
         var kopo = killer.abstractPhysicalObject.GetOnlineObject();
         var topo = target.abstractPhysicalObject.GetOnlineObject();
         if (kopo == null || topo == null) return;
+        if (target is not Player && target.TotalMass < 0.2f) return;
         foreach (var op in OnlineManager.players)
         {
             op.InvokeRPC(RPCs.KillFeedPvP, kopo, topo, context);
@@ -204,7 +205,6 @@ public static class DeathMessage
             var t = (target.apo as AbstractCreature).creatureTemplate.name;
             var realized = (target.apo as AbstractCreature).realizedCreature;
             if (!ShouldShowDeath(target)) return;
-            if (realized != null && realized.TotalMass < 0.2f) return;
             switch (context)
             {
                 default:
--- a/OnlineUIComponents/DeathMessage.cs
+++ b/OnlineUIComponents/DeathMessage.cs
@@ -161,6 +161,7 @@ public static class DeathMessage
             var k = killer.Template.name;
             var opo = target.abstractPhysicalObject.GetOnlineObject();
             var t = opo.owner.id.name;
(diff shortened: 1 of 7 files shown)


6fears7 (commented):
I see. So you're leveraging the playerSpecificHud to track the list and wipe it the next cycle / round since that's the default behavior SS is. 

Things to test: 

- [ ] Death events while chat is off 
- [ ] Arena death events PvP
- [ ] Arena events CvP
- [ ] Story friendly fire 
- [ ] Story deaths
- [ ] Explosions (higher odds of duping)

Quaxly:
Finished testing Arena and Story over LAN
* [x]  Death events while chat is off
* [x]  Arena death events PvP
* [x]  Arena events CvP
* [x]  Story friendly fire
* [x]  Story deaths
* [ ]  ~~Explosions (higher odds of duping)~~ (Explosions aren't fully implemented yet but as long as the killtag is set a death messages will appear and it won't dupe thanks to the new system)

OnlineUIComponents/DeathMessage.cs
```diff
@@ -159,11 +261,12 @@ public static void CreatureDeath(Creature crit)
...
             if (crit.killTag.realizedCreature is Player && !RainMeadow.isArenaMode(out var _))
             {
-                PlayerKillCreature(crit.killTag.realizedCreature as Player, crit);
+                PvPRPC(crit.killTag.realizedCreature as Player, crit, 0);
             }
             else if (crit is Player)
             {
-                CreatureKillPlayer(crit.killTag.realizedCreature, crit as Player);
+                //CreatureKillPlayer(crit.killTag.realizedCreature, crit as Player);
+                CvPRPC(crit.killTag.realizedCreature, crit as Player);
```
6fears7, line 269: I assumed CvPRPC was going to have the creature who killed a player as the death event, but testing in arena didn't have that happen?

OnlineUIComponents/DeathMessage.cs
```diff
@@ -159,11 +261,12 @@ public static void CreatureDeath(Creature crit)
         {
             if (crit.killTag.realizedCreature is Player && !RainMeadow.isArenaMode(out var _))
             {
-                PlayerKillCreature(crit.killTag.realizedCreature as Player, crit);
+                PvPRPC(crit.killTag.realizedCreature as Player, crit, 0);
```
6fears7, line 264: PvPRPC but mode isn't arena?

Quaxly: The method gets called by BaseGameMode.Killing instead if it's an Arena mode.

OnlineUIComponents/DeathMessage.cs
Quaxly: I'll test it more on my end later. It should be working in Story mode.

Quaxly: Should be fixed in https://github.com/henpemaz/Rain-Meadow/pull/693/commits/33908c73fbcd8ad1e47f9977410bd263704bc516

6fears7: Cool, will compile a build in a few days along with some arena / story fixes and get a playtesting session out.

6fears7:
Think this was tested a few days ago. How did it go?

Quaxly:
Everything works however the system seems to be unreliable sometimes in reporting deaths. I observed that they sometimes just wouldn't send for anyone. I believe it's network latency related and is caused when none of the clients see the cause of a death but I'm not super knowledgeable on the network front of things and how to prevent anything not being sent. The system is MUCH better than it was before I'll say that.

6fears7:
Gotcha, and you didn't experience any game crashing occurrences from this introduced code?

Quaxly:
No crashes to report relating to death messages. We had a few cases of pipe eating which I shared relevant crash logs for in the server but other than that mostly smooth sailing.

6fears7 (approved):

PR #697, opened 2025-02-13
Hotfix to reduce mod-sync strictness

forthfora:
Basically this check is actually the same as checking whether a mod defines a `world` folder (as opposed to `modify/world`)

so certain mods which have a world folder but not a modify/world folder now required sync (e.g. Camera Scroll)

probably worth looking at this again in the future but for now just get rid of it to revert to the old behavior
(RainMeadowModManager checks for modify/world atm, so would make sense to move it here instead for instance)

### DIFF
--- a/ModManager/RainMeadowModManager.cs
+++ b/ModManager/RainMeadowModManager.cs
@@ -58,6 +58,7 @@ namespace RainMeadow
             "blujai.rocketficer",
             "slugcatstatsconfig",
             "explorite.slugpups_cap_configuration",
+            "slime-cubed.slugbase",
         };
 
         public static string[] GetBannedMods()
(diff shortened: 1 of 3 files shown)


RainMeadowJSONGuide.md
```diff
@@ -0,0 +1,36 @@
+## Setup
```
forthfora, line 1: Squeezed some docs into this PR as well

ModManager/RainMeadowModManager.cs
```diff
@@ -58,6 +58,7 @@ public static string[] GetRequiredMods()
             "blujai.rocketficer",
             "slugcatstatsconfig",
             "explorite.slugpups_cap_configuration",
+            "slime-cubed.slugbase",
```
forthfora, line 61: Also seeing whether this can be snuck in, was highly requested as custom slugcat desync causes big issues

henpemaz (approved):

RainMeadowJSONGuide.md
TheLazyCowboy1, line 1: Thanks for this. I'm too lazy to try to interpret code, so this simple guide helps a lot!

PR #700, opened 2025-02-13
Mod Manager Rework (comment system)

forthfora:
Drafting this for consideration for a future update, would change the way the `meadow-bannedmods.txt` and `meadow-highimpactmods.txt` files are used

Issue: contents of these files is not automatically updated when the default lists change in meadow

Solution: use a syntax to determine whether a mod has been removed from the list on purpose, so that we can update the list when entries are missing

Propose using a `//` prefix, for instance in `meadow-bannedmods.txt`:

```
maxi-mol.mousedrag
fyre.BeastMaster
//slime-cubed.devconsole
zrydnoob.UnityExplorer
//warp
```

Would indicate Warp and Dev Console are excluded

Explanation of the system can automatically be added to the top of each file, e.g. for `meadow-syncrequiredmods.txt`:
```
// The following is a list of mods that must be synced between client and host:
// (if the host has these mods enabled/disabled, the client must match)
// To remove mod IDs and allow them to be de-synced, prefix the lines with '//', like this:
//mod-id-to-exclude
```


forthfora:
Needs testing, but works now as far as I know

ModManager/RainMeadowModInfoManager.cs
```diff
@@ -103,11 +103,11 @@ private static RainMeadowModInfo GetGeneratedModInfo(ModManager.Mod mod)
     {
         var modInfo = new RainMeadowModInfo();
 
-        // TODO: update conditions sometime
-        // if (mod.modifiesRegions)
-        // {
-        //     modInfo.SyncRequiredMods.Add(mod.id);
-        // }
+        // TODO: consider mod.modifiesRegions
```
forthfora, line 106: Want to consider: going back to checking `modifiesRegions` (i.e. checking for world directory), then adding camera scroll + expedition to the exclusion list so they're not high impact

Mods that use the world directory (but don't have a modify/world directory) could still do stuff that can desync I imagine, can't think of an example though so dunno if it's an actual concern

TheLazyCowboy1: Well... they could *replace* a file, probably gates/locks.txt, regions.txt, region map files, or region names. In theory, they could also add rooms/regions; but without modifying/replacing regions.txt or any world_xx.txt files, that wouldn't have any effect.

Any number of mods might store something in the worlds folder for any reason. I would sooner suggest searching for world/regions.txt, world/locks.txt, and world/*/world_*.txt specifically. Some amateur modder (like me when I made my first mod) might store stuff in a world folder that doesn't actually affect regions.

Just my opinion.

henpemaz: Might take some whitelisting or clever analysis of the actual files in there, but mods that replace room geo (and therefore den indexes) as well as room settings (and therefore item indexes) can be a source of desync yeah.

The `modify/world` check is there to pick up on new rooms which is where things majorly break.

ModManager/RainMeadowModManager.cs
```diff
@@ -1,76 +1,64 @@
...
 
-            var effectiveHighImpactMods = highImpactMods.Union(RainMeadowModInfoManager.MergedModInfo.SyncRequiredMods.Except(RainMeadowModInfoManager.MergedModInfo.SyncRequiredModsOverride)).ToList();
-            var effectiveBannedMods = bannedMods.Union(RainMeadowModInfoManager.MergedModInfo.BannedOnlineMods.Except(RainMeadowModInfoManager.MergedModInfo.BannedOnlineModsOverride)).ToList();
+            requiredMods = UpdateFromOrWriteToFile(SyncRequiredModsFileName, requiredMods, SyncRequiredModsExplanationComment);
+            bannedMods = UpdateFromOrWriteToFile(BannedOnlineModsFileName, bannedMods, BannedOnlineModsExplanationComment);
 
-            // (high impact + banned) - enabled
-            return effectiveHighImpactMods.Concat(effectiveBannedMods)
+            // (required + banned) - enabled
+            return requiredMods.Concat(bannedMods)
```
henpemaz, line 61: If we're renaming stuff I'd argue that either "sync required" or "high impact" are clearer in purpose than "required" for referring to "mods that require modlist sync"

henpemaz (dismissed):
good to merge if tested

henpemaz:
Also has conflicts in `Utils/Utils.cs`

forthfora:
Fixed the conflicts, updated explanation comments and renamed variables, still needs online testing:

- [ ] Test `meadow-highimpactmods.txt`
- [ ] Test `meadow-bannedmods.txt`
- [ ] Test high impact detection, especially with certain mods (e.g. Camera Scroll & Expedition), should be unchanged

bit busy atm so will have to be this weekend, if someone else finds time before then though feel free to give it a spin

GameModes/ArenaCompetitiveGameMode.cs
```diff
@@ -26,7 +26,6 @@ public class ArenaOnlineGameMode : OnlineGameMode
         public bool painCatEgg = RainMeadow.rainMeadowOptions.PainCatEgg.Value;
         public bool painCatLizard = RainMeadow.rainMeadowOptions.PainCatLizard.Value;
         public bool disableMaul = RainMeadow.rainMeadowOptions.BlockMaul.Value;
-        public int painCatThrowingSkill;
```
forthfora, line 29: Field is duplicated on main

forthfora:
Tested with LDani with several different mods, should be good to go as far as I can tell

6fears7 (approved):

6fears7:
Will need @henpemaz re-approval to merge

ModManager/RainMeadowModManager.cs
```diff
@@ -1,95 +1,87 @@
...
+            // To exclude mods on the list from these requirements, prefix the lines with '//', like this:
+            //mod-id-to-exclude
+
+            """;
 
-        public static string[] highImpactMods = {
-            "rwremix",
-            "moreslugcats",
-        };
+        public static string BannedOnlineModsExplanationComment =>
```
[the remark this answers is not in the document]
forthfora: It's a bit problematic, the way it checks if the comment needs to be added is checking whether the first line of the file matches the first line of the string, that means that whenever the language changes the comment would be reappended

or on second thought maybe I could loop through all languages and check each one to make it work, would prefer to just get this merged and do that in another PR though

PR #706, opened 2025-02-16
Synchronize Player NPCS

invalidunits:


6fears7:
Bro came out the cut with a PR

6fears7 (commented):
Couple of things from my playtesting so far with arena pup:

1. It seems like right now throwing a pup from back to ground with grasps are null messes up the overlap so they can collide with you but nobody else. Might be arena only, not sure but it's agony to figure out why. If you see that issue on your end in story lmk. 

2. I have a draft PR that adds a new AbstractStick for APOs we don't own so we can ensure we get a SlugOnBack stick. This may not be relevant to this PR with NPCs because we'll request them and be transferable I'd imagine, but jsuk

invalidunits:
> 1. It seems like right now throwing a pup from back to ground with grasps are null messes up the overlap so they can collide with you but nobody else. Might be arena only, not sure but it's agony to figure out why. If you see that issue on your end in story lmk.

Not Arena Only. I was encountering this issue in Story Mode on LAN. I'm pretty sure the main issue here is 
ChangeOverlap  not being called if you directly set slugOnBack.slugcat

> 3. Merging this will impact my current draft for Arena with the SlugOnBack property sync. Shouldn't be an issue since I'm working on its inclusion and our release cadence is slow but putting it out there for awareness

#634 and #699 ? 

It seems like most backpacking Slugpups issues generally effect both NPCs and real Players.  In that case, I'd rather ignore backpacking for now and only focus on NPCs exclusive issues such as breaking the Food Meter and being treated like actual players by both gates and Shelters. 

That way, Once your backpacking changes are finished, I can rebase this PR on yours or vise versa.

6fears7:
Definitely not #634. That'll be salvaged over time. That PR is brutal in scope. 

Focus on the NPCs. I'll figure out the overlap unless you want to take that part. If you encounter an issue going through gates with clones let me know because #634 has a solution to that.

invalidunits:
**Issues that should be fixed in this PR**

- [ ] SlugNPCs are considered when meeting shelter requirements. (possibly gates as well) 
- [ ] SlugNPCs change ids through cycles sometimes if not spawned by host.

**Issues that might be resolved by other PRs**
- [ ] Hand Holding sometimes causes SlugNPCs to get stuck in pipes. (especially if the their network owner is not in the room

**Needs Testing**
- [ ] SlugNPCs going through a gate

6fears7:
> **Issues that should be fixed in this PR**
> 
> * [ ]  SlugNPCs are considered when meeting shelter requirements. (possibly gates as well)
> * [ ]  SlugNPCs change ids through cycles sometimes if not spawned by host.
> 
> **Issues that might be resolved by other PRs**
> 
> * [ ]  Hand Holding sometimes causes SlugNPCs to get stuck in pipes. (especially if the their network owner is not in the room
> 
> **Needs Testing**
> 
> * [ ]  SlugNPCs going through a gate


re: Hand Holding. https://github.com/henpemaz/Rain-Meadow/pull/634/files#diff-9e900f50e7938d8932639b5693d1e05279d97642edad0158bb2f2ea84b5b2d80

you may find a better way than that but iirc that worked for dead  bodies that didn't belong to us. Maybe it'll be useful if you swap the dragged entity to see if it's ours and run the enter shortcut code.

Online/State/NPCStateState.cs
```diff
@@ -0,0 +1,74 @@
...
+        public bool HasMark;
+
+        [OnlineField]
+        public bool Glowing;
+
+        [OnlineField]
+        public bool Drone;
+
+        [OnlineField]
+        public bool HasCloak;
```
henpemaz, line 25: Even though all of these exist in the code I'd argue that they might not need sync? I haven't played the dlc campaigns but I can't think of a reason these would get used

henpemaz:
Poke me for approvals when needed, no blockers

invalidunits:
Seemingly the only issue left is Sheltering the the slugpups. It breaks WorldLoading and causes clients to spawn there own 
duplicate ghost slugpups which don't appear for host.

Gates need testing

invalidunits:
Opening this as Ready for review because SlugNPCs are fully functional.

6fears7:
LGTM initially. Will let @henpemaz address the state files. 

Looks like it was a lot of work exploring the problem, good job.

6fears7 (approved):

Game/RainMeadow.PlayerHooks.cs
```diff
@@ -64,19 +69,40 @@ private void PlayerCarryableItem_PickedUp(On.PlayerCarryableItem.orig_PickedUp o
         orig(self, upPicker);
     }
 
+    // This is Abysmal and doesn't work
```
6fears7, line 72: @invalidunits do we need to keep this commented code here?

invalidunits: Definitely not. I'll remove it when I get home.

Online/Entity/OnlinePhysicalObject.cs
```diff
@@ -109,6 +109,10 @@ public static OnlinePhysicalObject RegisterPhysicalObject(AbstractPhysicalObject
         public static OnlinePhysicalObject NewFromApo(AbstractPhysicalObject apo)
         {
             bool transferable = !RainMeadow.sSpawningAvatar;
+            if (apo.realizedObject is Player player) {
+                transferable = transferable || player.isNPC;
+            }
+
```
henpemaz, line 115: I'm almost sure this isn't needed, should have no effect anyways. That static bool wraps player instantiation only so I'd imagine no slugpups/npcs are registered during that, let alone anything is realized around there either.

Online/State/AbstractCreatureState.cs
```diff
@@ -7,6 +7,9 @@ public class AbstractCreatureState : AbstractPhysicalObjectState
         [OnlineField(group = "realized")]
         public WorldCoordinate destination;
 
+        [OnlineField(nullable = true)]
+        public ArtificialIntelligenceState? artificialIntelligenceState;
+
```
henpemaz, line 12: My gut feeling about something like this is that it should be for realized state only so in RealizedCreatureState, but I wouldn't be surprised if for slugpups specifically the game has conditions for them never abstractizing or some other bs like that...

Online/State/NPCStateState.cs
```diff
@@ -0,0 +1,74 @@
...
+        // [OnlineField]
+        // public bool HasCloak;
+
+        /*
+        [OnlineField]
+        public OnlineEntity.EntityId player;
+        */
+
+        [OnlineField(nullable: true)]
+        public OnlineEntity.EntityId? StomachObject;
```
henpemaz, line 33: PlayerStateState already has a `objectInStomach` field that will go unused? for npcs.

Either reuse it or add an `[OmitFields("objectInStomach")]` to this class, please

henpemaz (approved):
Honestly quite impressive job with the AI and social memory stuff, looks pretty good. A couple notes about optimization but ok as is if you guys need to.

invalidunits:
Will merge this after I test if gates work with Players. A follow up PR is definitely required for entityID and AI sync. Then my main focus will be Scavengers.

Online/State/NPCStateState.cs
```diff
@@ -0,0 +1,74 @@
...
+
+
+        // this goes completely unused. won't synchronize for now
+        // public bool DieOfStarvation;
+
+        // [OnlineField]
+        // public int KarmaLevel;
+
+        // [OnlineField]
+        // public bool HasMark;
```
6fears7, line 16: @invalidunits keep or toss?

invalidunits: Keeping these here because I have suspicion that they may come up later. (such as in Artificer cutscenes)

henpemaz (approved):

PR #710, opened 2025-02-21
Improvements to Kick handling

Quaxly:
Some extra checks and safeguards to prevent lobby breaking exploits. Steamworks doesn't natively support kicking/bans so workarounds will be needed.

### Goals:

- [x] Only accept KickToLobby RPCs issued by the lobby host on the client side
- [ ] Immediately boot players out of the lobby and then show the dialog upon being kicked.
- [ ] Stop sending game updates and RPCs to anyone the host has verifiably kicked if they don't get immediately disconnected. Effectively renders them as disconnected to other users in the lobby.
- [ ] Test it all

### DIFF
--- a/Online/RPCs.cs
+++ b/Online/RPCs.cs
@@ -61,8 +61,10 @@ namespace RainMeadow
         }
 
         [RPCMethod]
-        public static void KickToLobby()
+        public static void KickToLobby(RPCEvent rpc)
         {
+            RainMeadow.Debug($"{rpc.from} is trying to kick {rpc.to}");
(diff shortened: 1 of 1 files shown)


6fears7:
@Quaxly if you hit any roadblocks let me know

Quaxly:
Should be all good to merge. Testing in a LAN setting worked as expected and Steam shouldn't be much different.

6fears7 (approved):
@invalidunits any issues on your end?

invalidunits (dismissed):
Overall looks good. 
Could you make use of `OnlineManager.netIO.ForgetPlayer()` after ban or kick?

PR #716, opened 2025-02-23
Translation Support

forthfora:
Adds all the missing Translate() methods / redoes some code where necessary so that everything is technically possible to translate now.

Few notes:
- Lobby name needed to be reworked slightly, `'s Lobby` is no longer appended on set, instead it's appended when received. This results in a duplicated suffix atm, should be fine once everyone has these changes though
- For translators: use the `<USERNAME>'s Lobby` format in the translation files.
- The RAIN MEADOW title image can also be translated, append the name of the language to the image file. Falls back to English if translation specific image is not found.


forthfora:
should be good to go, needs thorough testing though

also just to be clear, this PR doesn't translate anything on its own, just provides the groundwork to make translations possible

forthfora:
@zbiotr changes are here, just so you're aware - if I've missed anything or it doesn't translate properly, let me know

zbiotr:
I didn't test it, but I think i've found some issues:

- `\` usually breaks the translation, like `\"cheats\"` in MeadowRemixOptions.cs, and `\n` in OnlineGameMode.cs; in the case of `\"cheats\"` just replace the quotation marks with `“”`
- Some texts don't have a Translate() method: "_Prepare for combat, [Slugcat]_" in Competitive.cs, "_CLEAR_" in MeadowEmoteHud.cs, "_Password Required_" in LobbySelectMenu.cs, and " _- {storyGameMode.region}_" in StoryOnlineMenu.cs
- Other ones will very likely not translate for having curly brackets inside Translate(): "_Press '[ChatButtonKey]' to chat, press '[ChatLogKey]' to toggle the chat log_" in ChatHud.cs, and "_Failed to join lobby.`<LINE>`{error}_" in LobbySelectMenu.cs
- I don't know if it already translates, but every instance of DialogNotify() doesn't have Translate() either

I'd recommend that you test this on your end, since I don't know how to do it myself (all the issues I mentioned are solely from experience translating this game, so I may be incorrect in some of them). I can also wait for you to finish the Chinese translation, I have no problem.

forthfora:
Latest commits should fix the arena crash + the newline issue

I remember you talking about the string interpolation issue too e.g. `Translate($"Text goes here: {someVariable}")`, yeah you're completely right that it won't work, should be replaced with `Translate("Text to translate:") + " " + Translate(someVariable)`, if you encounter any specific cases let me know and I can update the code (or feel free to do it in your PR)

6fears7:
@zbiotr Let me know how it looks on your end when you get the chance

zbiotr:
> if you encounter any specific cases let me know and I can update the code (or feel free to do it in your PR)

I think the last problem is that the "Stop Inputs While Spectating" option is overlapping the "Pointing" option in the Remix menu. I guess you should remove the "Pointing" one since it's not part of the actual code anymore, unless you want to reposition it.

Btw, I added some Translate() methods in my PR to the last cases where text wasn't getting translated (1c3daa346f236d3f78a8cef715e6c755424da384), so this should be ready to merge once that problem is fixed.

> Let me know how it looks on your end when you get the chance

Everything looks fine to me, except for the issue I previously mentioned, of course.

6fears7:
Pointing should already be removed from remix in main, so pull as necessary @forthfora

6fears7 (changes requested):
The remix page  needs to sync up with main. Pointing remix key binding needs to be removed.

forthfora:
Removed the binding, just wanna point out the configurable itself is still there, means that it'll be stuck at whatever the user last set it as before the update (unless they revert all options I guess), ignore me if that's not a concern though

6fears7:
Yeah I'll remove the functionality before release. Thanks for *pointing* it out.

6fears7 (approved):
Tested, looked good across modes.

henpemaz (approved):

PR #717, opened 2025-02-24
register and clean out missing remotes

6fears7:
- Most useful in story scenarios where a remote object may be left in unowned and unhandled state around gates

### DIFF
--- a/Online/Entity/OnlinePhysicalObject.cs
+++ b/Online/Entity/OnlinePhysicalObject.cs
@@ -172,6 +172,14 @@ namespace RainMeadow
             }
         }
 
+        public static void RegisterAndCleanOutRemoteEntity(AbstractPhysicalObject apo, OnlineResource rs)
+        {
+            RainMeadow.Debug("Unregistered entity being registered and exiting...");
+            var opo = RegisterPhysicalObject(apo);
+            opo.ExitResource(rs);
+
+        }
+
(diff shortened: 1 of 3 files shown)


henpemaz (commented):
I really don't see how this works.

If the problem is a local unhandled thing then it being registered as local and calling exit has no real consequences, since local stuff is controlled by the game anyways so stuff like "leaveimpl" never runs.

If the problem is a remote abandoned object then there's something amiss with the mechanism of how we remove remote stuff both at gates or when a resource becomes no-owner while still in use (host left region, client gate still doing its thing).

Besides, this re-registering picks up a new entityid (owner field differs) and would have no consequence on other clients?

Anyways, hard to follow if/how this works. Does this actually solve the problem or is it sort of blind firing?

henpemaz:
So all we need is that the APO has an OnlineEntity to fix that? My guess would be the same happens if a slugcat-on-back disconnects mid game (might be very much timing dependant and require 3 clients, disconnect comes in earlier than data of slug-no-longer-in-back).

If the issue is in incoming data referencing a missing slug, I think I'd prefer to unbreak the stickref/slugonbackref so that it referencing a missing object doesn't break as badly.

If the issue is that we're still referencing the leaving slug locally then the case of a slug-on-back could be added to `leaveimpl` so that the leaving slug is cleared out of those references. Similar to how there's clauses there to "lose all sticks".

To me "mysterious weird handling that fixes issue as side effect" is weird.

6fears7:
I also may need to return to revisit this idea since at the time I don't think there was a abstractstickrepr that accepted different owners, so it's possible that the issue this fixes is resolved from that. Will get back to you on it.

PR #718, opened 2025-02-24
Fix UDP Buffer Errors

invalidunits:


henpemaz (approved):
Looks good. Should I update codeowners for these files? Public or iu?

invalidunits:
Thanks!

> Looks good. Should I update codeowners for these files? Public or iu?

I wouldn't mind being the code owner of MatchMaking folder. That's my main focus right now.

PR #720, opened 2025-02-25
Arena  parrying improvements

6fears7:
Provides slightly more benefit to defender when they've parried on their end.


henpemaz (commented):
There's a few problems with this implementation.

I think the main one is that the logic of "ignore the next violence call in general" can be very faulty.

I suggest you store "isParried" either in the onlineWeapon (do we have that? is it worth it?), or OnlinePhysicalObject, or a CWT of weapon, then you can check if the source-object of violence has been parried.

You'll also need a reset to isParried outside of the violence RPC for weapons that got parried but never actually hit a target (miss or got parried on both clients). I suggest it being picked up or carried. Switching to thrown would be ideal but it's very problematic as the incoming spear-state will be as thrown for a few frames after the parry. Player-throw-weapon isn't its own rpc is it?

henpemaz:
Are CreatureViolence and HitByWeapon mutually exclusive? If we may receive both from the same weapon throw, then the parry as implemented will only count for one. A way around that would be to set the flag down "late", I've got a helper around in onlinemanager.runlate or similar for that sort of stuff.

Game/RainMeadow.PlayerHooks.cs
```diff
@@ -50,9 +50,23 @@ public void PlayerHooks()
...
     }
+
+    private void PlayerCarryableItem_PickedUp(On.PlayerCarryableItem.orig_PickedUp orig, PlayerCarryableItem self, Creature upPicker)
+    {
+        if (OnlineManager.lobby != null)
+        {
+            if (upPicker.IsLocal() && upPicker.abstractPhysicalObject.GetOnlineObject().didParry)
+            {
+                upPicker.abstractPhysicalObject.GetOnlineObject().didParry = false;
+            }
```
henpemaz, line 65: Checking if a bool is true to set it to false is pointless.

Plus, I'd argue that there's no need for local check, if remote picks stick parry should reset as well or it'll still be active.

6fears7:
> Are CreatureViolence and HitByWeapon mutually exclusive? If we may receive both from the same weapon throw, then the parry as implemented will only count for one. A way around that would be to set the flag down "late", I've got a helper around in onlinemanager.runlate or similar for that sort of stuff.

Unfortunately, HitByWeapon is not guaranteed to be called, especially in close quarters. I'm not sure why as I've ripped my hair out exploring violence and spear properties that might need syncing to make it happen but so far the behavior is that Violence can be expected to be called, but HitByWeapon is not. So in theory I suppose I could reduce down to just _Violence

6fears7:
- [x ] Run deferred on both HitByWeapon and Violence
- [x] Reset on pickup
- [ ] Playtest changes

henpemaz (approved):

PR #722, opened 2025-02-28
Mod Compatibility Improvements

TheLazyCowboy1:
Made two small changes to make mod porting slightly easier (for me):
1. Made an easy way to get custom RealizedPhysicalObject states. Also works for Creatures.
    - (My solution might seem weird, but I'm an amateur coder, and it works 100% fine)
2. Added a try/catch so my Room Connection Randomizer's map works.

Note: This was branched off of my mod filtering code, since that's my main branch atm, and I'm too lazy to change it.


Online/State/AbstractPhysicalObjectState.cs
```diff
@@ -41,9 +42,26 @@ public AbstractPhysicalObjectState(OnlinePhysicalObject onlineEntity, OnlineReso
             if (realizedState) this.realizedObjectState = GetRealizedState(onlineEntity);
         }
```
TheLazyCowboy1, line 44: This is the method I used for modded RealizedPhysicalObjectStates. I tested it for my Sword Mod; it works.

Online/State/AbstractCreatureState.cs
```diff
@@ -26,6 +26,14 @@ protected virtual CreatureStateState GetCreatureStateState(OnlineCreature online
...
         protected override RealizedPhysicalObjectState GetRealizedState(OnlinePhysicalObject onlineObject)
         {
+            //look for modded creature states
+            foreach (var func in CustomObjectStateHooks)
+            {
+                var result = func(onlineObject);
+                if (result != null)
+                    return result;
+            }
+
```
TheLazyCowboy1, line 36: Also put here for custom creatures, because otherwise custom creatures would always return RealizedCreatureStates.

Meadow/Items/MeadowCollectible.cs
```diff
@@ -26,12 +26,13 @@ public override void PlaceInRoom(Room placeRoom)
...
-                this.abstractPhysicalObject.pos.Tile = place;
-                abstractCollectible.placed = true;
-                RainMeadow.Debug("placed at " + abstractPhysicalObject.pos);
+                if (RoomSession.map.TryGetValue(placeRoom.abstractRoom, out var rs) && rs.TryGetData<MeadowRoomData>(out var mrd))
+                {
+                    var place = mrd.GetUnusedPlace(placeRoom); // todo extra reqs for places ie not-narrow etc
+                    this.abstractPhysicalObject.pos.Tile = place;
+                    abstractCollectible.placed = true;
+                    RainMeadow.Debug("placed at " + abstractPhysicalObject.pos);
+                }
```
TheLazyCowboy1, line 35: This fixed the map generation code for my Room Connection Randomizer.

6fears7:
@TheLazyCowboy1 please address the conflicted file to support the Utils.Translate() addition made to main

TheLazyCowboy1:
Ugh I've had to fix this one stupid conflict about 15 times... at least it's an easy fix.

6fears7 (changes requested):
Please either make a new PR with just these changes instead of having bleed from mod management, or update this PR to strip that content to make the diff review easier.

TheLazyCowboy1:
> Please either make a new PR with just these changes instead of having bleed from mod management, or update this PR to strip that content to make the diff review easier.

Ooops yeah my bad; I merged my other PRs but I forgot this one. It's merged as well as GitHub will allow.... which isn't perfectly, lol.

Meadow/Items/MeadowCollectible.cs
henpemaz: your mapgen shouldn't be placing items, the devtools map doesn't when it realizes room for the minimap rendering. fix your mod not mine

henpemaz:
The state system is override-friendly, and makestate is the hottest of hot-paths, I don't think it needs a second lookup in there.

All of this is already achievable by overriding OnlinePhysicalObject or OnlineCreature's `MakeState` but registering a proper new online type with definition onlinestate and realizedstate is probably too convoluted and requires a hook to OnlinePhysicalObject.NewFromApo.

Move the lookup to OnlinePhysicalObject's constructor and store the result, or look for a way to simplify registering OnlineEntity types.

Meadow/Items/MeadowCollectible.cs
TheLazyCowboy1: I will re-explain the issue in more detail for you:

Meadow mode is attempting to place collectibles before the Room Session exists. This completely breaks the game. Such an issue, regardless of its source, should definitely be fixed. This try/catch doesn't hurt anything, but it does ensure that the game doesn't break.

The dev tools map is typically used long after every room in the region has been fully realized. However, if any room has not yet been realized, this issue occurs. Because my mod wants to have a working map on cycle 1, it has to generate the map before the map is fully initialized, which is necessarily before every room has been fully loaded.

This is not a Rain World map generation issue.
This is not a mod issue on my end.
This is a Meadow mode trying to place collectibles before the room exists issue.

TheLazyCowboy1:
I'm confused. If you don't like the way that RealizedPhysicalObject states are currently initialized, then why not change the way that they are initialized?
I simply added a simple way for mods to use the same method that Rain Meadow itself currently utilizes. I see no reason why initializing states should be different for modded items than it is for other items.

> look for a way to simplify registering OnlineEntity types.

1. Do you really want me to go messing with your code when I don't fully understand it, lol? If you don't like the current system, then you probably shouldn't be asking me to re-work it. I don't really understand what you want me to change anyway.
2. Why do mod authors have to rework Rain Meadow's engine whenever they want to improve compatibility? Shouldn't everything be treated in the same way by the same system, instead of having modded items work totally differently than other items (which are treated identically by Rain World itself)?

henpemaz:
> Do you really want me to go messing with your code when I don't fully understand it, lol?
This snarky attitude of yours sucks ass, you're already messing with code you don't fully understand

TheLazyCowboy1:
It is a serious yes or no question: Do you want me to go try to rework the way that states are initialized?

henpemaz:
> It is a serious yes or no question: Do you want me to go try to rework the way that states are initialized?

If you feel there's a serious need for it (somehow the OnlineEntity Data API doesn't cover some use case of yours)

But be aware that this is a hot path and I'll be somewhat serious about how necessary it is versus how it performs.

TheLazyCowboy1:
Okay, now I see your point.

Manual hooks are still pretty easy to use. I'll just tell folks to use a manual hook for modded items. It's a sensible (albeit less than ideal) solution.
And I'll move the try/catch into a new PR.

PR #708, opened 2025-02-20
Mod Applier Fix

TheLazyCowboy1:
Fixes the excellently written ModApplier.cs code to function properly. Seriously, this code is great; who gave up on it???

## Tasks:
- [x] Re-implement the option to enable mods required by the lobby
- [x] Correctly order the mods that get enabled
- [x] Make the user immediately join the lobby after his mods apply IF no restart is required (I'm really stuck on this one; I have no clue how to do it)
- [x] Check for different mod orders
- [x] Make a popup asking whether the user wants to copy the lobby's load order
- [x] Test whether that functionality actually works consistently
- [ ] Test whether the error loading mods message actually works (how do I force a mod load error??)
- [x] Implement a DLC required check so that the mod applier doesn't try to enable unowned DLC
- [x] Test that the DLC required check actually works
- [x] Ensure that dependencies are correctly ordered when added
- [x] Move banned mods list to LobbyInfo
- [x] Make joining lobbies after mod changes (without a restart) succeed consistently
- [x] Make the restart functionality work for both Steam and LAN lobbies, with or without passwords
- [ ] Get playtesters to torture test this thing

### Pros:
- Makes playing with mods far easier for all users (a huge plus in my mind!)
- Decreases glitches, crashes, and bug reports due to improperly ordered mods
- Fixes the issue of being unable to join lobbies that modified ExtEnums (because the ExtEnums errors would occur before the mod check)
- Just... use it... I think it speaks for itself.

### Cons:
- Probably has some bugs we won't catch (it's simply a big enough PR; I've surely missed something)
- Might be quite a pain to maintain in the long-run
- My messy band-aid code isn't the best

PR #723, opened 2025-02-28
Extended High-Impact Mod Detection

TheLazyCowboy1:
Vastly expands/improves the detection of required and banned mods.
(To clarify: I use the term "high-impact" to refer to mods that are required if the host has them and banned if he doesn't.)

* Mod files are more rigorously checked for region changes.
* All custom slugcat mods (that use Slugbase) are marked as high-impact.
* Clients will detect if they have any of high-impact mods that the host does not, and will treat them as banned mods.
* Mod assemblies are marked as high-impact if they implement:
  - Fisobs (covers most custom items)
  - Critobs (covers most custom creatures)
  - PhysicalObjects (covers SO MUCH, including my Sword Mod)
  - Creatures (should already be covered by PhysicalObjects)
  - OnlineStates
  - RPCMethods
* Checks are put in place so that the (probably expensive) assembly checks aren't done more than once per mod.

Note: This PR builds off of #700 and #708 (which depends on #658)
I recommend only checking through Utils.cs, RainMeadowModManager.cs, and RainMeadowModInfoManager.cs.

This PR needs extensive testing to detect any loopholes I might have missed, but it doesn't break anything at least.


TheLazyCowboy1:
I have tested this to what I think is a sufficient extent, and as far as I can tell, it is working properly.

However, there is one small point I'm not sure about:
**If the host marks a mod as BannedOverride, should clients ignore that mod in their "RequiredMods" list?**

That is, I tried marking //LazyCowboy.SwordMod as a non-banned mod. However, when the client tried to join (with the Sword Mod active), the client noticed a mismatch between the required mods lists (not the banned mods lists; just the required ones) and thus refused to join. ...Is this acceptable?

TheLazyCowboy1:
@forthfora What do you think? On a scale from concerningly crazy to completely crazy, how crazy crazy am I??

forthfora:
Filter conditions look sane, should wait for the other PRs to get merged to look more closely though, it's hard to tell what's part of this PR in `RainMeadowModManager.cs`

forthfora:
> I have tested this to what I think is a sufficient extent, and as far as I can tell, it is working properly.
> 
> However, there is one small point I'm not sure about: **If the host marks a mod as BannedOverride, should clients ignore that mod in their "RequiredMods" list?**
> 
> That is, I tried marking //LazyCowboy.SwordMod as a non-banned mod. However, when the client tried to join (with the Sword Mod active), the client noticed a mismatch between the required mods lists (not the banned mods lists; just the required ones) and thus refused to join. ...Is this acceptable?
> 
> (Note: If the client marks the mod as exempt from requiredMods AND the host marks it as exempt from bannedMods, then it will work.)

This also needs more thought but my initial reaction is that a mod that's implicitly sync required should also have to be explicitly excluded from sync (like if a mod is not banned, that doesn't mean it's not sync required, that just means the host doesn't block it if the host doesn't have it)

TheLazyCowboy1:
> Filter conditions look sane

Seriously? I thought they were a terrible confusing mess, lol.

Regarding the question of overrides, I think the best solution is to include overrided required mods in the required mod list (with the comments preceding, of course). That way, when clients attempt to join, they can join regardless of whether they have the mod.

Otherwise, imagine the host overrides a mod to not be required. That would mean that NO client can join using that mod, yet the host has it. That seems to defeat the point.


Anyway, I'll keep this PR as a draft in the background until some of the PRs it depends on get merged. So if you're for some reason reading this, go work on something more important!

TheLazyCowboy1:
Brief note:
* Added OnlineState to the list of high-impact types.
* Also check for RPCMethods, using the same method used by RPCManager.

This should hopefully say goodbye to any issues due to registration of modded states and RPCs!

TheLazyCowboy1:
This ought to be tested with other Steam users to ensure that nothing unexpected happens. However, since most of this is basically client-side, I doubt it should be different when connecting through Steam than it is in my LAN tests.

As far as I know, this PR is basically ready for review, so if anyone wants to get a head start reviewing it, feel free.

PR #724, opened 2025-03-01
Object visual position improvements

TheLazyCowboy1:
Behold! My three lines of code!

Slightly improves the visual position of objects by properly updating lastPos and lastRotation.
That's it.
Just slightly.


Online/State/RealizedPhysicalObjectState.cs
```diff
@@ -61,6 +61,8 @@ public void CustomSerialize(Serializer serializer)
 
         public void ReadTo(BodyChunk c)
         {
+            c.lastLastPos = c.lastPos;
+            c.lastPos = c.pos;
```
TheLazyCowboy1, line 65: So complicated

Online/State/RealizedWeaponState.cs
```diff
@@ -41,6 +41,7 @@ public override void ReadTo(OnlineEntity onlineEntity)
             }
             weapon.thrownBy = thrownBy?.realizedCreature;
             if (weapon.grabbedBy != null && weapon.grabbedBy.Count > 0) { RainMeadow.Trace($"Skipping state because grabbed"); return; }
+            weapon.lastRotation = weapon.rotation;
```
TheLazyCowboy1, line 44: The difference is probably hardly noticeable, but it's better than nothing.

6fears7:
Depends on #658

TheLazyCowboy1:
(It technically doesn't depend on that at all; I'm just too lazy to rebase it)

6fears7:
Closing in favor of #733 due to git diff

Issue #687, opened 2025-02-10
Dedicated servers for GOG


invalidunits:
GOG players are currently unable to play online across different networks. Implementing Dedicated Tunnels would resolve this limitation, enabling GOG players to connect seamlessly. Additionally, Crossplay between GOG and Steam users would be possible.

Dedicated Tunnels would allow
- GOG players to connect online from different networks.
- Cross-platform compatibility between GOG and Steam versions.
- No more 250 player limit. Scug Battle Royale.

**Considerations**:

-   **Hosting Costs and Platform**: 
     Azure, Google Cloud, and AWS are all options. I might be able to pay for servers out of pocket, donations could also help

-   **Domain Name**: 
      A good domain name is important for tunneling, I own an unmaintained domain name afkpals.net.

-   ** Identification & Logging In **
      Players need a way to log in and verify their identity (for moderation and security reasons.) IPs are not sufficient, players need a consistent UUID.

henpemaz:
![](https://cdn.discordapp.com/emojis/872782599435726909.webp?size=128)

6fears7:
You either die in development hell or live long enough to become a cloud product

PR #727, opened 2025-03-05
Router Multiplayer

invalidunits:
For #687


invalidunits:
Removing milestone due to focus on other PRs.

niacdoial:
I might try to pick this one up, if you don't catch up to it yourself.
Can I ask what the goal of "router multiplayer" is?

- Is it "just" some sort of NAT-proofing of the existing LAN multiplayer ?
  - does NAT piercing on UDP even allow this? (if player 2 pierces NAT to connect to host 1, then receives an "unprompted" (as far as the NAT knows) packet from player 3)
  - if so, is it supposed to support changing who the host is?
- If not, what is the scope of this?

...alright thanks!

invalidunits:
> I might try to pick this one up, if you don't catch up to it yourself. (EDIT: this sounds rude... I meant that I want to try, but I wouldn't be too mad if you end up focusing back on this and taking this out of my hands) Can I ask what the goal of "router multiplayer" is?

No issues! I've been working hard IRL and on other mods so I haven't had the time to work on this. Hopefully that'll end soon. NAT requires a third party public server to distribute client information. The Router NetIO would communicate with public servers to make this possible.

This needs another application to work as a public server. I was attempting on an application which could do this but haven't worked on since the creation of this PR. You're welcome to take it up if your interested, hop on the Meadow discord server and we can talk about it.

PR #729, opened 2025-03-05
RPCs for Soft Compatible Mods

TheLazyCowboy1:
Made this for Push To Meow (and similar cosmetic mods). It gives a way for mods to basically send their own custom RPCs without having to register an RPC (and thus become a high-impact mod).

This should NOT be used unless the mod truly wants to remain low-impact!

Blocking players from joining lobbies if they had Push To Meow enabled sounds absurd; this is a better option.


TheLazyCowboy1:
Oh, and again I'll note:
My Mod Filter PR was basically my first time ever using GitHub, so I was stupid enough to do it in the "main" branch of my fork. I have tried for countless hours to rebase my other PRs; I can't figure it out. So, could you just... merge my Mod Filter PR so I don't have to worry about it anymore?

The only changes in this PR are in Online/ModdedDataRPC.cs

invalidunits:
Not a fan of this PR. Mods should use [RPCMethod] Attribute for RPCs. The current Main issue with that is load order conflicts  and optionality.

Adding an optional flag to RPCMethod, simply ignoring the RPC if it doesn't exist on the targeted system.
Storing RPCs by their originating Assembly could fix the load order issues.

TheLazyCowboy1:
> Storing RPCs by their originating Assembly could fix the load order issues.

Yeah, looking at it, it wouldn't be too difficult to, say, modify the RPC dictionaries to include the hash of their assembly.
I might try that.

TheLazyCowboy1:
This is outdated, overcomplicated, and doesn't work. #730 is an adequate replacement for this PR.

PR #730, opened 2025-03-06
Fix RPCs being dependent on load order

TheLazyCowboy1:
Instead of merely sending the index of an RPC, RPCs are now:
* Sent with a ushort hash of their assembly name (to differentiate between mods)
* Sent with an index still, but one relative only to other RPCs in the assembly (so load order independent)
* Sent with a byte indicating the length (in bytes) of their arguments
* Ignore RPCs with unrecognized assembly hashes, and just read through the bytes of the indicated length without serializing them.

I really need help testing this. Rain World acts weird when I try to have multiple instances using different mods at the same time.


Online/OnlineManager.cs
```diff
@@ -219,6 +219,8 @@ public void ProcessDeferredEvents()
 
         public static void ProcessIncomingEvent(OnlineEvent onlineEvent)
         {
+            if (onlineEvent.aborted) return; //don't process aborted events!
+
```
TheLazyCowboy1, line 223: Note: Used the aborted flag to prevent the OnlineManager from attempting to process unrecognized RPCs.

Online/RPCManager.cs
```diff
@@ -17,12 +17,14 @@ public class RPCMethodAttribute : Attribute
 
     public static class RPCManager
     {
-        public static Dictionary<ushort, RPCDefinition> defsByIndex = new Dictionary<ushort, RPCDefinition>();
+        //public static Dictionary<ushort, RPCDefinition> defsByIndex = new Dictionary<ushort, RPCDefinition>();
+        public static Dictionary<ushort, Dictionary<ushort, RPCDefinition>> defsByIndex = new();
```
TheLazyCowboy1, line 21: RPCs are now arranged by assemblyHash, THEN by index.

Online/RPCManager.cs
```diff
@@ -45,12 +47,12 @@ public static void SetupRPCs()
                     {
                         try
                         {
-                            RegisterRPCs(type);
+                            RegisterRPCs(type, (ushort)assembly.FullName.ToCharArray().GetHashCode()); //hashes the assembly NAME
```
TheLazyCowboy1, line 50: Please let me know if this should work. It seemed that assembly.FullName.GetHashCode() was returning random results, so I tried converting it to a char array, which probably doesn't make a difference...?

Online/RPCManager.cs
```diff
@@ -63,7 +65,8 @@ public static void SetupRPCs()
             }
         }
 
-        static ushort index;
+        //static ushort index;
+        static Dictionary<ushort, ushort> indices;
```
TheLazyCowboy1, line 69: Instead of a universal RPC index counter, each assembly has its own index counter.

Online/RPCManager.cs
```diff
@@ -234,20 +242,69 @@ public override void CustomSerialize(Serializer serializer)
...
+                    //finally, move back to the end
+                    serializer.stream.Position = argsEndPos;
+
+                    RainMeadow.Debug($"Sending RPC for assembly {handler.assemblyHash}, index {handler.index}, of length {argsLength}");
+                }
+                catch (Exception)
+                {
+                    RainMeadow.Error($"Error serializing RPC {this}");
+                    throw;
+                }
```
TheLazyCowboy1, line 278: Relocated the try/catch and handler.serialize methods; primarily just to make it easier to understand.

Online/RPCManager.cs
```diff
@@ -234,20 +242,69 @@ public override void CustomSerialize(Serializer serializer)
...
-                handler.serialize(this, serializer);
-            }
-            catch (Exception)
-            {
-                RainMeadow.Error($"Error serializing RPC {this}");
-                throw;
+                //read the header first no matter what
+                ushort modId = serializer.reader.ReadUInt16(),
+                    rpcIdx = serializer.reader.ReadUInt16();
+                byte argsLength = serializer.reader.ReadByte();
```
TheLazyCowboy1, line 285: Each RPC has: an assemblyHash, an index, and a length.
Is a byte enough? I assumed so. It's pretty easy to convert it to a short if not.

TheLazyCowboy1:
@invalidunits Is this better? This system seems a lot better, but I don't have much experience with serialization, so I might be doing everything wrong. It's kinda hard to playtest.

henpemaz:
I'm kind of against it, a lot of things rely on indexes anyways I'd rather save the bytes

TheLazyCowboy1:
Yikes. Looks like Rain Meadow modding support is pretty scanty. How about this: implement this, and then implement a compression algorithm, and you'll win out byte-wise.

invalidunits:
> @invalidunits Is this better? This system seems a lot better, but I don't have much experience with serialization, so I might be doing everything wrong. It's kinda hard to playtest.

This PR looks good but @henpemaz has the final say. 

If you reduce the the index and assembly hash to one byte, we have the same amount of information as before. but then the assembly hash is worthless and likely to conflict... 

A channel system might work well for this where each mod claims a channel and which channels are for which mods are sent with the lobby Join information, but that might be over engineering. I honestly have no clue.

TheLazyCowboy1:
Yeah, I'll change the mod hash to one byte, and I have an idea to ensure the index won't be a problem at one byte.
(Simply: if the index is > 255, increase the mod hash by 1!)

OR, I could keep it as just one ushort index, but have that index x-or'd with the assembly hash. That way it should be mostly unique.

TheLazyCowboy1:
Okay, is the PR acceptable now that it's literally only adding a **_single byte_** per RPC?

Think of all the advantages this could have for modmakers! If people want players to meow, they can install Push To Meow and hear others' meows. If they don't, they can just not enable Push To Meow!

henpemaz:
Again, unnecessary bytes on every rpc. If you want to make "rpcs for mods that aren't sync-required" then try and do just that instead of mangling with the existing systems.

henpemaz:
> Yikes. Looks like Rain Meadow modding support is pretty scanty. How about this: implement this, and then implement a compression algorithm, and you'll win out byte-wise.

Ah yes everyone got infinite CPU to run the heaviest mod of all times. Cut that attitude

PR #733, opened 2025-03-07
simplifed diff of cowboy object state pos pr

6fears7:
Reference: #724

### DIFF
--- a/Online/State/RealizedPhysicalObjectState.cs
+++ b/Online/State/RealizedPhysicalObjectState.cs
@@ -61,6 +61,8 @@ namespace RainMeadow
 
         public void ReadTo(BodyChunk c)
         {
+            c.lastLastPos = c.lastPos;
+            c.lastPos = c.pos;
             c.pos = pos;
             c.vel = vel;
         }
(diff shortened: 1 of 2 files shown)


Online/State/RealizedPhysicalObjectState.cs
```diff
@@ -61,6 +61,8 @@ public void CustomSerialize(Serializer serializer)
 
         public void ReadTo(BodyChunk c)
         {
+            c.lastLastPos = c.lastPos;
+            c.lastPos = c.pos;
```
henpemaz, line 65: c.pos is extrapolated and makes little sense to keep

when state data is received, 4 out of 6 times the game has already ran that frame and updated all objects (extrapolation), including updating lastpos and lastlastpos.

the game uses those to interpolate the graphics position at 60 fps from the 40 fps tick rate. normally it'd interpolate between "last tick's (interpolated or received) pos" and "new pos", now with this it'd be interpolating between "current extrapolated pos" and "new pos" which isn't right, it'll jump ahead of where its supposed to draw

fyi @TheLazyCowboy1

henpemaz:
Proper timing of object data will take more effort than this, this is wrong 4 out of 6 times so I'd rather not keep it

TheLazyCowboy1:
Hmm, makes sense. That can be fixed with 2 lines of code. I'll go do that.

PR #735, opened 2025-03-09
Open High-Impact / Banned Files Buttons

forthfora:
Adds buttons to open the high impact & banned mods files to the lobby creation menu.

Updates the contents before opening, and opens them in the default text editor.

### DIFF
--- a/Menu/LobbyCreateMenu.cs
+++ b/Menu/LobbyCreateMenu.cs
@@ -4,6 +4,7 @@ using Menu.Remix;
 using Menu.Remix.MixedUI;
 using Menu.Remix.MixedUI.ValueTypes;
 using System;
+using System.Diagnostics;
 using System.Linq;
 using RWCustom;
 using UnityEngine;
@@ -21,6 +22,8 @@ public class LobbyCreateMenu : SmartMenu
     private OpTextBox passwordInputBox;
     private OpCheckBox enablePasswordCheckbox;
     private MenuDialogBox? popupDialog;
+    private SimplerButton editSyncRequiredModsButton;
+    private SimplerButton editBannedModsButton;
     public override MenuScene.SceneID GetScene => ModManager.MMF ? manager.rainWorld.options.subBackground : MenuScene.SceneID.Landscape_SU;
 
     public LobbyCreateMenu(ProcessManager manager) : base(manager, RainMeadow.Ext_ProcessID.LobbyCreateMenu)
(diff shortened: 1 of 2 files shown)


forthfora:
Waiting on #700 to be merged first

TheLazyCowboy1:
I'm 100% not a UI person, but if I was in charge of Meadow's UI, I'd probably request that the buttons get moved to the left side of the screen. Because being in the same list as "visibility" and "password" makes it seem like it's equally important for users to customize, when really it's a more advanced config. But my opinion doesn't matter (and it shouldn't) regarding these matters!

forthfora:
Did a brief test and it should be good to go

The comment about the positioning is fair btw, if 2 or more people would like to see it moved then I'll mess around with it a bit more

6fears7:
The position on this makes a lot of sense. Most users assume we have outright banned certain mods and this lowers the barrier of entry to   helping them understand how to leverage the tool.

Have you tested this functionality while the game is in full screen?

forthfora:
> The position on this makes a lot of sense. Most users assume we have outright banned certain mods and this lowers the barrier of entry to helping them understand how to leverage the tool.
> 
> Have you tested this functionality while the game is in full screen?

Seemed fine for me on both windowed and full-screen at several different resolutions

Only concern I can think of would be how opening the files functions on other OSes, I've only tested on Windows, I don't see any reason it shouldn't work on Linux at least though

Menu/LobbyCreateMenu.cs
```diff
@@ -109,6 +112,26 @@ public LobbyCreateMenu(ProcessManager manager) : base(manager, RainMeadow.Ext_Pr
...
         }
+
+        var editSyncBannedLabel = new ProperlyAlignedMenuLabel(this, mainPage, Custom.ReplaceLineDelimeters(Translate("Control which mods are permitted on clients by editing the files below.<LINE>Instructions included within.")), new Vector2(500f, 325f), new Vector2(200, 20f), false);
+        mainPage.subObjects.Add(editSyncBannedLabel);
+
+        editSyncRequiredModsButton = new SimplerButton(this, mainPage, Translate("Edit High-Impact Mods"), new Vector2(500f, 275f), new Vector2(150f, 30f));
+        editSyncRequiredModsButton.OnClick += _ =>
+        {
+            RainMeadowModManager.GetRequiredMods();
+            Process.Start(AssetManager.ResolveFilePath(RainMeadowModManager.SyncRequiredModsFileName));
```
invalidunits, line 123: Process can throw an exception If the file doesn't exist. Please surround in an try catch block and log the output.

invalidunits (approved):
looks good LGTM

6fears7:
@invalidunits can you confirm Linux behavior is functioning as expected?

invalidunits:
> @invalidunits can you confirm Linux behavior is functioning as expected?

I'll check when I get home (7:00 EDT). Linux behavior shouldn't be a blocker though, Proton constantly improving and changing.

6fears7:
Ok. Works for me. If there's an issue can roll this back but I'm good to merge.

PR #737, opened 2025-03-11
RPC Error Prevention

TheLazyCowboy1:
Adds:
1. The **option** to have a byte in RPCs that specify their length, so that erroneous or unrecognized RPCs don't cause serialization of the entire packet to fail (and apparently crashes as a result...)
2. (Unimportant: Adds the option for RPCs to specify a particular index, so they can possibly avoid load-order dependency)

### Why is this important?
1. Crashes are happening already; this'll fix that.
2. People clearly want soft-compatible mods like Push To Meow. Admittedly, this isn't *important*, but it is desired.
3. Overall adds an extra layer of safety and stability for a quite cheap cost.

### Why is the length byte enabled by default?
Feel free to alter this if it's a total hold-up for you, but I think the logic is simple: RPCs should check for errors *unless otherwise specified*. If anyone (especially a modder) is implementing an RPC, we should want it to have some error prevention unless it is explicitly stated not to (e.g: It is thoroughly tested and reliable).
RPCs that are important to optimize in Rain Meadow should be specified as optimized; or if necessary, all RPCs in Meadow can be changed to skip error correction by a simple find/replace action in the code.

### Why not make this a separate system?
1. Do you really want to maintain two separate events that do exactly the same thing, except one has an extra byte?
2. This encourages stability all around. *Apparently*, people will make their own mods with RPCs that can break the game. This will make them have to go out of their way to send erroneous RPCs. Modders will probably use whatever method is most convenient, and adding a second inferior method will likely not help much.

I have extremely carefully arranged this code to be as unintrusive as possible. If there is anything that can be changed to be more accommodating, please let me know.

(Thoroughly tested locally; but marked as a draft until it can be reviewed by people who care very deeply about this system and then tested through Steam.)

### DIFF
--- a/Online/RPCManager.cs
+++ b/Online/RPCManager.cs
@@ -9,10 +9,17 @@ namespace RainMeadow
     // what sort of options are needed here?
     // "if target owner only"
     // "if resource active"
+    /// <summary>
+    /// runDeferred: If true, the RPC is processed after states are processed (at the end of the network-frame)
+    /// customIndex: The unique index of the RPC. If 0, will determine an unused index automatically. (MODS: Use a customIndex above 1000 please!!)
+    /// useErrorCorrection: If true, unrecognized RPCs and RPCs with erroneous data will be safely ignored. However, RPC size is limited to 255 bytes.
+    /// </summary>
     [AttributeUsage(AttributeTargets.Method)]
     public class RPCMethodAttribute : Attribute
     {
         public bool runDeferred; // run after state is processed, at end of network-frame
+        public ushort customIndex = 0; //used to specify the index, to prevent clients from having different indices, or 0 if determined by RegisterRPCs()
+        public bool useErrorCorrection = false;
     }
 
     public static class RPCManager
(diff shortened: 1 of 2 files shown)


Online/RPCManager.cs
```diff
@@ -30,40 +40,44 @@ public class RPCDefinition
...
+            RegisterRPCsInAssembly(mainAssembly, true);
+
+            RainMeadow.Debug("Registering RPCs for other assemblies");
             foreach (var assembly in AppDomain.CurrentDomain.GetAssemblies().ToList())
             {
-                bool isMain = assembly == Assembly.GetExecutingAssembly();
+                if (assembly != mainAssembly)
+                    RegisterRPCsInAssembly(assembly, false);
+            }
+        }
```
TheLazyCowboy1, line 61: I also split up the order in which assemblies are processed. Rain Meadow's RPCs are now always registered first.
Logic: No mod should ever be changing the indices of Meadow's RPCs. That's a recipe for disaster.

TheLazyCowboy1:
*sigh*...

I just don't want another PR to go down the drain. I've put a lot of very careful engineering and testing into this one.
But I very seriously doubt I can find the slightest way to improve its current implementation (on my own, at least), so I guess I have to mark it ready for review...

henpemaz:
closing in favor of #759

PR #452, opened 2024-11-14
Data chunk framework

henpemaz:

PR #738, opened 2025-03-13
Ext Enum Sync by Choc

TheLazyCowboy1:
Prevents most issues due to ExtEnums having different indices.

Currently marked as a draft because I'm not confident about the quality of my changes to Choc's code. Unfortunately, I mostly just tried random changes until it worked. I think the overall result is decent functionality-wise, but I still want to do some more testing with this.

### Why not just ensure mod load orders are the same?
According to a group of stubborn mod testers in the Rain Meadow Discord, identical mod load order does not guarantee identical ExtEnum index order. It seems that if mods are applied at different times or reordered, the enums can become jumbled up. While ensuring mod load order is undoubtedly important, this adds an extra level of stability to the game.

@youbitchoc Thoughts? (Your code did end up throwing critical errors in like 4 different places, so sorry for mangling it so severely.)

### DIFF
--- a/Online/Resource/Lobby.cs
+++ b/Online/Resource/Lobby.cs
@@ -99,16 +99,10 @@ namespace RainMeadow
         {
             return ExtEnumBase.valueDictionary.First(kvp => EnumTypeToID(kvp.Key) == shortName).Key;
         }
-        //adds a lazy hash of the assembly name, just in case
+        //appends the assembly name, just in case (shouldn't ever be necessary; just a precaution)
         private static string EnumTypeToID(Type T)
         {
-            return T.ToString() + "" + HashString(T.Assembly.FullName);
-        }
-        private static string HashString(string s)
-        {
-            int value = 0;
-            for (int i = 0; i < s.Length; i++) value ^= (s[i] << (i & 15)) | (s[i] >>> (16 - (i & 15)));
-            return value.ToString().Substring(0, 4); //keep the string short; it shouldn't ever be necessary, hopefully
+            return T.ToString() + ":" + T.Assembly.FullName;
         }
 
         public void RequestLobby(string? key)
--- a/Online/Resource/Lobby.cs
+++ b/Online/Resource/Lobby.cs
@@ -59,20 +59,33 @@ namespace RainMeadow
                 (configurableBools, configurableFloats, configurableInts) = OnlineGameMode.GetHostRemixSettings(this.gameMode);
 
                 //determine ExtEnum values to be used within the lobby
-                this.strEnumMap = ExtEnumBase.valueDictionary.ToDictionary(kvp => kvp.Key.ToString(), kvp => kvp.Value.entries.ToArray());
-                RainMeadow.Debug(string.Join("\n", strEnumMap.Select(kvp => $"{kvp.Key}: [ {string.Join(", ", kvp.Value)} ]")));
+                //only sync ExtEnums with less than 256 entries; and give an error message for enums that cannot be synced
+                this.strEnumMap = new(ExtEnumBase.valueDictionary.Count);
+                foreach (var kvp in ExtEnumBase.valueDictionary)
+                {
+                    if (kvp.Value.Count > 255) RainMeadow.Error($"More than 255 entries for ExtEnum {kvp.Key}");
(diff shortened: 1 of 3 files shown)


Online/Resource/Lobby.cs
```diff
@@ -55,15 +58,57 @@ public Lobby(OnlineGameMode.OnlineGameModeType mode, OnlinePlayer owner, string?
...
-
-
-
+        }
+        //the full type is quite excessive, so instead we will use the short type name and loop through all the options
+        private static Type GetExtEnumType(string shortName)
+        {
+            return ExtEnumBase.valueDictionary.First(kvp => EnumTypeToID(kvp.Key) == shortName).Key;
+        }
+        //adds a lazy hash of the assembly name, just in case
```
youbitchoc, line 102: I don't trust hashes but if it works with the big mods then sure

Online/Serialization/Serializer.Classes.cs
```diff
@@ -5,19 +5,23 @@ namespace RainMeadow
...
 #endif
             }
             if (IsReading)
             {
-                extEnum = (T)Activator.CreateInstance(typeof(T), new object[] { ExtEnum<T>.values.GetEntry(reader.ReadByte()), false });
+                int idx = reader.ReadByte();
+                extEnum = (T)Activator.CreateInstance(typeof(T),
+                    new object[] { ExtEnum<T>.values.GetEntry(
+                        OnlineManager.lobby.enumMapToLocal.TryGetValue(typeof(T), out var map) ? map[idx]
+                        : idx //default to the idx given... this isn't a nice solution, but it's a fallback
```
youbitchoc, line 23: log if fail
it'll probably blow up

youbitchoc (changes requested):
log extenum serialize fail

Online/Resource/Lobby.cs
TheLazyCowboy1: The problem is simply: If two mods (including Rain Meadow!) have ExtEnums named the same thing, it's quite possible for them to conflict. Adding 4 digits of an assembly hash decreases this already extremely-low chance to a basically non-existent one.
...Honestly, the hash might be overkill. It's anticipating an unlikely problem. But I hope it's probably still better than just letting the problem break stuff if it happens...?

Online/Serialization/Serializer.Classes.cs
TheLazyCowboy1: I'll add the log, yeah.

> it'll probably blow up

I only added this check because it WAS throwing errors without it. Apparently LobbyState has an ExtEnum somewhere within its subresources... And since LobbyState has to serialize before its ExtEnums can be synced, this error would occur first and break everything.
I don't know which ExtEnum is being serialized here, but hopefully it's an unimportant one...
I wish there was a way to avoid this and sync ExtEnums first, but it would probably be a lot more complicated.

youbitchoc: > LobbyState has to serialize before its ExtEnums can be synced

at this point map is still uninitialized

throw error if fail _after_ init

Online/Resource/Lobby.cs
youbitchoc: never trust chance

just fully qualify the extenums and abort if collision

TheLazyCowboy1: That's what I originally did. ...it was throwing errors due to breaking the max buffer size, lol.
No decent mod should have its ExtEnums named something very generic. This just adds a slight layer of protection against it.
That said, I'd like to convert it to a single char, because it'd actually cover more hashes.

youbitchoc: please stop trusting hashes to resolve collisions

> No decent mod should have its ExtEnums named something very generic

don't rely on this

if the packets are too big then compress the string or use #452

henpemaz:
> According to a group of stubborn mod testers in the Rain Meadow Discord, identical mod load order does not guarantee identical ExtEnum index order. It seems that if mods are applied at different times or reordered, the enums can become jumbled up.

Another front to handle this would be to require a restart for enabling/disabling msc/expedition and going online, they should be the main culprits for misaligned enums.

PR #739, opened 2025-03-13
Deflate Compression on Every Packet

TheLazyCowboy1:
This is very much a draft: I am still unsure how much this impacts the CPU.
However, functionality-wise, this is pretty much finished. The only choice is whether to use LZ4 or Deflate.

Current observations:
* Halves the amount of data sent (huge bonus for folks with low upload speeds, like me!).
* Definitely has a negative impact on CPU performance,
* But I genuinely cannot tell the difference during normal gameplay.

This really just needs: Some very good profiling and playtesting to determine if people think performance has decreased. Unfortunately, I am very bad at supplying either of those...

### DIFF
--- a/Online/Matchmaking/UdpPeer.cs
+++ b/Online/Matchmaking/UdpPeer.cs
@@ -712,6 +712,7 @@ namespace RainMeadow
 
 
         public void Send(byte[] packet, IPEndPoint endPoint, PacketType packet_type = PacketType.Reliable, bool begin_conversation = false) {
+            //packet = Compression.CompressBytes(packet); //done in SendRaw instead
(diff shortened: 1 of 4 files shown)


TheLazyCowboy1:
Conclusion: Deflate is too CPU-intensive. This PR is only being kept open for a little longer to serve as a reference for anyone wanting to implement LZ4 compression, which is WAY faster and definitely worth implementing.

But when it comes to finding a way to incorporate 3rd party algorithms, I'm out of my comfort zone, so I would really appreciate it if someone else could help me attempt it properly.

(I got a weird working solution a while ago, but it required a bunch of extra assemblies like System.Theading.Tasks.Extensions.dll, which had to the put in the plugins folder, which didn't seem right.)

PR #741, opened 2025-03-13
Infrequent State Sends

TheLazyCowboy1:
**THIS IS JUST A TESTING DRAFT! I AM NOT YET SUGGESTING THAT WE ACTUALLY IMPLEMENT THIS YET!!**

I've been getting annoyed having Rain Meadow get super glitchy in Five Pebbles due to the sheer quantity of neurons. That's a _lot_ of data that's being synced every tick.

So I had the basic idea of making states for distant realized objects sync less often. I haven't tested it very much, but it's quite difficult to tell any different in-game, because every object in the same room within 20 tiles is still synced every frame.

If this does get implemented, I would suggest it as an optional thing in the Remix menu. Maybe there could be an entire tab called "Optimization" with options like this that might slightly help those with poor internet.

Oh, and one last idea: What if instead of grouping all OnlineFields with unspecified groups into the same group called "default", we give each one with an unspecified name its own unique group name. This (from testing) significantly increases the number of deltas per State, but also then decreases the amount of data sent overall. Just an idea, though.

### DIFF
--- a/Online/Entity/EntityFeed.cs
+++ b/Online/Entity/EntityFeed.cs
@@ -11,6 +11,8 @@ namespace RainMeadow
         public OnlinePlayer player;
         public Queue<OnlineStateMessage> OutgoingStates = new(32);
         public OnlineEntity.EntityState lastAcknoledgedState;
+        private float cooldown = 1f;
+        private const float FULL_STATE_OPTIMIZATION = 0.5f;
 
         public EntityFeed(OnlineResource resource, OnlineEntity oe)
         {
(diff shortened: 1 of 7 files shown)


invalidunits:
Would be best to limit this feature to specifically Neuron flies in Five Pebbles. I can imagine it breaking in other features.

TheLazyCowboy1:
Well it's also a problem in Subterranean, etc. Neuron flies are just the best example of why it gets ridiculous and problematic.

> I can imagine it breaking in other features.

The intention of this (highly unfinished totally a draft) system is to simulate lag. If something works despite latency/lag/packet loss, it'll work the same here.
Why "simulate lag"? The basic idea is that IF a player lacks sufficient upload bandwidth (or even processing power) to send everything, then which data gets sent should be managed. The more important things (like an angry lizard right next to me) should be fully synced, but a pile of scavs or neurons in the other room can be skipped occasionally.
So what this system really needs is to be tied to something that detects how "laggy" things are. Thus, if there is no lag, this doesn't kick in; but if there is lag, then unimportant states get sent less often.

...this PR does need to be reworked, though... it shouldn't be attached to states themselves, but rather to resources.

PR #744, opened 2025-03-17
use ushort for port numbers (UdpPeer.cs)

Lizzie841:
[Info   :RainMeadow] 01:22:08|93769|UdpPeer.GetEndPointByName:Invalid port format: 64724
[Info   :RainMeadow] 01:22:56|142150|UdpPeer.GetEndPointByName:Invalid port format: 64724

ports go from 0 to 0xffff not from -0x7ffff to 0x7ffff

### DIFF
--- a/Online/Matchmaking/UdpPeer.cs
+++ b/Online/Matchmaking/UdpPeer.cs
@@ -702,7 +702,7 @@ namespace RainMeadow
                 }
             }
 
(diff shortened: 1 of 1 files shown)


invalidunits (approved):
Nice catch!

henpemaz (approved):
Good first pr 🎉

PR #745, opened 2025-03-17
allow long ip addresses

Lizzie841:
very annoying to have to type the resolved DNS ip instead of the domain name


6fears7:
@invalidunits anything else you think needs adding here?

invalidunits:
Looks good, but I'm worried people might use ipv6, which isn't explicitly supported. 
I'll fix this code up and properly ban or add support to ipv6 in #727  

Merging

PR #747, opened 2025-03-19
shortcut to hide chat with esc (ChatHud.cs)

Lizzie841:
untested; should work

### DIFF
--- a/OnlineUIComponents/ChatHud.cs
+++ b/OnlineUIComponents/ChatHud.cs
@@ -97,6 +97,11 @@ namespace RainMeadow
                     }
                 }
             }
+            else if (Input.GetKeyDown(KeyCode.Escape) && chatInputOverlay is not null)
+            {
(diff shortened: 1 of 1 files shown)


6fears7 (commented):
@SaddestUnderscore you cool with this?

henpemaz:
I saw some requests around to be able to "abort" something you're typing with ESC. Is this about that? And if so, does it properly clear out the input box?

SaddestUnderscore:
could you record a video to show? not certain what this fully does

6fears7:
Tested this. ESC will close out the chatinputbox and open the menu meaning the following:

1. This code is redundant
2. If we want to support clearing out the text chat (if a user feels like rewriting a message) we need to override the default impl of opening the menu with ESC and then probably add the string representation the ESC key in ChatTextBox.cs and set the lastSentMessage to empty string, and if string already empty then close the ChatTextBox

PR #748, opened 2025-03-20
fix rivulet ending softlock

Lizzie841:
not the best solution, but atleast it doesnt softlock


henpemaz:
not a blocker but would be nice to have basic sanity checks when receiving the RPCs, namely that it's coming from the lobby owner. if you add an `RPCEvent` arg it's auto-filled and you'll be able to check rpc.from (`RPCEvent.currentRPCEvent` also works).

6fears7 (approved):

PR #749, opened 2025-03-20
Brotli compression for every packet

Lizzie841:
instead of deflate, brotli works with better speed, doesnt require native dlls
based on https://github.com/henpemaz/Rain-Meadow/pull/739 work


henpemaz:
Personally I don't think blindly compressing every packet is the way to go. Rainworld is CPU bound, runs relatively well on lower-end pcs and the rest of the mod is already fairly heavy. Significantly increasing CPU usage is a tradeoff that needs to be carefully picked.

There's already a lot going on towards lowering the bandwidth, mainly delta-encoding (already acknowledged data isn't re-sent), and then also selective compression of large state-objects, for any state-object that amounts over 4kb in itself, which is usually world-session with lots of new creatures/items "definition" entries at the start of session and therefore easily compressible.

Imo, if there was a "clog detection" or general "lag detection" mechanism in place to detect significant lag of oneself or lag of specific players, then turning the knob of cpu x bandwidth in those scenarios would make sense. But not blindly.

Lizzie841:
thinking about it, probably lz4 would fare better anyways

PR #750, opened 2025-03-20
enforce total method ordering (RPCManager.cs)

Lizzie841:
based on
https://stackoverflow.com/questions/62036194/is-order-of-reflection-deterministic
https://stackoverflow.com/questions/7000711/what-is-the-order-of-returned-types-by-assembly-gettypes


henpemaz:
Any scenario where the current stuff fails or just trying to future-proof?

Online/RPCManager.cs
```diff
@@ -36,12 +36,16 @@ public static void SetupRPCs()
...
+            var assemblyList = AppDomain.CurrentDomain.GetAssemblies().ToList();
+            assemblyList.Sort((a, b) => String.Compare(a.GetName().FullName, b.GetName().FullName));
+            foreach (var assembly in assemblyList)
             {
                 bool isMain = assembly == Assembly.GetExecutingAssembly();
                 try
                 {
-                    foreach (var type in assembly.GetTypesSafely().ToList())
+                    var typeList = assembly.GetTypesSafely().ToList();
+                    typeList.Sort((a, b) => String.Compare(a.FullName, b.FullName));
```
invalidunits, line 47: `String.Compare(String, String)` might be platform dependent.
Please use `String.Compare(String, String, StringComparison.Ordinal)` when comparing strings not displayed to the user.

invalidunits (changes requested):
Looks good so far.

TheLazyCowboy1:
Has there been any issues regarding this so far? I don't think anyone has reported issues with RPCs being out of sync yet.

My PRs have been designed to solve two related problems:
1. If someone sends an RPC that other players don't recognize, the game breaks (e.g: Push to Meow has been crashing everyone's games). Ideally, this shouldn't be a problem if everyone has the same mod list; and as far as I know, it's not.
2. If any RPC somehow has any error (probably through a modded RPC doing something stupid), the game also breaks.

Just to be clear: I am 100% not against this. This seems like a good fallback to ensure everything is ordered properly and prevent possible bugs. But as far as I know (my knowledge is quite limited), no one has encountered these bugs so far. Still probably good to prepare for the future, though!

invalidunits (dismissed):
Looks good.

PR #751, opened 2025-03-20
increment peer timeout time from 3s to 15s (UdpPeer.cs)

Lizzie841:
should fix the many disconnects while playing over unstable connections

### DIFF
--- a/Online/Matchmaking/UdpPeer.cs
+++ b/Online/Matchmaking/UdpPeer.cs
@@ -742,7 +742,7 @@ namespace RainMeadow
             }
         }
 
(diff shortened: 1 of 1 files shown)


henpemaz:
@invalidunits this is your domain (should we update codeowners for this?)

invalidunits:
From what I've tested, 3s is very generous for the timeout. And it's important that the timeout is low to enable rejoining without errors or race conditions.

I'd rather have this as a remix setting with 3s as the default. 
Closing in favor of adding Peer Timeout and Heartbeat Time to the remix settings.

PR #753, opened 2025-03-20
udp timeout now as an option

Lizzie841:
now as an option
image


invalidunits:
Please add HeartBeat as an option as well. If the timeout is less than the heartbeat, It's impossible to join or host any lobby.

invalidunits (approved):
Looks good!

henpemaz:
All your PRs seem to have tabulation issues, do you mind sorting that?

invalidunits (approved):

PR #756, opened 2025-03-21
oxipng + optivorbis

Lizzie841:
optimizing assets with oxipng and optivorbis, shaves off a grandiose amount of 10 MB total


Mod/scenes/oxipng
henpemaz, line 1: not needed, right?

henpemaz:
I think you left a spurious file in there, otherwise pretty cool.

If possible amend and force-push the deletion so the file isn't left in the repo.

henpemaz (approved):

PR #759, opened 2025-03-21
Soft RPCs, rpc reordering and version bump

henpemaz:
Sends a module GUI and ushort length that is used to skip the message body if the receiver does not have the module, takes a separate event index (we have plenty of those left). No changes to base RPCs other than ensuring that meadow ones are loaded first.

### DIFF
--- a/Mod/modinfo.json
+++ b/Mod/modinfo.json
@@ -1,7 +1,7 @@
 {
   "id": "henpemaz_rainmeadow",
   "name": "Rain Meadow",
-  "version": "0.1.2.1",
+  "version": "0.1.3.01",
   "target_game_version": "v1.9.15",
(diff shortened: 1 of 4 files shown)


TheLazyCowboy1:
This won't fix anything unless all mods with RPCs are marked as high impact. (Which I do have a PR for, but it still needs testing; so that's a moot point). I would highly recommend at least making the normal RPCMethod internal, in order to discourage mods from using it.

(Renaming the old RPCMethodAttribute would be convenient to modders so they don't have to edit all their RPCs, but it's not necessary.)

We want to babyproof this system, because we cannot trust that we can control every Rain Meadow mod.

henpemaz:
> This won't fix anything unless all mods with RPCs are marked as high impact.

Which they should. There's already a simple way to do it with including a rainmeadow.json in the mod. And if they don't want to, then they can stick to SoftRPCs.

> We want to babyproof this system

No, we don't. Making mods isn't for idiots and needs no babyproofing.

Mod/modinfo.json
```diff
@@ -1,7 +1,7 @@
 {
   "id": "henpemaz_rainmeadow",
   "name": "Rain Meadow",
-  "version": "0.1.2.1",
+  "version": "0.1.3.01",
```
invalidunits, line 4: Is this a typo? It should be 0.1.3.0 or 0.1.3.1

TheLazyCowboy1:
> Making mods isn't for idiots and needs no babyproofing.

I have made mods for Rain Meadow that use RPCs, therefore this is objectively false. (Some mod makers can be unintelligent, inexperienced, or careless).
The need (as Push to Meow has taught us) is not encouraging competent modders to make stable mods (although this is a good thing), but rather to prevent inexperienced modders from ruining public lobbies. Someone looked at how Meadow does RPCs, copied that method of implementation, and uploaded the mod. I don't think he has a clue what high-impact mods are. The result is that public lobbies crash a lot whenever someone joins with the mod.

...yeah, just banning all mods with RPCs should do the trick. So basically the usefulness of this PR depends on #723 

...and the implementation here is a little inefficient and disorganized, but so long as it causes no issues, it's probably fine.

Mod/modinfo.json
henpemaz: I usually do like this for release-candidates, then remove the final digit for actual releases

PR #761, opened 2025-03-21
Move modsync buttons to remix, reorder remix tabs, reorg main tab

henpemaz:


henpemaz:
@forthfora looks ok?

Game/MeadowRemixOptions.cs
```diff
@@ -179,6 +184,17 @@ public override void Initialize()
             introroll.OnValueChanged += (UIconfig config, string value, string oldValue) => { if (value == "Downpour" && introroll.Menu.manager.rainWorld.dlcVersion == 0) downpourWarning.Show(); else downpourWarning.Hide(); };
             downpourWarning.Hidden = PickedIntroRoll.Value != IntroRoll.Downpour && introroll.Menu.manager.rainWorld.dlcVersion == 0;
 
+            editSyncRequiredModsButton.OnClick += _ =>
+            {
+                RainMeadowModManager.GetRequiredMods();
+                System.Diagnostics.Process.Start(AssetManager.ResolveFilePath(RainMeadowModManager.SyncRequiredModsFileName));
```
forthfora, line 190: Process.Start may be worth wrapping in a try catch in-case it fails for some reason, the file should always exist from calling GetRequiredMods though

Game/MeadowRemixOptions.cs
```diff
@@ -179,6 +184,17 @@ public override void Initialize()
...
+            };
+
+            editBannedModsButton.OnClick += _ =>
+            {
+                RainMeadowModManager.GetBannedMods();
+                System.Diagnostics.Process.Start(AssetManager.ResolveFilePath(RainMeadowModManager.BannedOnlineModsFileName));
```
forthfora, line 196: same as above ^

forthfora:
took a look in-game, seems fine yep

see my comment for a quick suggestion (invalidunits pointed this out in the last PR actually but it got merged before I could make the change)

PR #764, opened 2025-03-21
fix StartJoiningLobby (by tlc)

Lizzie841:


6fears7 (approved):

6fears7:
DirectJoin expects checkMods to be false, works for me.

PR #776, opened 2025-03-25
Make creature names translatable

zbiotr:
Makes the names of the creatures to be able to be translated in death messages.
This also translates in Spanish every possible one that may appear either by killing them or dying from them.

(Unrelated, but this PR translates a message regarding Inv's campaign as well.)


henpemaz:
There's a conflict (people were reworking some of the system messages) otherwise looks good

henpemaz (approved):

PR #779, opened 2025-03-26
Rain Meadow being disable by mod applier fix & Add option to opt out of mod applier

invalidunits:


invalidunits:
Drafting this as I'm not really set on the Idea of allowing players to avoid banned mods because they clicked a button in the settings menu. I added a warning but I'm not sure that's enough.

6fears7:
Having it as a toggleable makes sense to me. 

More info on why meadow was listed as a mod to manage: 

https://github.com/henpemaz/Rain-Meadow/pull/708#issuecomment-2676419819

6fears7:
I would like this merged. We used it this morning to fix an issue where a player hosting gave me a mod mismatch with rain meadow and trying to apply it just disabled it entirely.

henpemaz:
Not a fan of making it too easy (the button thing) to bypass a mod mismatch, the game will be broken 90% of the time for the client joining like that

invalidunits:
Closing in favour of #798

PR #786, opened 2025-03-31
Arena: re-use nightcat and remove nc if watcher active

6fears7:


6fears7:
Tested with Watcher active and deactive

PR #788, opened 2025-04-01
Fix SlugSlamConditions

invalidunits:


6fears7:
https://www.youtube.com/watch?v=J9FImc2LOr8

6fears7:
I don't think we need to update Weapon_HitThisObject too but just raising awareness:

```
            if (!obj.FriendlyFireSafetyCandidate() && obj is Player && self is Spear && self.thrownBy != null && self.thrownBy is Player)
            {
                return true;
            }
```

6fears7:
LGTM, merge at your discretion

6fears7 (approved):

PR #795, opened 2025-04-06
Add arena color config

Dustylemny:
Added a color config which also replace getting colors from rain meadow options to getting colors from rw saves
this is how the color config looks now, uses remix sliders so the button to open it only appears when remix is on
Screenshot (96)
colors in action
Screenshot (97)


6fears7 (commented):
Looks awesome, I'll give it a whirl later

6fears7 (approved):
Tested, works, pretty slick.

6fears7:
passing to @henpemaz due to additions to Menu/, Dialog/ and Object/ contents

henpemaz (approved):

PR #796, opened 2025-04-06
do not invalidate JSON upon invalid characters

Lizzie841:
would help with more creatures pack obliterating the meadow progress data

### DIFF
--- a/Meadow/MeadowProgression.cs
+++ b/Meadow/MeadowProgression.cs
@@ -695,7 +695,7 @@ namespace RainMeadow
             progressionData.characterUnlockProgress++;
             if (progressionData.characterUnlockProgress >= characterProgressTreshold)
             {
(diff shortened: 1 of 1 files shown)


Meadow/MeadowProgression.cs
```diff
@@ -695,7 +695,7 @@ public static Character CharacterProgress()
             progressionData.characterUnlockProgress++;
             if (progressionData.characterUnlockProgress >= characterProgressTreshold)
             {
-                if (NextUnlockableCharacter() is Character character)
+                if (NextUnlockableCharacter() is Character character && character != null)
```
henpemaz, line 698: null already fails the `is` constraint

Lizzie841: mb

henpemaz (approved):

PR #802, opened 2025-04-12
Implement Button Scroller to spectator list and double press for kicking

Dustylemny:
Also includes adding indicators when the host presses the kick button once (ex: player's text becoming red, a text popping up and sound)
and removes the double sprites which i can see vaguely as the host sometimes

### DIFF
--- a/Menu/Objects/ButtonScroller.cs
+++ b/Menu/Objects/ButtonScroller.cs
@@ -133,13 +133,21 @@ namespace RainMeadow
         }
         public void RemoveButton(int index, bool constrainScroll)
         {
-            if (buttons.Count > index)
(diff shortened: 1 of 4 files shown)


6fears7:
Only thing that's a little weird is how the button jumps to the pos initially when spectator hud opens. Other than that, love it.

Dustylemny:
@6fears7 wdym by that?

Dustylemny:
ill see what i can do to ease the difference

Dustylemny:
@6fears7 hopefully this should the button jumping to the pos or at least makes it less jarring

6fears7:
@Dustylemny this ready for review?

Dustylemny:
@6fears7 mhm

6fears7:
very slick

PR #803, opened 2025-04-13
add watcher fast travel to list of banned mods

Lizzie841:
mainly to prevent sync issues with watcher campaigns, separate pr cuz also seems to affect normal(?)

### DIFF
--- a/Mod/RainMeadow.json
+++ b/Mod/RainMeadow.json
@@ -16,6 +16,7 @@
         "henpemaz_spawnmenu",
         "autodestruct",
         "DieButton",
+        "CuteFatalis.WacherFastTravel",
         "emeralds_features",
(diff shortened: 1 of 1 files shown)


6fears7 (approved):

henpemaz:
This change doesn't affect the host, only clients.

Shouldn't it instead be added to the incompatible mods list?

6fears7:
> This change doesn't affect the host, only clients.
> 
> Shouldn't it instead be added to the incompatible mods list?

@Lizzie841

Lizzie841:
> > This change doesn't affect the host, only clients.
> > Shouldn't it instead be added to the incompatible mods list?
> 
> @Lizzie841

in such case - itd be probably worth to show the player such mod is incompatible, as a warning perhaps?

PR #804, opened 2025-04-14
prevent possible dupes at RWGAME creation

Lizzie841:
a solution to prevent duplicate toys in WAUA_TOYS is similar to this, so i asked myself: "Why not in RainWorldGame ctor", and then i did.
may be the wrong approach, does seem to work somewhat fine through (haven't seen dupe batfruits for exampleh)

### DIFF
--- a/Game/RainMeadow.GameHooks.cs
+++ b/Game/RainMeadow.GameHooks.cs
@@ -72,6 +72,7 @@ namespace RainMeadow
             {
                 var c = new ILCursor(il);
                 var skip = il.DefineLabel();
+                // pole mimics are the last AbstractCreature to be created, whereas pink lizards are the first
                 ILLabel pmLoop = null;
                 c.GotoNext(moveType: MoveType.After,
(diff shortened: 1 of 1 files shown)


6fears7:
Will be good to run a longer playtesting session and see what kicks back, especially at 5p save or outskirts.

6fears7:
https://github.com/henpemaz/Rain-Meadow/blob/5a058bcadffab5fedc97b3e25c312741b29b78b3/GameModes/StoryGameMode.cs#L81 Wondering then if this is broken.

6fears7:
I need to review the CIL's C# code but I think ShouldLoadCreatures is working as intended and this IL is duplicating work. The other dupes likely fall outside of this scope. If you feel otherwise, let me know (and bonus points for video-evidence!)

henpemaz:
Can you add a comment to the IL hook explaining what it's doing? I don't feel like reverse engineering it

PR #805, opened 2025-04-14
Unsync Neurons Pos

6fears7:


6fears7:
Looking into ShouldSyncAPO at all for neurons. Could do a regional check. Not syncing them in 5P cuts latency down from 50ms (locally!) to 3ms

invalidunits:
> Not syncing them in 5P cuts latency down from 50ms (locally!) to 3ms
That's Insane 😵‍💫

Has anyone experimented with lowering the frequency of sent states? Sending APO states every 10 frames might save us a ton of bandwidth and latency with less game play sacrifices.

6fears7:
> > Not syncing them in 5P cuts latency down from 50ms (locally!) to 3ms
> 
> That's Insane 😵‍💫
> 
> Has anyone experimented with lowering the frequency of sent states? Sending APO states every 10 frames might save us a ton of bandwidth and latency with less game play sacrifices.

If you wanna point out where we manage this it will save me some repo searching. 

However, I'm not sure if going that high in the abstraction would be helpful over a "Sync in Room / Sync in World / Register APO" attribute that we have in OnlineGameMode. 

Going to also see about not syncing them in world. Too much time is spent loading a ton of neurons that we really shouldn't need to care about from host. 

#804 might impact client's ability to have any local entities which could disparately impact some of our unsynced APOs as well.

6fears7:
This PR partially helps, still have tremendous room session zipping in 5p. Wondering if we can work it so that we can mark specific APOs as client-side management only. No registration, no syncing. Clients spawn and simulate themselves. Also thinking about locusts in Watcher :skull:

invalidunits:
> Also thinking about locusts in Watcher 💀

Thankfully locusts in the watcher do not use the APO system at all. We shouldn't need to sync or make any changes for them to work.

PR #809, opened 2025-04-16
do not recreate slugcats list each time we call function, just cache it and redo it when initializing

Lizzie841:
its unescesary

### DIFF
--- a/Arena/ArenaHelpers.cs
+++ b/Arena/ArenaHelpers.cs
@@ -5,19 +5,19 @@ namespace RainMeadow
 {
     public static class ArenaHelpers
     {
-        public static List<SlugcatStats.Name> allSlugcats = null;
-        public static List<SlugcatStats.Name> baseGameSlugcats = null;
-        public static List<SlugcatStats.Name> vanillaSlugcats = null;
-        public static List<SlugcatStats.Name> mscSlugcats = null;
+        public static List<SlugcatStats.Name> allSlugcats = new List<SlugcatStats.Name>();
+        public static List<SlugcatStats.Name> baseGameSlugcats = new List<SlugcatStats.Name>();
+        public static List<SlugcatStats.Name> vanillaSlugcats = new List<SlugcatStats.Name>();
+        public static List<SlugcatStats.Name> mscSlugcats = new List<SlugcatStats.Name>();
         public static readonly List<string> nonArenaSlugs = new List<string> { "MeadowOnline", "MeadowOnlineRemote" };
 
         public static void InitializeBuiltinTypes()
         {
             // reinitialize
-            vanillaSlugcats = new List<SlugcatStats.Name>();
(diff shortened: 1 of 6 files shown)


6fears7:
LGTM will resolve conflicts later

6fears7 (dismissed):

6fears7:
@Lizzie841 one last release please ;)

henpemaz:
1. you changed the line endings of Arena/ArenaLobbyMenu.cs and it picks up the entire file as changed, can you fix that please

2. MSC can be turned on and off at will without restarting the game, caching enums is problematic, imo the best way around it is to build those lookups on arenasession ctor

Lizzie841:
> #822 makes the list a static readonly. Would you rather go that route or this?

what if someone enables MSC
activates the ctor for the static
then disables MSC

Dustylemny:
> > #822 makes the list a static readonly. Would you rather go that route or this?
> 
> what if someone enables MSC activates the ctor for the static then disables MSC
This does not affect the AllSlugcats which requires a method(checks MSC and adds mscSlugs into an instance containing vanilla slugs) AllSlugcats currently is used one time so it's fine

Lizzie841:
> #822 makes the list a static readonly. Would you rather go that route or this?



> > > #822 makes the list a static readonly. Would you rather go that route or this?
> > 
> > 
> > what if someone enables MSC activates the ctor for the static then disables MSC
> 
> This does not affect the AllSlugcats which requires a method(checks MSC and adds mscSlugs into an instance containing vanilla slugs) AllSlugcats currently is used in baseslugcats so it's fine if this is what you were worried about (maybe the issue may be deactivated msc, msc list loads but enabled msc later and causing the list to still have null enums)
> 
> I'll go ahead and try the scenario

point is not have a method that "redoes the list each call", just cache it, like recreating the list for something that only changes on modchange/msc/etc is not nescesary

Dustylemny:
> > #822 makes the list a static readonly. Would you rather go that route or this?
> 
> > > > #822 makes the list a static readonly. Would you rather go that route or this?
> > > 
> > > 
> > > what if someone enables MSC activates the ctor for the static then disables MSC
> > 
> > 
> > This does not affect the AllSlugcats which requires a method(checks MSC and adds mscSlugs into an instance containing vanilla slugs) AllSlugcats currently is used in baseslugcats so it's fine if this is what you were worried about (maybe the issue may be deactivated msc, msc list loads but enabled msc later and causing the list to still have null enums)
> > I'll go ahead and try the scenario
> 
> point is not have a method that "redoes the list each call", just cache it, like recreating the list for something that only changes on modchange/msc/etc is not nescesary
>
how about this, cache at arena menu ctor? the only time the list is called is on arena menu ctor (Multiplayer menu ctor itself calling Arena Image's slugcat is null so either way it will return "Multiplayer02") which is immediately cleared and replaced by user profiles

6fears7 (approved):
good stuff

PR #813, opened 2025-04-16
Add saving colors to Arena when exiting to lobby

Dustylemny:
generally tested any save errors with story mode, so far it works fine, and colors last set by player gets loaded, should also save when playing a round in arena


6fears7:
closing this and will reopen in new pr since I biffed it

PR #822, opened 2025-04-18
Force Arena Color Dialog Shutdown and Arena Menu UI cleanup

Dustylemny:


6fears7 (commented):
Loving it

Dustylemny:
holding this for now till the unique id part is settled with

Online/OnlinePlayer.cs
```diff
@@ -201,8 +200,11 @@ public bool Equals(OnlinePlayer other)
         {
             return other != null && id == other.id;
         }
-        public override int GetHashCode() => id.GetHashCode();
-
+        public override int GetHashCode() => id.GetHashCode();
+        public string GetUniqueID()
```
6fears7, line 204: @henpemaz this is the only portion that should require your approval. Method to make it easier to get Steam vs LAN IDs.

Arena/ArenaLobbyMenu.cs
Dustylemny, line 1: Um ill fix this later

Online/OnlinePlayer.cs
henpemaz: Could have definitely used just inlobbyid and a dictionary<int,whatever> but sure why not

henpemaz (approved):

6fears7 (approved):

PR #823, opened 2025-04-18
lobby chat

Lizzie841:


6fears7:
If you can upload a pic that would be swell.

6fears7 (commented):
@SaddestUnderscore any thoughts on this?

Menu/LobbySelectMenu.cs
```diff
@@ -326,6 +326,7 @@ public void StartJoiningLobby(LobbyInfo lobby, string? password = null, bool che
         
         public void RequestLobbyJoin(LobbyInfo lobby, string? password = null)
         {
+            ChatLogManager.ResetPlayerColors();
```
SaddestUnderscore, line 329: can we clear colors when we leave the lobby instead of having of having to join a new one?

SaddestUnderscore (dismissed):
apart from that, looks pretty good to me

henpemaz (approved):

6fears7 (approved):

PR #824, opened 2025-04-18
Add UniqueIDs to OnlinePlayers

Dustylemny:
this is to prevent sync issues like this
image
image
(currently with steam nicknames theres no real id to parse from which can lead to dictionary that save id errors (when two users use the same name), and OnlinePlayer.id.ToString() gets the same as id.name) 
Seen as lobby data stores "DustyLemny" but my friend nicknamed me as Lemon (tries parsing Lemon, ended up getting no value thus i was shown as surv)

this pr should fix those possible sync issues 

currently figuring out how to use lan unique ids
image
so far with using realIDs it syncs the other players when the user joins

the main goal of this is to introduce a more universal ID -> between every lobby


6fears7:
@invalidunits just want to make you aware of this. I had been assuming that MeadowPlayerId.ToString() simply converted the representation of the playerID to string and dusty pointed out its really just returning the name. So we discussed having a unique identifier that could be leveraged to get players in LAN or Steam.

Dustylemny:
@6fears7 so far steam works however the part when i said i would use the port and address for a id is not a good idea considering its not good for playtesting if i can recall that lan supports multiple devices coming from the same address

Dustylemny:
@invalidunits i hope to hear from you about this and abt LAN RealID match making because the current idea is we generate a random index for the player on LanMatchMaking ctor however i wondering if the new index can cause issues and does it require active transport similar to address or you can just generate on your own side and it will sync for other lan players?

Dustylemny:
i unfortunately dont have the required to perform LAN testing so i dont know much about it

Online/Matchmaking/LANMatchmakingManager.cs
```diff
@@ -107,19 +102,21 @@ public override bool Equals(MeadowPlayerId other)
                 }
                 return false;
             }
+            public override string RealID()
+            {
+                return name;//$"{endPoint.Port}|{string.Join(",", endPoint.Address.GetAddressBytes().Select(x => x.ToString()))}"; //idk yet
```
invalidunits, line 107: The `return name;` is a placeholder, so I won't complain about it.

The commented code might not work. While we can settle on a name for other players, we have no idea what our endpoint is or what other's are calling us. (we could ask them but that's a can of worms I don't want to open, relating to changing a lanplayerid during playtime)

See UDPPeerManager.getInterfaceAddresses() and the diagram below
lanpeer2peerimg

I wish you good luck figuring out how to implement this function. I'm not even sure myself. :/

Online/Matchmaking/SteamMatchmakingManager.cs
```diff
@@ -44,7 +44,10 @@ public override bool Equals(MeadowPlayerId other)
             {
                 return other is SteamPlayerId otherS && steamID == otherS.steamID;
             }
-
+            public override string RealID()
+            {
+                return steamID.m_SteamID.ToString();
```
invalidunits, line 49: 8 bytes becoming a string of up to 19 characters (38 bytes) is a bad deal. I'm assuming `RealID()` is for differentiating MeadowPlayerIds so byte serialization might be the way.

invalidunits (commented):
> i unfortunately dont have the required hardware to perform LAN testing so i dont know too much about it
   If you need to test LAN, you can open multiple instances of the game by running the .exe file directly.   

Does this PR have some uncommitted arena changes on your side. Only the matchmaking stuff is there, and I'd like to see how it's implemented there.

Online/Matchmaking/LANMatchmakingManager.cs
invalidunits: The diagram's unnecessary, but making stuff like this keep me motivated

6fears7:
Implementation goal: 

Use a dictionary of string, int or int, int where the key represents the unique player ID that we're thinking of making and value is  slugcat index to capture player's selected slug. 

So whatever code dusty may provide is likely to show that.

Dustylemny:
@invalidunits 
The arena clean up part will be part of a different PR, basically this how its played out
image (upon clicking you set your new slugcat index with your RealID to a dictionary)
image (other players upon joining will parse it with the RealID) - RealID should be universal in a sense

Dustylemny:
RealID isnt gonna be serialised with MeadowPlayerID or gonna change anything with equals but rather a string you can call and recieve to store a lobby data. using an integer may be a bit difficult to implement and transfer

Dustylemny:
steam already has a unique identification system so why not use that? however we need a way to support casting and string seems to be a universal cast

Dustylemny:
@invalidunits for the index idea part it's smth like this 
image

Dustylemny:
image
this is a clearer demostration on how there are sync issues

Online/Matchmaking/LANMatchmakingManager.cs
Dustylemny: the diagram is confusing, can you explain what is it trying to say?

Dustylemny:
@invalidunits just wondering how do LAN players recieve a new player? like using LanPlayerID, is it more towards syncing (using serializer) or making a new instance from the package they send?

Dustylemny:
@invalidunits so far this is how it looks like, index gets randomized from 0 to (one less than max int) from LanMatchmaking ctor

you load up with your randomised index.
upon joining it should sync your index to other players similar to your name?

Online/Matchmaking/SteamMatchmakingManager.cs
Dustylemny: doenst byte store less? do you mean like some sort of byte array?

Dustylemny:
suddenly lan works for me now so its strange, but maybe it makes sense since i made that thought back during eariler version of meadow idk either that or i am stupid earlier

Dustylemny:
image
image
so far it syncs

invalidunits:
> @invalidunits just wondering how do LAN players recieve a new player? at LanPlayerID side, is it more towards syncing (using serializer) or making a new instance from the package they send? Asking you this since you were the one who implemented LAN as far as I know.

See `void LANMatchmakingManager.AcknoledgeLANPlayer(OnlinePlayer)`
If we are the host we send a `ModifyPlayerListPacket` to every other player, which contains the player being added to the lobby. The joining player recieves a `ModifyPlayerListPacket` as list of all players in the lobby.
On recieving the packet we create a new `OnlinePlayer` for each of the endpoints and call `void LANMatchmakingManager.AcknoledgeLANPlayer(OnlinePlayer)` to add them to the player list

Dustylemny:
@invalidunits i already did it, this should be fine right? (did it for serialize and deserialize as well i checked)
image
(for modifylistpacket)

and 

image
(for RequestJoinPacket which i think its relevant as well for syncing?)

Online/LANMatchmakingManager.cs
```diff
@@ -0,0 +1,388 @@
+using System;
```
invalidunits, line 1: I'm assuming this is an unintended duplicate

Online/Matchmaking/SteamMatchmakingManager.cs
invalidunits: Yes use a BinaryWriter to write the steamID.

Online/LANMatchmakingManager.cs
Dustylemny: i dont see a double system? somehow github be tweaking and think i made 600+ line changes when i made a single change of value when editing manually on git

Dustylemny: oh wait
i see it now, thanks for informing me

Online/Matchmaking/SteamMatchmakingManager.cs
Dustylemny: hmm but theres gonna be casting issues when it comes to LAN index and for future lobbies, are you suggesting I use bytes and convert them to string? and for your question, it isnt necessary gonna be part of differenciating MeadowPlayerIds if you mean Equals() methods

Online/Matchmaking/LANMatchmakingManager.cs
```diff
@@ -107,29 +122,40 @@ public override bool Equals(MeadowPlayerId other)
...
+            public override string RealID()
+            {
+                return index.ToString();
+            }
+            public int index;
+        }
+        public LANMatchmakingManager()
+        {
+            System.Random randomizer = new();
+            PlayerIndex = randomizer.Next();
```
invalidunits, line 134: Remember to seed your random

Online/Matchmaking/SteamMatchmakingManager.cs
Dustylemny: or are you suggesting the RealID returns straight up bytes?

Online/Matchmaking/Packets/RequestJoinPacket.cs
```diff
@@ -20,10 +22,14 @@ public override void Process()
             {
                 var matchmaker = (LANMatchmakingManager)MatchmakingManager.instances[MatchmakingManager.MatchMakingDomain.LAN];
 
-                if (LanUserName.Length > 0) {
+                if (LanUserName.Length > 0) 
+                {
                     processingPlayer.id.name = LanUserName;
                 }
-
+                if (Index > -1)
```
invalidunits, line 29: If the packet has an invalid index, then the packet entirely is invalid. packets shouldn't be able to opt out of this

Online/Matchmaking/Packets/RequestJoinPacket.cs
```diff
@@ -20,10 +22,14 @@ public override void Process()
...
+                {
+                    (processingPlayer.id as LANMatchmakingManager.LANPlayerId)!.index = Index;
```
invalidunits, line 31: Any index conflicts should be resolved here. loop through all of the players and make sure nobody has the same index.

Online/Matchmaking/LANMatchmakingManager.cs
Dustylemny: LANMatchmakingManager ctor only loads once in meadow, so it's fine for now

Online/Matchmaking/Packets/RequestJoinPacket.cs
Dustylemny: so what do you suggest?

Online/Matchmaking/Packets/RequestJoinPacket.cs
Dustylemny: @invalidunits will this update with mePlayer?

Online/Matchmaking/Packets/RequestJoinPacket.cs
Dustylemny: used an error debug instead

PR #834, opened 2025-04-22
playtest fixes

Lizzie841:
[not draft]

### DIFF
--- a/Game/RainMeadow.PlayerHooks.cs
+++ b/Game/RainMeadow.PlayerHooks.cs
@@ -230,9 +230,8 @@ public partial class RainMeadow
         {
             if (self.slugcat.isNPC) return;
 
-            if (self.slugcat.input[0].jmp) self.owner.slugOnBack.DropSlug();
-
-            self.slugcat.standing = true; // SlugNPCs do this in there AI. but it looks right for all players.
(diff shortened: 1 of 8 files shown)


Dustylemny:
@Lizzie841 i have a suggestion for scrolling chat log. having a certain keybind may not be ideal as currently it does not support controllers. and plus using remix options may be too tedious. Maybe a slider or buttons to go up instead

Game/RainMeadow.PlayerHooks.cs
```diff
@@ -230,9 +230,8 @@ private void SlugOnBack_Update(On.Player.SlugOnBack.orig_Update orig, Player.Slu
         {
             if (self.slugcat.isNPC) return;
 
-            if (self.slugcat.input[0].jmp) self.owner.slugOnBack.DropSlug();
-
-            self.slugcat.standing = true; // SlugNPCs do this in there AI. but it looks right for all players.
+            if (self.slugcat.input[0].jmp) self.owner.slugOnBack.DropSlug(); //NOTE: makes self.slugcat null!
```
6fears7, line 233: @invalidunits just want to point you to this bug we found and ensure you are cool with this. It looks like it's OR'ing instead of AND'ing.

invalidunits: Fixed this exact issue in my backpacking PR. Looks good.

6fears7 (approved):

PR #837, opened 2025-04-23
tummy dupe potential fix

Lizzie841:


Story/StoryOnlineMenu.cs
```diff
@@ -81,10 +81,11 @@ public SlugcatStats.Name CurrentSlugcat
 
         public StoryOnlineMenu(ProcessManager manager) : base(manager)
         {
+            manager.rainWorld.progression.ClearOutSaveStateFromMemory(); // please
```
6fears7, line 84: Just for standardization with vanilla story can we put it in StartGame()?

6fears7: Right before:

`            manager.RequestMainProcessSwitch(ProcessManager.ProcessID.Game);`

6fears7 (approved):

PR #850, opened 2025-04-28
batfly no dupe experiment

Lizzie841:


6fears7:
How did this go?

Lizzie841:
> How did this go?

needs further testing, issue is the mass batfly spawn only happens really with more than 2 players which makes it hard to repro alone

Lizzie841:
Does not work

PR #797, opened 2025-04-06
watcher fixes (first echo, crash with hud)

Lizzie841:
- HUD for players does not display for now, unfortunately
- warp is global and takes EVERYONE with them (only host should initiate through)
- 4 letter region names (hard coded for now)

todo?
[ ] prolly needs to fix the new karma being weird
[ ] fix memory out of - stream thing (big rooms? self reference?)
[ ] make warps less wonky (work as if they were gates/insta-spawn everyone at warp point)
[ ] fix the flickering lights on warp point (self made war points are invisible for some reason)
[ ] fix outer rim?

PR #857, opened 2025-04-30
6fears7 arena warcher

6fears7:
Depends on #797


6fears7:
- [ ] Test decreasing rippleLevel to 0.5f and see if the awkward invix vfx box around slugcat disappears so we just update the body / feet sprite to have 0f alpha.

PR #858, opened 2025-04-30
Engine: don't update weapon state if transferring

6fears7:
Problem: Weapon states are being updated while new owners are in the middle of using the weapon. 

Example: Spears thrown in Arena by Player 2 that were owned by Player 1 receive a "Free" Weapon.Mode in the middle of the player 2 throwing the Spear, denying a kill and updating the Mode erroneously. 

Solution: Don't update State if we're transferring.

### DIFF
--- a/Online/State/RealizedSpearState.cs
+++ b/Online/State/RealizedSpearState.cs
@@ -1,4 +1,5 @@
 using RWCustom;
+using Steamworks;
 using System;
 using UnityEngine;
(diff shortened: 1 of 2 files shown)


6fears7:
- [ ] Need to address why spears in walls aren't climbable when this code is set, didn't see anything in debug

henpemaz (approved):

PR #859, opened 2025-05-01
persistent password

Lizzie841:
> it was just so that you could essentially have a default password and if you wanted a private lobby just flip the checkbox and your normal password would already be there


T1mbits (approved):
LGTM, just curious if you would also want to save the user's chosen password on leaving the create menu as a suggestion

Lizzie841:
> LGTM, just curious if you would also want to save the user's chosen password on leaving the create menu as a suggestion

seems like that would be a bit counterintuitive, why would you put a password and have it be saved but not create a lobby?

T1mbits:
I just thought of it as a form field of sorts, you might fill out a form and then leave for whatever reason, you wouldn’t expect the form to get just wiped. just a suggestion though, feel free to ignore it

PR #860, opened 2025-05-01
Weapons fix

invalidunits:
Should fix monks spear throw


Online/Entity/OnlinePhysicalObject.cs
```diff
@@ -491,16 +492,101 @@ public override string ToString()
...
+                    RainMeadow.Error("Invalid collision result");
+                    return;
+                }
+
+                if (collision_obj.apo.realizedObject == null) {
+                    RainMeadow.Error("Object not realized");
+                    return;
+                }
+
+                result = new SharedPhysics.CollisionResult(collision_obj.apo.realizedObject,
```
6fears7, line 537: I mentioned this in DMs to you but the invoked RPC here may null reff on the onAppendagePos, possibly because the spear is lodged in a non-appendage. May be worth implementing a ternary for that.

Online/State/RealizedSpearState.cs
```diff
@@ -78,7 +78,8 @@ override public bool ShouldPosBeLenient(PhysicalObject po)
         {
```
6fears7, line 78: #858 Not fully related but found this to assist in stopping "tossed spears" since previous owners were sending last state weapon mode of "Free" while new owner was setting it to "Thrown". However this seems to stop users from being able to grab spears in walls. Need to test vert spears but found that weird.

Probably has no impact to this PR and will test it again after merge but just sharing other WIP changes to spear.

6fears7 (commented):
Based on our findings last night, it definitely seems like we need Spear_HitSomething to notify close-range targets. This might be because GraspRef is interferring our damage calculations by running before/middle of Violence and HitByWeapon. Part of me thinks if we defer the MakeStick we can handle violence first then manage the stick request afterwards. Needs testing (as always ;__;)

6fears7:
Some positive updates:

Got it to work, here's the sauce:

1. Setting the RunDeferred on MakeStick does help clean up the violence actions at least in logs. It's not required for Monk's damage to register properly but I think it makes sense to keep it deferred.
2. When I use the Spear_HitSomething RPC, the spear's damageBonus is set to 1. This will immediately kill.
3. Tried setting the thrownBy and slugcatstats throwing skill in the reciever's RPC to the creature and 0 respectively to trigger spearBonusDamage = 0.
4. That didn't work.
5. ThrownSpear is where that action happens.
6. So I RPC'd ThrownSpear
7. The magic happened.



Bare minimum here's the data we need:
1. RPC Spear_HitSomething
2. RPC Player_ThrownSpear
3. in PlayerStateState, sync playerState.permanentDamageTracking

Note: I tested PhysicalObject_HitByWeapon used /unused, couldn't tell difference for this. Probably safe to keep.


Ideal:
1. See #1-3
2. Defer MakeStick


MONK DAMAGE BABY

6fears7 (approved):

Online/Entity/OnlinePhysicalObject.cs
```diff
@@ -2,6 +2,7 @@
 using System.Globalization;
 using System.Linq;
 using System.Runtime.CompilerServices;
+using Unity.Burst.Intrinsics;
```
henpemaz, line 5: noise

henpemaz (dismissed):
state changes seem ok.

I don't full follow the changes to online villain but if it works it works

6fears7:
I'll resolve conflits

Online/State/RealizedSpearState.cs
```diff
@@ -23,6 +23,9 @@ public class RealizedSpearState : RealizedWeaponState
         [OnlineFieldHalf(group = "spear")]
         private float spearDamageBonus;
 
+        [OnlineFieldHalf]
```
henpemaz, line 26: half a bool?

henpemaz:
don't forget to test before merging

Online/State/RealizedSpearState.cs
6fears7: ah I was looking for why rain meadow was failing to start when testing and that's the line thank you

henpemaz (approved):

PR #846, opened 2025-04-26
Arena Menu Overhaul

T1mbits:
Arena Lobby UI Overhaul since the current UI is getting quite messy and isn't well suited for the requirements of online multiplayer. The new UI will be based on A1ex's concept art.

This PR has also absorbed many other, occasionally non-arena related, features. Below is a list of all features/major changes included in this PR.
- New Arena Menu
- Separation of online and local arena settings
- Moving most arena remix settings into the new menu
- ButtonScroller reworks
- SmartMenu tweaks
- Random Slugcat option
- Reworked ready system
- Allow players to join an active arena game at the start of a round
- Added additional serialization types to handle null type

<details>
<summary>Screenshots</summary>

image
image
image
image
image
</details>

<details>
<summary>TODO</summary>

- [x] Replicate basic arena setting elements
- [x] Tab system for MenuObjects
    - [x] Filled in Arena Playlist tab
    - [x] Filled in Match Settings tab
    - [x] Filled in Slugcat Abilities tab
- [x] Arena Playlist element (must be recreated from scratch)
- [x] Chat element
- [x] Player display element
- [x] Scug select
- Sync lobby data between clients
    - [x] Scug select state
    - [x] Chat
    - [x] Settings
    - [x] Playlist
- [x] Sync UI elements to arena/remix settings
- Online Game sync
    - [x] Start online game
    - [x] handle joining mid game
- [x] move arena remix settings to lobby menu
- [x] #853 
</details>
<details>
<summary>Concept Art</summary>

image
image

</details>

PR #863, opened 2025-05-03
Add Random Selection to Arena

invalidunits:


Arena/ArenaHelpers.cs
```diff
@@ -76,6 +87,9 @@ public static void RecreateSlugcatCache()
                     }
                 }
             }
+
+            selectableSlugcats.AddRange(allSlugcats);
+            selectableSlugcats.Add(null);
```
6fears7, line 92: 👀 what

6fears7 (commented):
I would like @T1mbits and @Dustylemny to weigh in on these changes. 

Couple things: 

1. I find it a little peculiar to set another state with  randomPlayinAs. I was actually anticipating a random slug option to just iterate through the list and pick a random index and set the PlayingAs to that. 

Additionally, if the random slug changes per round, in the PlayerResult_ctor you can just find the playerNumber or index, get the backing online player, then set player.playerClass to that new value so by the time stats are shown we've covered the image loading. I'm going that route for managing rejoins if a player leaves to change classes and rejoins next level load. 

Other than that, really bizarre that null is being added to the list and I know that that's what the original developers did but also just weirds me out versus us just using Survivor. Can you provide more background on that?

2. Ok I only had 1 thing I guess :p

Arena/ArenaHelpers.cs
invalidunits: Rain world vanilla uses null as the Random selected Slugcat. This could be changed to a dummy slugcat but null works well enough.

T1mbits: adding an extra dummy slugcat extenum would also allow a proper "random" slugcat option in the new slugcat select menu without much extra work

Dustylemny: might as well just use -1 for random slugcat dont yah think?

6fears7:
Team has requested this PR remain open and unmerged until after 1.5.0 menu is complete so that we can pretty much cherry pick it into that.

invalidunits:
Drafting this until the Arena team's menu rework is finished.

invalidunits:
Closing in favor of #846

PR #868, opened 2025-05-08
meadow main menu button did not play sound

T1mbits:
#852 except now it's not on my fork's main branch so it doesn't get deleted if I try to sync my fork


invalidunits:
The git diff is weird, but there is only one line change.

T1mbits:
yeah I'm not sure why it looks like that, I think it's something to do with my formatter

invalidunits:
Since it's one line do you mind copying and pasting it in GitHub web editor. It's not that big of deal so I'll merge it regardless.

T1mbits:
the change is from differing line endings I think on further research, but yeah I'll create a new PR

PR #869, opened 2025-05-08
final boss beaten

6fears7:


6fears7:
@invalidunits  this has been tested to not crash, just raising awareness for you. We were previously only overwriting on save, which creates issues for incoming players.

PR #873, opened 2025-05-10
watcher abilities on story

Lizzie841:
depends on #797 ; made separate pr because i do not understand if this is OK or Rejected


6fears7:
Looks fine. Need to address two players playing as watcher entering ripple space still I think

6fears7:
Why not just set ripple to karma levels in non watcher campaigns?

Lizzie841:
> Why not just set ripple to karma levels in non watcher campaigns?

arti campaigns

6fears7:
Waiting on watcher main PR before taking a look at this

PR #876, opened 2025-05-11
Improve cape water and gravity Interactions

invalidunits:


6fears7 (commented):
Just a General reminder going through warps in watcher null reff the cape. Not game breaking but throws errors every frame.

6fears7 (approved):

6fears7:
Merge at discretion

PR #877, opened 2025-05-11
Shortcut fixes & Backpack specific abilities.

invalidunits:
Started as a fix for some of the reversion i did with #808  Added specific abilities for saint and Gourmand when on the back of player slugcats.

### DIFF
--- a/Online/Entity/OnlinePhysicalObject.cs
+++ b/Online/Entity/OnlinePhysicalObject.cs
@@ -269,6 +269,7 @@ namespace RainMeadow
                     RainMeadow.Debug($"topos Node defined? {topos.NodeDefined}");
 
                     newRoom.absroom.AddEntity(apo);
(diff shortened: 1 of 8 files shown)


Online/Entity/OnlinePhysicalObject.cs
```diff
@@ -252,7 +252,7 @@ protected override void JoinImpl(OnlineResource inResource, EntityState initialS
             RainMeadow.Debug($"{this} joining {inResource} at {poState.pos}");
             try
             {
-                AllMoving(true);
+                beingMoved = true;
```
6fears7, line 255: Note to self to test arena MULTI_ROOM maps to ensure no weirdness came in

henpemaz (approved):
A bit skeptical about the changes to OPO but happy to try them out

PR #881, opened 2025-05-14
patch for 1.10.4

Lizzie841:


Game/RainMeadow.LoadingHooks.cs
```diff
@@ -10,8 +10,15 @@ public partial class RainMeadow
         private void LoadingHooks()
         {
             On.WorldLoader.ctor_RainWorldGame_Name_Timeline_bool_string_Region_SetupValues += WorldLoader_ctor;
-
             On.WorldLoader.Update += WorldLoader_Update;
+            On.WorldLoader.UpdateThread += (On.WorldLoader.orig_UpdateThread orig, WorldLoader self) => {
```
invalidunits, line 14: Don't use lambda hooks. Create a new function instead.

Game/OnlineController.cs
```diff
@@ -13,6 +13,8 @@ public OnlineController(OnlineEntity ent, Player self)
 
         public override Player.InputPackage GetInput()
         {
+            if (ent == null) return new Player.InputPackage();
```
invalidunits, line 16: `OnlineController.ent` being null is a very bad sign, Having it break here might be better.

Online/Resource/WorldSession.Entities.cs
```diff
@@ -5,7 +5,7 @@ namespace RainMeadow
 {
     public partial class WorldSession
     {
-        private List<AbstractPhysicalObject> earlyApos = new(); // stuff that gets added during world loading
+        private List<AbstractPhysicalObject> earlyApos = []; // stuff that gets added during world loading
```
invalidunits, line 8: Why?

invalidunits (changes requested):
Looks good!

Lizzie841:
Would this do?

Game/OnlineController.cs
Lizzie841: leftover

6fears7:
This has been tested successfully, I notified @henpemaz of this PR for review as well.

henpemaz:
rewrote some of this in !884 this can probs be closed

PR #882, opened 2025-05-14
water splash reduction experiment

Lizzie841:


Lizzie841:
only mititages issue (drops to 10 fps instead of 1 fps) but does not fix it completely

Game/RainMeadow.GameHooks.cs
```diff
@@ -54,6 +55,28 @@ private void GameHooks()
...
+                var skip = il.DefineLabel();
+                int recursion = 0;
+                c.GotoNext(moveType: MoveType.After,
+                    i => i.MatchLdarg(out recursion),
+                    i => i.MatchLdcI4(10),
+                    i => i.MatchBle(out skip)
+                );
+                c.MoveAfterLabels();
+                c.Emit(OpCodes.Ldarg, recursion);
+                c.EmitDelegate((int value) => OnlineManager.lobby == null || value < 5);
```
youbitchoc, line 71: if this only happens for remote creatures then consider changing the online check to !IsLocal()

also personally I'd negate the conditions to `!creature.IsLocal() && value >= 5` then Brfalse

Lizzie841: It also happens for your owned creatures

youbitchoc: but only in meadow?

PR #900, opened 2025-05-24
null ext enum support

Lizzie841:


invalidunits:
#863 Implements this, It's likely to be integrated with the Arena Overhaul #846.

PR #905, opened 2025-05-28
Fix Hit ThrownBy

6fears7:
maybe some solutions are simple

### DIFF
--- a/Online/Entity/OnlinePhysicalObject.cs
+++ b/Online/Entity/OnlinePhysicalObject.cs
@@ -560,7 +560,7 @@ namespace RainMeadow
                 {
                     if (result.Value.obj != null && result.Value.obj is Creature c)
                     {
-                        c.killTag = (this.apo.realizedObject as Weapon).thrownBy.abstractCreature;
+                        c.SetKillTag((this.apo.realizedObject as Weapon).thrownBy.abstractCreature);
                     }
                     (this.apo.realizedObject as Weapon)!.HitSomething(result.Value, true);
                 }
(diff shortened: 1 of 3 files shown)


6fears7:
https://discord.com/channels/1094716194180841602/1377457464949346465/1377482198474821712 Still need to address why it sometimes still does not want to capture a proper PvP kill feed

Online/Entity/OnlinePhysicalObject.cs
```diff
@@ -556,10 +556,15 @@ public void WeaponHitSomething(RealizedWeaponState statewhenhit, OnlineCollision
                 statewhenhit.ReadTo(this);
                 SharedPhysics.CollisionResult? result = null;
                 hit.BuildCollisionResult(out result);
-                if (result.HasValue) {
+                if (result.HasValue)
+                {
+                    if (result.Value.obj != null && result.Value.obj is Creature c)
+                    {
+                        c.killTag = (this.apo.realizedObject as Weapon).thrownBy.abstractCreature;
```
invalidunits, line 563: Could the kill tag be timing out?
```suggestion
                        c.SetKillTag((this.apo.realizedObject as Weapon).thrownBy.abstractCreature);
```

invalidunits (dismissed):
No blockers

Online/Entity/OnlinePhysicalObject.cs
6fears7: need to test this. I think it's from the fact that each time it failed to register it also notified in weapon state that it would not update state because it was transferring to new owner so I need to wrap that transfer around the weapon mode instead of the whole state because we are ALSO using thrownBy in the weapon state so it seems like it should trigger the killtag if that's set correctly versus being in the RPC.


But if the state has a risk of transferring then it seems like the thrownBy in the state is only useful part of the time.

6fears7:
Need to revisit this

6fears7:
Newest changes with SetKillTag still only work on 2 kills then stop working. Very bizarre.

PR #910, opened 2025-05-30
No-op on negative indices when spawning at shortcuts

Lizzie841:
draft for now, maybe this isn't a good solution - maybe it is


invalidunits:
Duplicate of #877 which should fix this issue when merged.

PR #911, opened 2025-05-30
fix sanitize on null story client data

ghost:
found while playtesting, draft because read the comment:

>             // NOTE: May start game before client data is fully created, how can we fix? - does it need to be fix?


invalidunits:
I've never seen this happen, storyClientData is initialized as soon as the lobby becomes active, so it wouldn't make sense for it to be null before you join the lobby.

Closing for now.

PR #918, opened 2025-06-06
fix incorrect scug body weight

yuzugamer:
moves slugcatStatsPerPlayer stuff to before body weight is set, so that slugcatStats returns the correct stats


6fears7 (approved):

invalidunits (approved):
Looks good!
Make sure you test this with slugNPCs before merging.

yuzugamer:
not sure if you were asking me or not, but i went ahead and tested npcs and it seemed fine

invalidunits:
Nice work!

PR #927, opened 2025-06-18
patch spears clipping in edge cases

yuzugamer:
note: this is a vanilla issue, which i planned on creating a separate mod to fix, but considering it's by far most noticeable in rain meadow (and meadow will probably be blamed for it), i thought it would fit here
patches the following edge cases (specifically for slugcat on slugcat violence, so as to minimize invasiveness), which would cause spears to inexplicably miss/'not register':
weapons flying in between body chunk hitboxes, and subsequently missing (by snapping to the nearest body chunk if it goes in between)
weapons missing from being thrown too close to a small body chunk, and starting on the other side (by making weapons start farther back)

i haven't done super thorough testing, but both patches function properly and the game doesn't immediately break

### DIFF
--- a/Game/RainMeadow.GameplayHooks.cs
+++ b/Game/RainMeadow.GameplayHooks.cs
@@ -6,6 +6,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using UnityEngine;
+using RWCustom;
 
 namespace RainMeadow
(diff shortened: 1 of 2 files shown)


yuzugamer:
accidentally added a separate commit somehow, whoops

PR #928, opened 2025-06-18
patch spears clipping in edge cases

yuzugamer:
note: this is a vanilla issue, which i planned on creating a separate mod to fix, but considering it's by far most noticeable in rain meadow (and meadow will probably be blamed for it), i thought it would fit here

patches the following edge cases (specifically for slugcat on slugcat violence, so as to minimize invasiveness), which would cause spears to inexplicably miss/'not register':
- weapons flying in between body chunk hitboxes, and subsequently missing (by snapping to the nearest body chunk if it goes in between)
- weapons missing from being thrown too close to a small body chunk, and starting on the other side (by making weapons start farther back)

i haven't done super thorough testing, but both patches function properly and the game doesn't immediately break
also it might be a little bit better if the SharedPhysics.TraceProjectileAgainstBodyChunks is an il hook, but i spent about two hours fumbling around with it and inexplicably being unable to

### DIFF
--- a/Game/RainMeadow.GameplayHooks.cs
+++ b/Game/RainMeadow.GameplayHooks.cs
@@ -92,8 +92,12 @@ namespace RainMeadow
         // meant to fix spears going in between body chunks in arena mode, but could also be useful for other potential pvp-focused gamemodes
         public static SharedPhysics.CollisionResult SharedPhysics_TraceProjectileAgainstBodyChunks(On.SharedPhysics.orig_TraceProjectileAgainstBodyChunks orig, SharedPhysics.IProjectileTracer projTracer, Room room, Vector2 lastPos, ref Vector2 pos, float rad, int collisionLayer, PhysicalObject exemptObject, bool hitAppendages)
         {
+            if (OnlineManager.lobby == null || !(RainMeadow.isArenaMode(out var arena) ? arena.weaponCollisionFix : RainMeadow.rainMeadowOptions.WeaponCollisionFix.Value))
+            {
+                return orig(projTracer, room, lastPos, ref pos, rad, collisionLayer, exemptObject, hitAppendages);
+            }
             var result = orig(projTracer, room, lastPos, ref pos, rad, collisionLayer, exemptObject, hitAppendages);
-            if (OnlineManager.lobby != null && result.chunk == null && result.onAppendagePos == null && exemptObject is Player)
+            if (result.chunk == null && result.onAppendagePos == null && exemptObject is Player)
             {
                 // only need to check the collision layer that slugcats occupy
                 foreach (var obj in room.physicalObjects[1])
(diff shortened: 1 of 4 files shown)


Game/RainMeadow.GameplayHooks.cs
```diff
@@ -86,6 +89,49 @@ private void SocialEventRecognizer_CreaturePutItemOnGround(On.SocialEventRecogni
             }
         }
 
+        // meant to fix spears going in between body chunks in arena mode, but could also be useful for other potential pvp-focused gamemodes
+        public static SharedPhysics.CollisionResult SharedPhysics_TraceProjectileAgainstBodyChunks(On.SharedPhysics.orig_TraceProjectileAgainstBodyChunks orig, SharedPhysics.IProjectileTracer projTracer, Room room, Vector2 lastPos, ref Vector2 pos, float rad, int collisionLayer, PhysicalObject exemptObject, bool hitAppendages)
+        {
+            var result = orig(projTracer, room, lastPos, ref pos, rad, collisionLayer, exemptObject, hitAppendages);
```
6fears7, line 99: Would like for this to be an early return:

if (OnlineManager.lobby == null) {

return orig(projTracer, room, lastPos, ref pos, rad, collisionLayer, exemptObject, hitAppendages);

}

6fears7:
passing to @invalidunits  for any additional thoughts.

6fears7 (dismissed):

invalidunits (changes requested):
Looks good!

This needs be a remix setting for Players and maybe Creatures. The hosts choice should sync in the Lobby data of all game modes.

It'll should be ready to merge after that.

invalidunits:
>  I cannot think through scenarios where a spear passes through that a user presumes "I expect this to happen"
I can imagine some people liking the vanilla physics. People should be able to opt out since this change is a physics change and unrelated to RainMeadow.

yuzugamer:
i'm not super sure how i feel about making it work for creatures, since i mainly intended it as a patch for pvp from a competitive standpoint, because arena has undoubtedly started to become a competitive scene to some degree. i still could add a remix setting for it to work on creatures, just not sure it makes much sense

turning the changes into a remix setting is pretty reasonable though, i'll definitely do that. should i add it to the remix menu, or just make the configurable field and let someone else handle the menu stuff?

6fears7:
@yuzugamer make it a remix option. We will integrate it into the new arena UI soon enough. Thank you

yuzugamer:
i made it into a remix setting. i wasn't really sure how to handle syncing, since this is kind of a special case and normally the only remix settings that are always synced are vanilla ones, so i just threw it into ArenaLobbyData, since that's where it'll normally be relevant

invalidunits (approved):

PR #933, opened 2025-06-23
add '.' separator for rooms/regions

ghost:
draft since I'm not sure if this is a good idea or not

### DIFF
--- a/Online/OnlineManager.cs
+++ b/Online/OnlineManager.cs
@@ -2,6 +2,7 @@
 using Steamworks;
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


6fears7:
The only icky part is a gate room is "GATE_XX_ROOMNUMBER

Not sure if that will impact this change

Lizzie841:
> The only icky part is a gate room is "GATE_XX_ROOMNUMBER
> 
> Not sure if that will impact this change

I have an idea

invalidunits (approved):
The comment on line 337 is outdated.

PR #922, opened 2025-06-16
Custom Packet

Quaxly:
The CustomPacket allows for third party mods to utilize lower level networking. Mods can send and receive custom data using a key. Mods can subscribe to CustomManager to receive and process any CustomPacket with a matching key.

PR #935, opened 2025-06-25
Data chunk framework Rework

invalidunits:
Possibly Supersedes #452 and #922,


Online/Resource/Lobby.cs
```diff
@@ -305,7 +303,15 @@ protected override void NewParticipantImpl(OnlinePlayer player)
                 RainMeadow.Debug($"Assigned inLobbyId of {nextId} to player {player}");
                 nextId++;
                 // todo overflows and repeats (unrealistic but it's a ushort)
+
```
6fears7, line 306: still thinking of holding on this commented data?

invalidunits: I removed it.

Online/Chunks/DataChunk.cs
```diff
@@ -0,0 +1,340 @@
...
+            this.destination = destination;
+            this.toPlayer = toPlayer;
+            this.chunkId = chunkId;
+            if (reliable)
+            {
+                checked
+                {
+                    // todo gzip data and keep track of gzip status
+                    int sliceCount = (data.Count + sliceSize - 1) / sliceSize;
+                    if (sliceCount > (int)byte.MaxValue) throw new InvalidProgrammerException("Too much data! Try increasing the slice count");
```
henpemaz, line 76: Shouldn't it  be the other way around?

henpemaz:
in the context of a voice-chat implementation I still think #922 is more appropriate

shoehorning reliable and unreliable into a large-chunk-sync system seems off

the chunk system is for large, reliable data (think one-time-sync of big tables at the start of the session)
the rpc system is for small, reliable data (rpcs!)
the state system is for small(?), unreliable data (game state, for which eventual-consistency is fine)

yes there's no coverage for large unreliable, but is that relevant for VC? to my eyes VC is small-unreliable
we do online ticks at 20 tps, that's 50ms of voice, which should be about 2kb per tick at 80kbps audio (voip standard)

I think if we were to slice and send voip data over several ticks that'd already be counterproductive latency-wise, it probably doesn't benefit from being buffered and sliced even more than it already is

I think we could support more destinations for State messages in a way to let mods send custom data through there, but #922 is just as good in enabling that

invalidunits:
https://github.com/henpemaz/Rain-Meadow/pull/954#issuecomment-3086423191
Drafting this for now. I'll work with @Quaxly to get a solution for VC, Possibly including #922.

PR #936, opened 2025-06-25
Arena: Add ACs to room Creature list

6fears7:
Fixes issues with rot, worms, etc.. not seeing an AC to track


invalidunits:
I'm pretty sure this is entirely fixed by adding `abstractCreature.Room.AddEntity(abstractCreature);` to line 173 of `BaseGameMode.cs` in `SpawnPlayers`.

PR #939, opened 2025-06-29
assorted (possible) fixes, and changes, in UDP networking

niacdoial:
I feel like there's no way I all of my findings are legit.
anyway here are my possible bugfixes and changes.

- dealt with possible buffer overruns in UdpPeerManager's SendRaw and Receive methods
- merged Heartbeat/Acknowledgement packets (reduce the likelyhood that an Reliable packets need to be resent)
- prevent the exponential growth of Heartbeat packets going around (all Heartbeats required a Heartbeat response until now)
- ~~changed how ModifyPlayerListPacket detects "the current player" so that it doesn't malfunction when >2 people share a machine (...which likely only happens when debugging, unless multi-seat computers are involved)~~ nope, turns out I misunderstood what `LANPlayerID.isLoopback()` means.
- make sure the begin_conversation flag persists when a Reliable packet fails its first delivery
- changed how the counters are used in Acknowledgement logic (now, if a reliable packet is skipped (somefluffinghow), this would get logged!)
- patched isEndpointLocal for correctness
- now saving 2 bytes off of every endPoint sent over the network

### DIFF
--- a/Online/Matchmaking/Packets/ModifyPlayerListPacket.cs
+++ b/Online/Matchmaking/Packets/ModifyPlayerListPacket.cs
@@ -33,8 +33,8 @@ namespace RainMeadow
             lanids = lanids.Where(x => x.endPoint != null);
 
             bool includeme = lanids.FirstOrDefault(x => x.isLoopback()) is not null;
-
-            lanids = lanids.Where(x => !x.isLoopback());
+            if (includeme)
+                lanids = lanids.Where(x => !x.isLoopback());
             var processinglanid = (LANMatchmakingManager.LANPlayerId)processingPlayer.id;
             UDPPeerManager.SerializeEndPoints(writer, lanids.Select(x => x.endPoint).ToArray(), processinglanid.endPoint, includeme);
 
@@ -62,7 +62,7 @@ namespace RainMeadow
                     players[i].id.name = reader.ReadNullTerminatedString();
                 }
             }
-                
+
             else if (modifyOperation == Operation.Remove)
                 players = endpoints.Select(x => lanmatchmaker.GetPlayerLAN(x)).ToArray();
 
@@ -79,7 +79,7 @@ namespace RainMeadow
                     {
                         if (((LANMatchmakingManager.LANPlayerId)players[i].id).isLoopback()) {
                             // That's me
-                            // Put me where I belong.
+                            // Put me where I belong. (...at the end of the player list?)
                             OnlineManager.players.Remove(OnlineManager.mePlayer);
                             OnlineManager.players.Add(OnlineManager.mePlayer);
                             continue;
--- a/Online/Matchmaking/LANMatchmakingManager.cs
+++ b/Online/Matchmaking/LANMatchmakingManager.cs
@@ -12,12 +12,12 @@ using System.Diagnostics.PerformanceData;
 
 
 namespace RainMeadow {
-    
+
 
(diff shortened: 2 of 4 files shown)


niacdoial:
(of course, feel free to tell me to back some of those changes off, or to change the target branch, ...or anything really, heh)

Online/Matchmaking/Packets/ModifyPlayerListPacket.cs
```diff
@@ -32,9 +32,10 @@ public override void Serialize(BinaryWriter writer)
             var lanids = players.Select(x => (LANMatchmakingManager.LANPlayerId)x.id);
             lanids = lanids.Where(x => x.endPoint != null);
 
-            bool includeme = lanids.FirstOrDefault(x => x.isLoopback()) is not null;
 
-            lanids = lanids.Where(x => !x.isLoopback());
+            bool includeme = lanids.FirstOrDefault(x => x == (LANMatchmakingManager.LANPlayerId)OnlineManager.mePlayer.id) is not null;
+            if (includeme)
+                lanids = lanids.Where(x => x != (LANMatchmakingManager.LANPlayerId)OnlineManager.mePlayer.id);
```
invalidunits: The `IsLoopback` in the original is fine. The comparison does the same thing.

Online/Matchmaking/Packets/ModifyPlayerListPacket.cs
```diff
@@ -77,9 +78,9 @@ public override void Process()
...
                     for (int i = 0; i < players.Length; i++)
                     {
-                        if (((LANMatchmakingManager.LANPlayerId)players[i].id).isLoopback()) {
+                        if (((LANMatchmakingManager.LANPlayerId)players[i].id) == OnlineManager.mePlayer.id) {
                             // That's me
-                            // Put me where I belong.
+                            // Put me where I belong. (...at the end of the player list?)
                             OnlineManager.players.Remove(OnlineManager.mePlayer);
                             OnlineManager.players.Add(OnlineManager.mePlayer);
                             continue;
```
invalidunits: See review Above.

For your comment, we need a matching player list for each player to resolve who owns the lobby when the host leaves. We're re-adding ourselves to the correct position, which wont necessarily be at the end of the **completed** list.

invalidunits (commented):
This looks nice, I'd rather keep refactoring to major features or fixes. I'll draft this for the Router PR.

Online/Matchmaking/Packets/ModifyPlayerListPacket.cs
niacdoial: ok, I'll revert that

Online/Matchmaking/Packets/ModifyPlayerListPacket.cs
niacdoial: Ok, thank you!

invalidunits:
> fair enough! Do you want me to make another version of this PR that only includes fixes that don't refactor how the network is used?

I wouldn't mind that, Thanks a lot.

niacdoial:
reverted what I said I would, and rebased this PR on top of the new one

6fears7:
@niacdoial is this still relevant?

niacdoial:
@6fears7 I don't think so.
basically all low-level networking improvements are rolled up in the big RouterNetworkDomain branch

PR #943, opened 2025-07-03
UdpPeer.cs: fix a couple of bugs

niacdoial:
"redux" version of #939, without changes in how the packets themselves are structured.

bugs fixed:
- a couple of possible buffer buffer overruns in Receive and SendRaw
- the begin_conversation flag being dropped when a packet needs to be resent
- miscount in the number of heartbeats (or queued Reliable packets) to send


niacdoial:
...wait I haven't tested this, give me five minutes

niacdoial:
done. There was an embarrassing mistake that prevented compilation altogether, but that seems like that's all from my vague testing.

invalidunits (approved):
Looks Good!

PR #952, opened 2025-07-07
Story: dont end red early if player dies

6fears7:


6fears7:
```
        private void RainWorldGame_GoToStarveScreen(On.RainWorldGame.orig_GoToStarveScreen orig, RainWorldGame self)
        {
            if (isStoryMode(out var storyGameMode))
            {
                if (OnlineManager.lobby.isOwner)
                {
                    foreach (var player in OnlineManager.players)
                    {
                        if (!player.isMe) player.InvokeOnceRPC(StoryRPCs.GoToStarveScreen, storyGameMode.myLastDenPos);
                    }
                }
                else if (RPCEvent.currentRPCEvent is null)
                {
                    // tell host to move everyone else
                    OnlineManager.lobby.owner.InvokeOnceRPC(StoryRPCs.GoToStarveScreen, storyGameMode.myLastDenPos);
                    return;
                }
            }

            orig(self);
        }
```

This is also present in starve logic, not sure if we want to keep that.

Story/StoryHooks.cs
```diff
@@ -1366,7 +1366,7 @@ private void RainWorldGame_GoToRedsGameOver(On.RainWorldGame.orig_GoToRedsGameOv
                 else if (RPCEvent.currentRPCEvent is null)
                 {
                     // tell host to move everyone else
-                    OnlineManager.lobby.owner.InvokeOnceRPC(StoryRPCs.GoToRedsGameOver);
+                    //OnlineManager.lobby.owner.InvokeOnceRPC(StoryRPCs.GoToRedsGameOver);
```
invalidunits, line 1369:
```suggestion
                    if (self.StoryCharacter != SlugcatStats.Name.Red)
                    { 
                        OnlineManager.lobby.owner.InvokeOnceRPC(StoryRPCs.GoToRedsGameOver);
                    }
```

invalidunits (changes requested):
This needs testing for gourmand's and artificer's endings. Host will still end the game early if they die.

This isn't a high priority issue though so it will work as a temporary solution. We can apply it for only Hunter's campaign and check later if the other campaigns need this.

Story/StoryHooks.cs
6fears7: I think we can remove the entire client's logic. I'm assuming we only want Host to be responsible for sending everyone to Red's game over:

```
 if (OnlineManager.lobby.isOwner)
                {
                    foreach (var player in OnlineManager.players)
                    {
                        if (!player.isMe) player.InvokeOnceRPC(StoryRPCs.GoToRedsGameOver);
                    }
                }
```

PR #956, opened 2025-07-10
AFK Sleeping Animation

CptnRedstone:
When inputs haven't been touched for at least 30 seconds, and slugcat has been in a valid place to sleep for 5 seconds, they'll enter the shelter sleeping animation until they input something or are pushed into an invalid position. Applies to all online gamemodes.

### DIFF
--- a/Online/State/RealizedPlayerState.cs
+++ b/Online/State/RealizedPlayerState.cs
@@ -110,6 +110,12 @@ namespace RainMeadow
         private bool flipDirection;
         [OnlineField]
         private bool glowing;
+        [OnlineFieldHalf]
+        public float sleepCurlUp;
+        [OnlineField]
+        public int touchedNoInputCounter;
+        [OnlineField]
+        public int ableToSleep;
         [OnlineField(nullable = true)]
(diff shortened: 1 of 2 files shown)


Online/State/RealizedPlayerState.cs
```diff
@@ -110,6 +110,12 @@ public class RealizedPlayerState : RealizedCreatureState
         private bool flipDirection;
         [OnlineField]
         private bool glowing;
+        [OnlineFieldHalf]
+        public float sleepCurlUp;
+        [OnlineField]
+        public int touchedNoInputCounter;
+        [OnlineField]
+        public int ableToSleep;
```
henpemaz, line 118: this might sound crazy but I find it hard to justify 12 bytes per player packet for afk sleep, let me see what I can do

CptnRedstone:
It's possible to remove ableToSleep entirely by basically capping touchedNoInputCounter if the player isn't in a valid place to sleep. It'd save some packet size, but, seems hacky which is why I avoided it. I'm not aware of any problems it'd cause though; would that be preferable?

PR #961, opened 2025-07-11
Hashed Capes & Downloading Capes

invalidunits:


Game/CustomCosmetics/SlugcatCape.cs
```diff
@@ -72,31 +147,33 @@ static public void FetchCapes()
...
+                if (entries.TryGetValue(hashed_cape_str, out Color found_entry))
                 {
-                    return entry.color;
+                    cape_cache.Add(player, found_entry);
+                    return found_entry;
                 }
             }
+
+            if (player is LANMatchmakingManager.LANPlayerId lanPlayer)
+            {
```
6fears7, line 170: you wanna keep these for testing?

henpemaz (approved):

Game/CustomCosmetics/SlugcatCape.cs
invalidunits: Yup.
It also works as a way for players to experience different cape colors.

PR #965, opened 2025-07-14
filters for modded gamemodes

ghost:


6fears7 (approved):
LGTM, assuming you've tested this with menagerie or another mode

Lizzie841:
> LGTM, assuming you've tested this with menagerie or another mode

image

PR #967, opened 2025-07-14
relative to camera pointing

Lizzie841:
should fix SBcamerascroll pointing
should also fix pointing being weird in multirooms


6fears7 (commented):
You must have some hate for Vector2s

6fears7 (approved):

PR #971, opened 2025-07-15
change default opacity of chat to stronger

Lizzie841:
tied to prev pr


6fears7:
I'll test both options later today. I would prefer that we have a very low initial opacity to let the user know it exists but it's nearly transparent.

6fears7:
@Lizzie841 After testing, I would like to keep the opacity level low by default.

PR #977, opened 2025-07-18
Revert "fix pvp message spam on abyss, rework to use ext enum"

6fears7:
Reverts henpemaz/Rain-Meadow#964


6fears7:
going to merge for now. Sorry liz, there's a lot of changes here that impact a lot that require more testing

PR #983, opened 2025-07-19
Team Battle: add count

6fears7:


6fears7:
image

haters will say its fake

6fears7:
@Dustylemny late night and my brain is cooked so lmk if that looks good

PR #992, opened 2025-07-21
Stop PlacedObjects from being loaded before the RoomSession is available.

invalidunits:
This should also alleviate any shortcut issues.


6fears7:
I included this in a few builds, didn't seem to have any issues.

PR #993, opened 2025-07-21
1.6.1 Hotfixes

6fears7:
fix starved clients on host win, potential fix for spear phasing, and fix ff in story

- [x] TODO: Fix spectator jumping back to Player[0] on activate. Only Player[0] is present in Story unless Jolly is active. We never add online players to the story player list.

- [ ] Playtest more aggressively

### DIFF
--- a/Game/RainMeadow.GameplayHooks.cs
+++ b/Game/RainMeadow.GameplayHooks.cs
@@ -167,7 +167,7 @@ namespace RainMeadow
 
                 if (obj is Creature c && WeaponIsDangerous(self))
                 {
-                    return !c.FriendlyFireSafetyCandidate(self.thrownBy);
+                    return orig(self, obj) && !c.FriendlyFireSafetyCandidate(self.thrownBy);
                 }
 
             }


Game/RainMeadow.GameplayHooks.cs
```diff
@@ -155,21 +158,19 @@ bool WeaponIsDangerous(Weapon weapon)
...
+                    return false;
+                }
 
-            if (obj is Creature c && c.FriendlyFireSafetyCandidate(self.thrownBy) && WeaponIsDangerous(self))
-            {
-                return false;
-            }
+                if (obj is Creature c && WeaponIsDangerous(self))
+                {
+                    return !c.FriendlyFireSafetyCandidate(self.thrownBy);
```
6fears7: Story code was falling through to default hook strangely enough, but never worked. If I manually set story.friendlyFire above this line, it worked fine. So I figure we put this at the bottom of the online logic and return its inverted bool.

Game/RainMeadow.GameplayHooks.cs
```diff
@@ -155,21 +158,19 @@ bool WeaponIsDangerous(Weapon weapon)
 
         private bool Weapon_HitThisObject(On.Weapon.orig_HitThisObject orig, Weapon self, PhysicalObject obj)
         {
-            if (!self.IsLocal())
+            if (OnlineManager.lobby != null)
```
6fears7, line 161: I removed self.IsLocal(). I'm thinking that some clients are still negotiating spears in some arena lobbies so even if the spear isn't ours, let's try and make it work.

OnlineUIComponents/SpectatorOverlay.cs
```diff
@@ -142,11 +142,7 @@ public SpectatorOverlay? Overlay
                 {
                     OnlineCreature? spectatee = Overlay?.spectatee?.GetOnlineCreature();
                     var spectatable_avatars = GetSpectableAvatars().ToList();
-                    if (player.isMe)
-                    {
-                        spectatee = null;
```
6fears7, line 147: I removed this behavior. Now with Jolly, you can click yourself if you want to cycle through avatars. Or if youre not using jolly, you'd just activate camera back on you.

OnlineUIComponents/SpectatorOverlay.cs
```diff
@@ -158,15 +154,16 @@ public SpectatorOverlay? Overlay
                         }
                         else
                         {
-                            spectatee = spectatable_avatars.FirstOrDefault();
```
6fears7, line 161: Find us if we can, else find whoever.

OnlineUIComponents/SpectatorHud.cs
```diff
@@ -105,12 +116,6 @@ public override void Update()
                 }
                 spectatorOverlay.forceNonMouseSelectFreeze = hud.parts.Find(x => x is ChatHud) is ChatHud { chatInputActive: true };
                 spectatorOverlay.Update();
-
-                if (spectatorOverlay.spectatee is null && isSpectating)
```
6fears7, line 109: This creates a lot of issues opening the overlay. If I'm in a previous region and I close and reopen spectate, I crash. Now the behavior is just giving control to the OnClick event.

Game/RainMeadow.GameplayHooks.cs
invalidunits: Looks good.

Check the origin to preserve mod compat
```suggestion
                    return orig(self, obj) && !c.FriendlyFireSafetyCandidate(self.thrownBy);
```

invalidunits (approved):
Nothing blocking, LGTM

PR #1002, opened 2025-07-25
Fix Story Custom Colors UI

Dustylemny:
based on
https://discord.com/channels/1094716194180841602/1155268800669831270/1397572683390259305

Thought it would be easier to use SlugcatSelectMenu's jolly-coop handling itself as it does what it should do


invalidunits (changes requested):
The changes are nice but this refactors a lot of the multiple avatar stuff and the jolly Coop menu code, I'd prefer separating the Jolly Co-OP code since It reads better.

PR #1004, opened 2025-07-28
Remix color crash fix and some Jolly Co-OP menu changes

invalidunits:
Could supersede #1002, fixes the same issues without any massive refactoring.


Dustylemny:
Did some testing, and noticed some bugs are still not yet fixed here

image
Selecting ur slugcat on slugcat list will not update the custom color config (leads to that slidersetvalue issue in https://discord.com/channels/1094716194180841602/1155268800669831270/1397572683390259305 when sliders are open) 

this can be seen with/without jolly as long that color interface is accessible

Game/RainMeadow.JollyHooks.cs
```diff
@@ -568,8 +574,27 @@ private void RoomCamera_ChangeCameraToPlayer(On.RoomCamera.orig_ChangeCameraToPl
...
+                return SlugcatColorableButton.GetFileForSlugcat(className, false);
+            }
+
+
+            return orig(self, className, colorIndexFile);
+        }
+
         private void JollyPlayerSelector_Update(On.JollyCoop.JollyMenu.JollyPlayerSelector.orig_Update orig, JollyCoop.JollyMenu.JollyPlayerSelector self)
         {
+            if (self.menu is StoryOnlineMenu SOM)
```
Dustylemny: Was checking on the code changes and noticed this

Sorry for bothering a bit but this will never be true as JollyPlayerSelector's menu usually would be `JollySetupDialog`
If this is unintended you should use `self.menu.processmanager.currentMainLoop` instead

6fears7 (changes requested):
Please review Dusty's comments

invalidunits:
> Found another crash. This time when campaign slugcat is turned on, color config doesnt refresh to campaign which can lead to a crash 

I've been slacking on GitHub for a while and I'm sorry for the lack of response. I've just resolved this, Thanks for the report.

Dustylemny (changes requested):
So far everything works well.
except when you dont have rain world remix
```NullReferenceException: Object reference not set to an instance of an objectRainMeadow.StoryOnlineMenu.Update () (at /Story/StoryOnlineMenu.cs:283)MainLoopProcess.RawUpdate (System.Single dt) (at <0d3d3bd248b74a708fd84538ba1e82f5>:0)ProcessManager.Update (System.Single deltaTime) (at <0d3d3bd248b74a708fd84538ba1e82f5>:0)```

recommend to see if there any more errors without remix :shrug:

**Comment by 6fears7** [comment]
@Dustylemny lmk if this works for you

File: Menu/Components/SlugcatColorableButton.cs [hunk 2]
```diff
@@ -42,7 +42,12 @@ public static string GetFileForSlugcatIndex(SlugcatStats.Name? slugcat, int colo
             if (ModManager.MSC && slugcat == MSCScugs.Sofanthiel && randomizeSofSlugcatPortrait)
                 return $"Multiplayerportrait{UnityEngine.Random.Range(0, 5)}{deadIndex}-{slugcat.value}";
 
-            return $"Multiplayerportrait{(ModManager.MSC && slugcat == MSCScugs.Slugpup ? 4 : colorIndex)}{deadIndex}-{slugcat.value}";
+            if (ModManager.MSC && slugcat == MSCScugs.Slugpup)
+            {
+                if (colorIndex != 0) colorIndex = 4;
```
6fears7, line 47: If the portrait is renamed to 00 and 01, I would think this needs to be 0

6fears7: @invalidunits

Dustylemny (approved):
So far so fine
just slightly inconsistent behaviour on some actions will follow `per slugcat custom colors enabled` behaviour while other actions will follow `global custom colors enabled` behvaiour but overall no crashes or anything

6fears7 (approved):
Approving to bypass a re-approval from Henpe after commit changes, please make sure to review the multiplayerportrait comment and address if needed in a future PR

Menu/Components/SlugcatColorableButton.cs
invalidunits: This is fine since we are adding a new portrait. The original color index 4 one remains.

PR #1007, opened 2025-08-15
fix for wine freebsd

Lizzie841:


Online/Matchmaking/UdpPeer.cs
```diff
@@ -569,22 +569,24 @@ public UDPPeerManager() {
...
+                        port = DEFAULT_PORT + i;
+                        alreadyinuse = activeUdpListeners.Any(p => p.Port == port);
+                        if (!alreadyinuse)
+                            break;
+                    }
+    
+                    if (alreadyinuse) {
+                        throw new Exception("Failed to claim a socket port");
+                    }
+                }  catch (NullReferenceException e) {
```
invalidunits:
```suggestion
                }  catch (Exception e) {
```
Might as well catch everything here.

ghost: I digress, that exception and only that exception is specific to FreeBSD. We don't need to catch everything - it may hinder other errors that may occur.
Unless your plan is also to "ignore if it fails/throws"?

invalidunits: Not a big issue if it throws, since we can just attempt to bind to the default port

invalidunits (approved):

PR #1014, opened 2025-08-21
add piggyback toggle and banning

6fears7:
image

https://github.com/user-attachments/assets/843f6d13-6da1-4400-b8fb-abe2c66887b7

### DIFF
--- a/Arena/ArenaHooks.cs
+++ b/Arena/ArenaHooks.cs
@@ -453,6 +453,10 @@ namespace RainMeadow
                         return false;
                     }
                 }
+                if (isArenaMode(out var arena) && !arena.piggyBack)
+                {
+                    return false;
+                }
             }
             return orig(self);
         }
--- a/Arena/ArenaHooks.cs
+++ b/Arena/ArenaHooks.cs
@@ -453,10 +453,6 @@ namespace RainMeadow
                         return false;
                     }
                 }
-                if (isArenaMode(out var arena) && !arena.piggyBack)
-                {
-                    return false;
-                }
             }
             return orig(self);
         }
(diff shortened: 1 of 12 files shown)


Arena/ArenaHooks.cs
```diff
@@ -453,6 +453,10 @@ private bool CanPutSlugToBack(Func<Player, bool> orig, Player self)
                         return false;
                     }
                 }
+                if (isArenaMode(out var arena) && !arena.piggyBack)
+                {
+                    return false;
```
6fears7: @invalidunits Is this where jumping on heads comes from? Need to know where to return

invalidunits: Piggybacking comes from `Player_GrabUpdatePiggyBack`. You can also override `OnlineGameMode.PlayersCanStack` to disable quick piggy backing

invalidunits (dismissed):

Dustylemny:
> 1. The slug button supports description like that
image
It should, you can check the id matches "scug select" above in ArenaOnlineLobbyMenu.UpdateInfoText() and put the description there.

Im okay with having this but however the main thing is that does it add itself when you become a new host for the lobby?

GameModes/ArenaCompetitiveGameMode.cs
Dustylemny, line 1:
```
         if (arenaClientSettings.playingAs == RainMeadow.Ext_SlugcatStatsName.OnlineRandomSlugcat)
            {
                System.Random random = new System.Random((int)DateTime.Now.Ticks);
                avatarSettings.playingAs = ArenaHelpers.allSlugcats[random.Next(ArenaHelpers.allSlugcats.Count)]!;
                arenaClientSettings.randomPlayingAs = avatarSettings.playingAs;
            }
```
I noticed this, I am assuming we allowing people who pick random to be allowed to play banned slugcats as well? if so feel free to ignore this

invalidunits (approved):

PR #1016, opened 2025-08-22
Vulture Grub and Hazer Sync

Quaxly:
Syncs InitiateSignal in Vulture Grub and prevents AttemptCallVulture from running for remote player
Syncs the spraying bool in Hazers via a state.
Added RealizedHazerState to sync the amount of bites as well.

Will Open the PR up once I've fully tested.

### DIFF
--- a/Game/RainMeadow.CreatureHooks.cs
+++ b/Game/RainMeadow.CreatureHooks.cs
@@ -37,6 +37,34 @@ namespace RainMeadow
             On.Vulture.DropMask += Vulture_DropMask;
             On.BigSpider.BabyPuff += BigSpider_BabyPuff;
             On.VultureGrub.AttemptCallVulture += VultureGrub_AttemptCallVulture;
+
+            IL.Hazer.Update += Hazer_HasSprayed;
+            IL.Hazer.Die += Hazer_HasSprayed;
+        }
+
+        private void Hazer_HasSprayed(ILContext il)
+        {
+            try
+            {
+                var c = new ILCursor(il);
+                c.GotoNext(moveType: MoveType.After,
+                    i => i.MatchLdarg(0),
+                    i => i.MatchLdcI4(1),
+                    i => i.MatchStfld<Hazer>(nameof(Hazer.hasSprayed))
+                    );
+                c.Emit(OpCodes.Ldarg_0);
+                c.EmitDelegate((Hazer hazer) =>
+                {
+                    if (OnlineManager.lobby != null && hazer.abstractPhysicalObject.GetOnlineObject(out var opo) && opo.isMine)
+                    {
+                        opo.BroadcastRPCInRoomExceptOwners(opo.HazerSpraySync, hazer.spraying, hazer.inkLeft);
+                    }
+                });
+            }
+            catch (Exception e)
+            {
+                Logger.LogError(e);
+            }
         }
 
         private void VultureGrub_AttemptCallVulture(On.VultureGrub.orig_AttemptCallVulture orig, VultureGrub self)
--- a/Online/Entity/OnlinePhysicalObject.cs
+++ b/Online/Entity/OnlinePhysicalObject.cs
@@ -628,5 +628,13 @@ namespace RainMeadow
                 grub.vultureCalled = true;
             }
         }
+        [RPCMethod]
+        public void HazerSpraySync(bool spraying, float inkLeft)
+        {
+            if (apo.realizedObject is null || apo.realizedObject is not Hazer hazer) return;
+            hazer.spraying = spraying;
+            if (spraying) hazer.hasSprayed = true;
+            hazer.inkLeft = Mathf.Clamp01(inkLeft);
+        }
     }
 }
--- a/Online/State/RealizedHazerState.cs
+++ b/Online/State/RealizedHazerState.cs
@@ -4,10 +4,6 @@
     {
         [OnlineField]
         byte bites = 3;
-        [OnlineField]
-        bool spraying;
-        [OnlineField]
-        bool tossed;
         public RealizedHazerState() { }
 
         public RealizedHazerState(OnlinePhysicalObject onlineEntity) : base(onlineEntity)
@@ -15,8 +11,6 @@
             var hazer = (Hazer)onlineEntity.apo.realizedObject;
 
             this.bites = (byte)hazer.bites;
-            this.spraying = hazer.spraying;
-            this.tossed = hazer.tossed;
 
         }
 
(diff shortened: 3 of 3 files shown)


Game/RainMeadow.CreatureHooks.cs
```diff
@@ -36,6 +36,24 @@ private void CreatureHooks()
             On.EggBug.DropEggs += EggBug_DropEggs;
             On.Vulture.DropMask += Vulture_DropMask;
             On.BigSpider.BabyPuff += BigSpider_BabyPuff;
+            On.VultureGrub.AttemptCallVulture += VultureGrub_AttemptCallVulture;
+        }
+
+        private void VultureGrub_AttemptCallVulture(On.VultureGrub.orig_AttemptCallVulture orig, VultureGrub self)
+        {
+            if (OnlineManager.lobby != null)
```
youbitchoc: if IsLocal then orig(self)

Online/State/RealizedHazerState.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    public class RealizedHazerState : RealizedPhysicalObjectState
+    {
+        [OnlineField]
+        byte bites = 3;
+        public RealizedHazerState() { }
+
+        public RealizedHazerState(OnlinePhysicalObject onlineEntity) : base(onlineEntity)
+        {
+            var grub = (Hazer)onlineEntity.apo.realizedObject;
```
youbitchoc: rename grub

Online/Entity/OnlinePhysicalObject.cs
```diff
@@ -615,5 +617,13 @@ public void Explode(Vector2 pos)
                     RainMeadow.Error($"unknown explode {this}"); return;
             }
         }
+
```
youbitchoc: imo if you're already making a state then might as well sync it as state
set bool true while hazer is hazing, unset once smoke stops

Game/RainMeadow.GameplayHooks.cs
```diff
@@ -73,6 +75,33 @@ public void GameplayHooks()
             On.SocialEventRecognizer.CreaturePutItemOnGround += SocialEventRecognizer_CreaturePutItemOnGround;
         }
 
+        private void Hazer_Update(On.Hazer.orig_Update orig, Hazer self, bool eu)
+        {
+            bool spraying = false;
+            if (OnlineManager.lobby != null && self.IsLocal())
```
youbitchoc: I would get the opo var, then replace these checks with a simple opo != null

youbitchoc (changes requested):

Game/RainMeadow.CreatureHooks.cs
```diff
@@ -41,18 +41,7 @@ private void CreatureHooks()
...
-                {
-                    Error($"Entity {self} doesn't exist in online space!");
-                    return;
-                }
-                if (!opo.isMine)
-                {
-                    return;
-                }
-            }
+            if (OnlineManager.lobby != null && !self.IsLocal()) return;
```
youbitchoc: you don't need the onlinemanager.lobby check, islocal handles that for you (hence the name)

Game/RainMeadow.GameplayHooks.cs
```diff
@@ -83,22 +83,9 @@ private void Hazer_Update(On.Hazer.orig_Update orig, Hazer self, bool eu)
                 spraying = self.spraying;
             }
             orig(self, eu);
-            if (OnlineManager.lobby != null && self.IsLocal() && spraying != self.spraying)
+            if (OnlineManager.lobby != null && self.IsLocal() && self.abstractPhysicalObject.GetOnlineObject(out var opo))
```
youbitchoc: don't use islocal when you want the object to be online

```cs
if (OnlineManager.lobby != null && self.abstractPhysicalObject.GetOnlineObject(out var opo) && opo.isMine)
```

Game/RainMeadow.GameplayHooks.cs
```diff
@@ -83,22 +83,9 @@ private void Hazer_Update(On.Hazer.orig_Update orig, Hazer self, bool eu)
...
-                if (opo.roomSession.isOwner && (opo.isMine || RPCEvent.currentRPCEvent is not null))
-                {
-                    opo.BroadcastRPCInRoom(opo.HazerSync, self.spraying, self.hasSprayed);
-                }
-                else if (RPCEvent.currentRPCEvent is null)
-                {
-                    if (!opo.isMine) return;  // wait to be RPC'd
-                    opo.roomSession.owner.InvokeOnceRPC(opo.HazerSync, self.spraying, self.hasSprayed);
-                }
+                opo.BroadcastRPCInRoom(opo.HazerSync, self.spraying, self.hasSprayed);
```
youbitchoc: oh, so you meant you were gonna stick with the rpc? bleh sure

youbitchoc (changes requested):

Game/RainMeadow.GameplayHooks.cs
Quaxly: Would it be better to use RealizedHazerState to sync the bool or use this which calls an RPC when it gets changed?

youbitchoc: imo state would be better

Online/State/RealizedHazerState.cs
```diff
@@ -0,0 +1,29 @@
...
+        [OnlineField]
+        bool spraying;
+        public RealizedHazerState() { }
+
+        public RealizedHazerState(OnlinePhysicalObject onlineEntity) : base(onlineEntity)
+        {
+            var hazer = (Hazer)onlineEntity.apo.realizedObject;
+
+            this.bites = (byte)hazer.bites;
+            this.spraying = hazer.spraying;
```
youbitchoc: ~~if spraying is immediately unset then state may not have the opportunity to sync it~~
~~instead you should watch for spraying=true in Hazer.Update then keep it true _until_ the hazer has finished spraying~~

youbitchoc: ~~unfortunately you need to store that data somewhere accessible to RealizedHazerState~~
~~so either keep a CWT of hazers:reallySpraying (kinda silly)
or make an OnlineHazer to store that data (see OnlineBubbleGrass)~~

youbitchoc: ok it isn't immediately unset so this is fine

youbitchoc (approved):

Online/State/RealizedHazerState.cs
Quaxly: awesome sauce, thanks for all the help

PR #1019, opened 2025-08-24
Various fixes and improvements to afk sleep

CptnRedstone:
- Afk sleep can now be triggered manually by holding exclusively down for 5 seconds without stopping.
- Fixed the sleeping animation not cancelling when stunned/killed (technically a vanilla bug, but not one you'd ever see).
- Fixed Spearmaster not closing their eyes during afk sleep.
- Fixed afk sleep rarely triggering when piggybacked onto someone else. I wanted to embrace this behavior but the animation has a lot of issues that'd take way too much effort to fix.
- Fixed the "Don't afk sleep when waking up" check not working (like 4 times).
- Fixed an oddity where shelters closing would cause afk sleep to stop.

### DIFF
--- a/Game/RainMeadow.PlayerHooks.cs
+++ b/Game/RainMeadow.PlayerHooks.cs
@@ -1083,9 +1083,11 @@ public partial class RainMeadow
     public class PlayerExtras
     {
         public bool afkSleep;
+        public int manualSleepDownCounter;
+        public int timeSinceShelterWakeup;
     }
     public int afkSleepRequiredTime = 1200;
-    public int timeSinceShelterWakeup;
+    public int manualSleepRequiredTime = 200;
 
     private void Player_Update1(On.Player.orig_Update orig, Player self, bool eu)
     {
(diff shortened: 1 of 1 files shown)


CptnRedstone:
I don't think a playtest ping would hurt. I'm pretty confident in most of the changes, and ultimately it's pretty much entirely graphical-only, but more eyes is always better. The only potential issue I'm aware of is that you won't be able to afk sleep in a shelter until you've left the shelter at least once, but I certainly don't consider that a major problem.

CptnRedstone:
Actually give me a minute because you made me mentally double-check some stuff and I did just find a new bug.

CptnRedstone:
Fixed. That was the only thing I was worried about really, but, still might as well do a playtest.

Game/RainMeadow.PlayerHooks.cs
```diff
@@ -1163,19 +1163,23 @@ private void Player_Update1(On.Player.orig_Update orig, Player self, bool eu)
...
                 }
             }
-            if (extas.afkSleep)
+            if (extras.afkSleep)
             {
                 self.standing = false;
                 self.sleepCurlUp = Mathf.Max(wasSleepCurlUp, self.sleepCurlUp); // prevent decay
                 self.sleepCurlUp = Mathf.Min(1f, self.sleepCurlUp + 0.02f); // add up
             }
+            if (self.sleepCurlUp > 0 && (self.dead || self.Stunned)) //When stunned or dead, the vanilla code stops decaying sleepCurlUp so we need to handle it ourselves. Technically this fixes an esoteric vanilla bug too.
```
invalidunits: This is intended, Spearmaster cannot sleep. (He draws in his cycle animation)

CptnRedstone: I do not believe this is intended as he still closes his eyes during the regular sleep animations. I believe that code is intended to make him open his eyes slightly faster when waking up. If it *is* intended, then it still needs adjustment because that same code causes him to blink rapidly instead of just keeping his eyes open normally.

invalidunits:
Thoughts on allowing manual sleep by holding down like jolly Co-OP?

CptnRedstone:
Would be a bit annoying since it's currently based off of touchedNoInputCounter, but I could probably find a way to get that working smoothly enough. Will look into it as long as I don't hear a "wait hold on" from anyone.

invalidunits (approved):

PR #1025, opened 2025-08-27
Timeline fix

invalidunits:


6fears7 (approved):
LGTM, it looks like LoadTimelineIn is the new version of LoadWorldAs

PR #1026, opened 2025-08-27
OverworldSession Resource

invalidunits:
I split off the management of WorldSessions from the Lobby into a new OverworldSession resource. 

- Allows players to manage there regions even when the host hasn't loaded in yet. (Meadow mode)
- Adds potential for downloading custom regions while inside the lobby.

This doesn't have to be merged immediately.

### DIFF
--- a/Online/OnlineManager.cs
+++ b/Online/OnlineManager.cs
@@ -340,28 +340,22 @@ namespace RainMeadow
         {
             if (lobby != null)
             {
+                if (rid == "@overworld") return lobby.overworld;
                 if (rid == ".") return lobby;
-
-                if (rid == "arena" && lobby.worldSessions.TryGetValue(rid, out var arenaRegionz)) return arenaRegionz;
-
-                if (rid.Contains("arena"))
+                if (lobby.overworld.isActive)
                 {
-                    string modifiedRid = rid.Replace("arena", "");
-
-                    if (lobby.worldSessions["arena"].roomSessions.TryGetValue(modifiedRid, out var roomSession))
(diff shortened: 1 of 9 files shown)


Online/OnlineManager.cs
```diff
@@ -340,28 +340,22 @@ public static OnlineResource ResourceFromIdentifier(string rid)
         {
             if (lobby != null)
             {
+                if (rid == "@overworld") return lobby.overworld;
                 if (rid == ".") return lobby;
-
-                if (rid == "arena" && lobby.worldSessions.TryGetValue(rid, out var arenaRegionz)) return arenaRegionz;
-
-                if (rid.Contains("arena"))
```
6fears7, line 362: I have some  concerns on this change with arena RIDs.

As for everything else, I'm wondering if we can explore shifting some of the lobby-level data in story & arena to the overworld. Especially for story, it could put us [one step closer](https://www.youtube.com/watch?v=sKtl7-_6AjE&list=RDsKtl7-_6AjE&start_radio=1) to multi-region support in the future.

Overall, just needs aggressive testing. Hopefully I can get to some of that soon.

henpemaz (approved):
Not gonna lie, I'm a bit scared of this one, but hey if it works it works.

I had a similar thing in one of my earliest prototypes but discarded it due to redundancy, but there was no region initialization to worry about back then.

The biggest downside might be the lockstep-requesting of both overworld and then region when the client goes in-game, maybe one day we should add a timeout to those.

henpemaz (approved):

PR #1023, opened 2025-08-27
Update StoryOnlineMenu.cs

6fears7:
https://github.com/henpemaz/Rain-Meadow/pull/898/files#diff-7d48a5d354ad0f626abf2b9322a24527765f6c7b9f2ca91180d494ba763236f3

This was regressed from a previous fix. Please let me know if this works for you.

PR #1029, opened 2025-08-31
Story menu fix

invalidunits:
Replacing #1023

### DIFF
--- a/Story/StoryLobbyData.cs
+++ b/Story/StoryLobbyData.cs
@@ -235,6 +235,7 @@ namespace RainMeadow
             public bool hasGlow;
             public bool hasMark;
             public bool ascended;
+            public bool altEnd;
             public string shelterName;
             public int gameTimeAlive;
             public int gameTimeDead;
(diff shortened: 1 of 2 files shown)


Story/StoryLobbyData.cs
```diff
@@ -164,26 +201,158 @@ public override void ReadTo(ResourceData data, OnlineResource resource)
...
+                    serializer.writer.Write(hasGlow);
+                    serializer.writer.Write(hasMark);
+                    serializer.writer.Write(ascended);
+                    serializer.writer.Write(altEnd);
+
+                    serializer.writer.Write(cycle);
+                    serializer.writer.Write(gameTimeAlive);
+                    serializer.writer.Write(gameTimeDead);
+                    serializer.writer.Write((byte)karma);
+                    serializer.writer.Write((byte)Mathf.FloorToInt(rippleLevel / 0.25f));
```
6fears7, line 310: interesting

6fears7 (approved):
With changes like this, obviously having a few playtests / extended playtest would be desired. Beyond that, LGTM. Seems like you're going for a route of managing the menu state  as a "state" to clean up perhaps any deviant logic a client might be experiencing? I'd be curious to know what motivated this change. 

Also, looks like you removed the Slug pup rain background for clients? :(

PR #1030, opened 2025-09-01
fix: HitSomething: fixes the reference for didParry

6fears7:
- Fixes the didParry reference
- Hoping this improves local client protection from being killed when they parried on their end.

### DIFF
--- a/Online/Entity/OnlinePhysicalObject.cs
+++ b/Online/Entity/OnlinePhysicalObject.cs
@@ -550,7 +550,7 @@ namespace RainMeadow
                 hit.BuildCollisionResult(out result);
                 if (result.HasValue) {
                     OnlinePhysicalObject? onlineResult = result.Value.obj.abstractPhysicalObject.GetOnlineObject();
-                    if ((OnlineManager.lobby != null) && onlineResult != null && onlineResult.didParry)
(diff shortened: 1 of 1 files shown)


Online/Entity/OnlinePhysicalObject.cs
```diff
@@ -543,18 +543,19 @@ public void BuildCollisionResult(out SharedPhysics.CollisionResult? result) {
...
                 statewhenhit.ReadTo(this);
                 SharedPhysics.CollisionResult? result = null;
                 hit.BuildCollisionResult(out result);
                 if (result.HasValue) {
+                    OnlinePhysicalObject? onlineResult = result.Value.obj.abstractPhysicalObject.GetOnlineObject();
+                    if (OnlineManager.lobby != null && onlineResult != null && onlineResult.didParry)
+                    {
+                        RainMeadow.Debug("Parried!");
+                        OnlineManager.RunDeferred(() => onlineResult.didParry = false);
+                        return;
```
invalidunits, line 558:
```suggestion
                        HittingRemotely = false;
                        return;
```

invalidunits (changes requested):
Set Hitting Remotely to false before returning.

invalidunits (approved):

henpemaz (approved):

PR #1033, opened 2025-09-04
Fix saint ending:

6fears7:
- Fixes Saint ending not running. 
~~- Also it removes all other **non local** Abstract Creatures from the rooms during void sea. How did this work without always nuking the game for a local player?~~

### DIFF
--- a/Game/RainMeadow.EntityHooks.cs
+++ b/Game/RainMeadow.EntityHooks.cs
@@ -363,16 +363,11 @@ namespace RainMeadow
                         }
                         StoryRPCs.GoToSaintEnding(null);
                     }
-                    else
-                    {
-                        orig(self, callback, roomName);
-                    }
+
                 }
             }
-            else
-            {
-                orig(self, callback, roomName);
-            }
+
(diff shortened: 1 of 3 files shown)


Game/RainMeadow.EntityHooks.cs
```diff
@@ -343,9 +343,31 @@ public void OverWorld_InitiateSpecialWarp_SingleRoom(On.OverWorld.orig_InitiateS
...
+                        else if (RPCEvent.currentRPCEvent is null)
+                        {
+                            // tell host to move everyone else
+                            OnlineManager.lobby.owner.InvokeOnceRPC(StoryRPCs.GoToSaintEnding);
+                        }
+                        StoryRPCs.GoToSaintEnding(null);
+                    }
+                    else
+                    {
+                        orig(self, callback, roomName);
```
6fears7, line 368: should probably add back orig

invalidunits:
Try overriding `ShouldSyncAPOInWorld` to return false when we are in the final cutscene. Everything else looks good.

PR #1038, opened 2025-09-09
Better Controller UI navigation

CptnRedstone:
Fixes and improves many bugs and frustrations related to navigating meadow UIs with a controller; as-is many buttons jump to unexpected places, and some are difficult or impossible to reach as a result. Includes frameworks for making future improvements far easier. Adds a meadow logging "warn" channel (why don't we have one?). Full list of changes follows:

User-facing:
	- The lobby creation, meadow lobby, and meadow pause menus have had their binds completely fixed and marginally redesigned for more intuitive movement.
	- Similarly, the lobby selection and arena lobby menus have significant, and marginal improvements respectively.
	- The lobby creation and arena lobby menus have smarter default cursor positions.

Known issues:
	- Certain parts of the arena lobby menu are still horrible to navigate. I may work on additional fixes for this later, but the problems all stem from the dynamic lists, which will be extremely difficult to programmatically fix due to limitations in determining an element's actual position within the UI. The lobby select menu has similar problems, but they're *far* less intrusive due to having a much lower list density.
	- The lobby select menu's "sort lobbies by X" button should move to the refresh button when up is pressed, but it jumps to a different button instead. I have no idea why this happens, how to fix it, or what makes this one particular button problematic (and not the almost identical one right below).

Technical:
	- Certain existing bind functions now run as needed instead of every tick.
	- Adds 6 new extension functions to help refine and automate rebinding UI elements. These should make all future UI binding operations trivial, with the sole exception of the aforementioned variable-length lists.
	- Added a RainMeadow.Warn() logging channel. Went to go use it, and found we didn't actually have one for some reason, so now we do. If for some reason a warn channel is *unwanted*, the Warn()s I wrote should be changed to Error() rather than Debug().

### DIFF
--- a/Extensions.cs
+++ b/Extensions.cs
@@ -334,5 +334,30 @@ namespace RainMeadow
                 }
             }
         }
+        ///<summary>Directionally binds a list of UI elements to a target element. For example, with fromObjects[A,B,C] and ToObject D, Bind A→D, B→D, and C→D.</summary>
+        public static void TryMassBind(List<MenuObject> fromObjects, MenuObject toObject, bool left = false, bool right = false, bool top = false, bool bottom = false)
+        {
+            foreach (MenuObject fromObject in fromObjects)
+            {
+                TryBind(fromObject, toObject, left, right, top, bottom);
+            }
+        }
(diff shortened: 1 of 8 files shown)


invalidunits:
Looks, good. 

Whoever merges this should use the Squash and Merge option.

PR #1042, opened 2025-09-12
Fix Arena Timer being invisible in some languages

Dustylemny:


Arena/ArenaTimer.cs
```diff
@@ -86,7 +86,8 @@ public override void Draw(float timeStacker)
                 if ((safetyCatchTimer > 300)) // Something went wrong with the timer. Let's move on
                 {
                     showMode = TimerMode.Countdown;
-                };
+                }
+                ;
```
6fears7: awkward floating semi colon but no big deal

Dustylemny: huh seems like my ide automatically does that

PR #1044, opened 2025-09-16
Chatbar Fixes

CptnRedstone:
- Fixes the ingame and arena chatbars' cursor and text selection graphics being significantly, and slightly displaced to the left respectively.
- Fixes an issue with the ingame chatbar where Shift+Left when already at the textbox edge would reset the selection.
- Fixes an issue with the ingame chatbar where Ctrl+Shift+Left when already at the textbox edge would spam logs with OOBs.
- Fixes the ingame chatbar's keybinds' repeat rate and delay being affected by framerate.
- Fixes the ingame chatbar's visual and actual cursor position not necessarily being synced.
- Fixes an issue with the arena chatbar where Home, End, and Ctrl A would increasingly fail to register as FPS increases.

### DIFF
--- a/Menu/Components/ChatTextBox2.cs
+++ b/Menu/Components/ChatTextBox2.cs
@@ -118,7 +118,7 @@ namespace RainMeadow.UI.Components
             if (cursorIsInMiddle)
             {
                 if (cursorSprite.element.name != "pixel") cursorSprite.SetElementByName("pixel");
-                cursorSprite.x = cursorPosition + 10 + screenPos.x;
+                cursorSprite.x = cursorPosition + 11 + screenPos.x;
                 cursorSprite.height = 13;
             }
             else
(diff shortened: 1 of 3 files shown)


6fears7:
@Dustylemny please review

Dustylemny:
Lgtm

side note tested and certain circumstances like echos can lower game ticks which makes in-game chatbar look noticeably slower in for selection inputs

CptnRedstone:
@6fears7 @Dustylemny Should be fixed now. I also adjusted the delay (0.75 → 0.5) and repeat rate (20/s → 30/s) to make it a bit snappier and better align with holding regular keys (like A or backspace). Assuming those timings are unity and not OS. Super easy to change now either way.

CptnRedstone:
Actually should I apply the deltatime logic to the arena chatbar also, or leave well enough alone?

6fears7:
keep it untouched at this present time

PR #1045, opened 2025-09-17
Profiler

Quaxly:
Adds more profiling capabilities to Rain Meadow which should be useful for debugging those really laggy regions like Five Pebbles and optimizing the engine more.

When using Dev tools press "=" to switch to the profiler view. 

Advanced profiling can be turned on in the remix setting. (Doesn't seem to be working just yet so for now just edit the code) Upon restart the game will perform a large Harmony patch to allow profiling individual methods.

With Advanced Profiling on in the Profiling menu press "-" to create a collapsed stack file. This file will be in you game's root folder and can be viewed on a site like https://www.speedscope.app/

image

### DIFF
--- a/Game/MeadowRemixOptions.cs
+++ b/Game/MeadowRemixOptions.cs
@@ -56,8 +56,6 @@ public class RainMeadowOptions : OptionInterface
 
     public readonly Configurable<IntroRoll> PickedIntroRoll;
 
-    public readonly Configurable<bool> AdvancedProfiling;
-
     public enum IntroRoll
     {
         Meadow,
@@ -139,8 +137,6 @@ public class RainMeadowOptions : OptionInterface
         StopMovementWhileSpectateOverlayActive = config.Bind("StopMovementWhileSpectateOverlayActive", false);
 
         ChatBgOpacity = config.Bind("ChatBgOpacity", 0.2f);
-
-        AdvancedProfiling = config.Bind("AdvancedProfiling", false);
     }
 
     public override void Initialize()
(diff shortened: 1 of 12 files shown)


Quaxly:
The profiler view bc I forgot to upload it. You'll get some basic information like FPS, Memory Usage and Bandwidth on the left side and on the right is a pie chart for visualizing resource usage. You can switch charts using '[' and ']'. 

The Pie Chart currently has State Write Times, Entity count, Online Physical Object count, Failed Comparisons (thanks Invalid) and Player Writes. This should help a lot at figuring out a lot of performance related issues with Meadow.

https://github.com/user-attachments/assets/1bb6ea5e-5c67-463c-bda1-21854d171e8f

henpemaz (approved):
Looks awesome.

Would be nice to include the MakeState methods for the multiple state classes, as well as the delta-encoding and decoding methods in rootdeltastate.

I wonder if it might be the case of automatically patching methods on the own mod going forward, with an opt-in or opt-out annotation.

Either way these can be addressed in a future PR, merge at your discretion

6fears7:
@Quaxly You good to merge this?

Game/MeadowRemixOptions.cs
```diff
@@ -70,6 +70,7 @@ public enum IntroRoll
     private UIelement[] OnlineArenaSettings;
     private UIelement[] OnlineStorySettings;
     private UIelement[] OnlineLANSettings;
+    private UIelement[] OnlineAdvancedSettings;
```
6fears7, line 73: unused?

Quaxly: Yes, it was for the original advanced profiler setting but I moved it to use a launch parameter instead.

henpemaz (approved):

PR #1046, opened 2025-09-17
Pointing tweaks and fixes

CptnRedstone:
- Scugs now prefer to point with their primary hand, rather than their offhand, matching both jolly and community-preferred behavior. This by proxy fixes pups being entirely unable to point with items.
- Fixed an issue where spear pointing angles would completely break when normal walking left/right.
- Fixed an issue that'd sometimes select the fallback hand to point, even if the preferred was usable. This could cause pointing with the "wrong" item, item position freakouts for other clients, and I suspect it was the cause for occasional failure to point when crouched (as I can no longer reproduce that).


6fears7:
- [ ] Need to test this myself following convo with Redstone to see if jitter is introduced (I doubt it) or was always present

PR #1051, opened 2025-09-22
Scavenger and grasp stealing fixes

invalidunits:


Game/RainMeadow.CreatureHooks.cs
```diff
@@ -36,6 +40,36 @@ private void CreatureHooks()
             On.EggBug.DropEggs += EggBug_DropEggs;
             On.Vulture.DropMask += Vulture_DropMask;
             On.BigSpider.BabyPuff += BigSpider_BabyPuff;
+
+            On.Creature.Grab += Creature_Grab;
+            On.Creature.SwitchGrasps += Creature_SwitchGrasps;
+        }
+
+
+        private bool Creature_Grab(On.Creature.orig_Grab orig, Creature self, PhysicalObject obj, int graspUsed, int chunkGrabbed, Creature.Grasp.Shareability shareability, float dominance, bool overrideEquallyDominant, bool pacifying)
```
6fears7, line 89: My preference with any hook is to do an early return if the OnlineManager.lobby is null just as a safety/ sanity check that we are not introducing anything into base game while meadow is installed vs at the end. It's not a blocker, just a consideration.

Game/RainMeadow.GameplayHooks.cs
```diff
@@ -535,7 +535,7 @@ private bool Weapon_HitSomething<WeaponT>(Weapon_orig_HitSomething<WeaponT> orig
 
                 return ret;
             }
-            else if (self.IsLocal())
+            else if (WeaponOnline.isMine || WeaponOnline.isPending)
```
6fears7, line 554: > isPending => isRequesting || isReleasing || isWaitingForState
 
 I don't know that a new RealizedSpearState will be useful if we're in the middle of the transition of ownership,
 
```
        {
            base.ReadTo(onlineEntity);
            if (onlineEntity.isTransfering)
            {
                RainMeadow.Debug($"Transferring {onlineEntity}, not updating state!");
                return;
            }
```
 
  but I imagine that the only time this condition evaluates to true is if its watingForState, so perhaps it doesn't matter.

Online/Serialization/Serializer.Data.cs
```diff
@@ -1329,43 +1330,43 @@ public void Serialize(ref Dictionary<string, int> data)
             }
         }
 
-            public void Serialize(ref Dictionary<int, int> data)
-            {
+        public void Serialize(ref Dictionary<int, int> data)
```
6fears7: I think this section was just formatting changes

Online/Serialization/Serializer.Data.cs
```diff
@@ -1532,6 +1533,40 @@ public void Serialize(ref List<KeyValuePair<byte, ushort>> data)
             }
 #if TRACING
             if (IsWriting) RainMeadow.Trace(this.Position - wasPos);
+#endif
+        }
+
+        public void SerializeNullableValueType<T>(ref T? value) where T : struct
```
6fears7: This is curious. Normally I'd expect the data type to be annotated with `"nullable=true"` in the state field, but I suppose you're looking to serialize a type of `T?`, so then it'd need both a `nullable=true` AND be of type `T?` Is this a scenario necessary or can we simply mark whatever you're using this for as a nullable value in the state field?

Online/State/AI/ArtificialIntelligenceState.cs
```diff
@@ -4,42 +4,51 @@
...
-            if (onlineCreature.apo is AbstractCreature creature) {
-                if (creature.creatureTemplate.AI && creature.abstractAI.RealAI is not null) {
-                    InitializefromAI(creature.abstractAI.RealAI);
-                }
-            }
-        }
-        public void InitializefromAI(ArtificialIntelligence artificialIntelligence) {
+
+        [OnlineField(nullable: true)]
+        public OnlineEntity.EntityId? focusCreature;
```
6fears7, line 14: I see now your use case. I'll defer to henpe on this

6fears7 (dismissed):

Online/Serialization/Serializer.Data.cs
```diff
@@ -1532,6 +1533,40 @@ public void Serialize(ref List<KeyValuePair<byte, ushort>> data)
...
+#if TRACING
+            long wasPos = this.Position;
+#endif
+            if (IsWriting)
+            {
+                writer.Write(value.HasValue);
+                if (value.HasValue)
+                {
+                    var inputParams = new object[] { value };
+                    GetSerializationMethod(typeof(T), false, false, false).Invoke(this, inputParams);
```
henpemaz: Mhhh sorry no, a lookup like this in the middle of serialization code is a no-no, a reflection invoke is just as bad too.

If we need this functionality it need to be built into a dynamic method through IL-emit or expressiontrees.

henpemaz (changes requested):

6fears7 (dismissed):

Online/State/RealizedPhysicalObjectState.cs
```diff
@@ -84,10 +90,22 @@ public void CustomSerialize(Serializer serializer)
...
+
+            if ((pos - c.pos).sqrMagnitude < (closeEnough * closeEnough))
+            {
+                c.pos = Vector2.Lerp(pos, c.pos, 0.5f);
+            }
+            else
+            {
+                c.pos = pos;
+            }
+
```
henpemaz: Not a fan of universal easing of all bodychunks.

Is this for objects held in hand so they aren't as jittery? If so lets limit it to objects in hand.

henpemaz: On that note, we're also marking grasped objects as pos-lenient, so wouldn't this be redundant?

Online/Serialization/Serializer.ICustomSerializable.cs
```diff
@@ -257,6 +263,103 @@ internal static MethodInfo MakeSerializationMethod(Type fieldType, bool nullable
...
+            {
+                // T internalvalue = default(T);
+                // if (serializer.isWriting)
+                // {
+                //     serializer.writer.Write(value.HasValue);
+                //     if (value.HasValue)
+                //     {
+                //          internalvalue = value.Value;
+                //     }
+                //     goto callSerialize;
```
henpemaz: I think this is wrong. If there's no value, there's no need to serialize it.

In fact serializing it would desync the reader, because the reader will skip reading it.

Online/Serialization/Serializer.ICustomSerializable.cs
```diff
@@ -257,6 +263,103 @@ internal static MethodInfo MakeSerializationMethod(Type fieldType, bool nullable
...
+
+                il.Emit(OpCodes.Ldarga, 1);
+                il.Emit(OpCodes.Ldloc, internalValue);
+                il.Emit(OpCodes.Newobj, fieldType.GetConstructor([t]));
+                il.Emit(OpCodes.Stobj, fieldType);
+                il.Emit(OpCodes.Ret);
+                return dynMethod;
+            }
+
             return typeof(Serializer).GetMethod(nullable ? "SerializeNullable" : "Serialize", new[] { fieldType.MakeByRefType() });
```
henpemaz, line 374: Solid stuff with the IL generation.

I think it'd be preferable though if we used a hand-written method from Serializer where available, before defaulting to generate.

So the steps would be roughly:
- If serializer has suitable method, return method
- If generatable, generate
- Raise method not found

Online/Serialization/Serializer.ICustomSerializable.cs
invalidunits: good catch

Online/State/RealizedPhysicalObjectState.cs
invalidunits: This is a leftover from another PR, I'll remove it when I can.

6fears7 (approved):
Tested, worked

Game/RainMeadow.GameplayHooks.cs
```diff
@@ -551,7 +551,7 @@ private bool Weapon_HitSomething<WeaponT>(Weapon_orig_HitSomething<WeaponT> orig
...
```
6fears7, line 554: The phase fix was here all along

henpemaz (approved):
Looks great! It's awesome to see someone mastering all these tools!

PR #1054, opened 2025-09-24
Multipage handling for arena scug selection

CptnRedstone:
Handling for multiple pages of slugcats in the arena character select menu, so really high counts of modded scugs don't start to flow off of the screen. Includes full rebinds for controller support, and will only render one row when a second isn't needed for a cleaner look.
Also forces the "random" icon to be between vanilla and modded scugs, and adds a proper modulo function to extensions, since for some abhorrent reason C#'s % doesn't handle negative dividends correctly.

Based on https://github.com/henpemaz/Rain-Meadow/pull/1038, will require that to be merged first.


6fears7:
Depends on #1038

henpemaz (approved):

PR #1056, opened 2025-09-24
Story client debug info on dev tools 

CptnRedstone:
In story lobbies, players' names in the dev tools overlay will have an additional [] that will contain any amount of the following flags:
S = Ready to shelter
G = Trying to go through a gate
D = Dead
L = Waiting in lobby
This is intended to make it easier for lobby hosts to identify who is preventing a gate or shelter from working, whether a bug or not.

### DIFF
--- a/Online/DebugOverlay.cs
+++ b/Online/DebugOverlay.cs
@@ -1,3 +1,4 @@
+using Ionic.Zlib;
 using RWCustom;
 using System;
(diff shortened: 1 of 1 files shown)


henpemaz (approved):
Good functionality to have for story

That second TryGetData should maybe be inside the if for safety but can be addressed later

PR #1062, opened 2025-09-27
DLL Cache

invalidunits:
Store all dynamic methods in a into a DLL. This theoretically makes the startup faster and works great with Dnspy for debugging serialization.


6fears7 (dismissed):
This is pretty far outside my C# wheelhouse. Great idea though for debugging. Blindly approving and deferring to Henpe

henpemaz:
This does seem cool for debugging, as for performance we're talking about 0.4s to generate those. Lets wait for the watcher dust to settle a bit before we get this and the overlay changes in

PR #1063, opened 2025-09-27
Revert 992 load when available

6fears7:


6fears7:
sorry, I have an actual PR I was working from. This is irrelevant

PR #1069, opened 2025-09-28
Watcher50 storymenu

invalidunits:
Hopefully the last story menu update

### DIFF
--- a/Story/StoryOnlineMenu.cs
+++ b/Story/StoryOnlineMenu.cs
@@ -532,18 +532,26 @@ namespace RainMeadow
         {
             if (OnlineManager.lobby.isOwner)
             {
+                for (int i = 0; i < slugcatPages.Count; i++)
+                {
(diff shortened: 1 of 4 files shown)


Story/StoryOnlineMenu.cs
```diff
@@ -535,22 +533,53 @@ private void SetupSlugcatList()
 
         void RefreshPages()
         {
-            if (!OnlineManager.lobby.isOwner || !isPrevOwner)
+            if (OnlineManager.lobby.isOwner)
```
6fears7, line 536: What do you think of my implementation? check isPrevOwner

invalidunits: Copied from discord

The top half of the RefreshPage if exists so that If the lobby get's transferred to you, you'll see your own savefiles instead of the previous owners.

Of course if your implementation is more stable feel free to revert these changes.

PR #1073, opened 2025-10-02
1.7.1

6fears7:
## Hotfix 1.7.1 TODO

- [x] - Force modmanager restart on MSC/Watcher change
image

- [x] - Add dropslug() to dead scugs
- [x] - (maybe) resolve watcher warp null refs
- [x] - Fix SetInteractionState on chat process shutdown
- [x] ???

Anyone got anything else they feel should be included in this batch?


niacdoial:
I don't know if it's new with 1.7 or if it's somehow exclusive to the router stuff, but there was a bug where, when somebody changed bodies in Meadow mode, some other player might half-freeze (nothing can move but particle effects and audio still play) while NREs are spammed in the console.
I already bandaid-fixed this in a different PR, but I wonder if it's linked to a deeper problem.

henpemaz (approved):

PR #1074, opened 2025-10-02
fix? NRE sometimes happening when another player changes species

niacdoial:
actually I have no idea if this might cause more issues but I don't think so given there's already something that discards nulls in this line


henpemaz (approved):
Eh usually I'm not a fan of tossing null checks where not absolutely necessary, I'll let this one slip

6fears7:
@niacdoial you want to keep or throw this PR?

PR #1075, opened 2025-10-03
Multi Screen Arena usability.

Quaxly:
Currently the multi screen arenas are practically unusable due to a lack of items among a few other issues. This PR hopes to address this by adding support for loading our own settings in place of the vanilla settings. That way we can add more items and change settings around if needed.

To change an arena's settings in Meadow only simply make a settings file and end it with `_meadowsettings.txt`

I'll go through each multi arena and add more items where they're needed.


henpemaz (approved):
Good to merge, we could add that new folder as public in CODEOWNERS too

PR #1087, opened 2025-10-07
Sync Watcher Objects and Hazards.

Quaxly:
The Watcher DLC added a lot of room hazards that need to be synced. The following need to be synced.
- [x] Geysers (Uses `RainWorldGame.clock` for it's timing)
- [x] WaterLevelCycle (Includeds FluxWaterFall & FluxDrain, uses new parameters in RainCycle)
- [ ] FlameJet
- [ ] LethalThunderStorm
- [ ] SKLightningRod

If there's anything else that needs to be synced I'll add it to this list.

### DIFF
--- a/Online/Resource/WorldSession.cs
+++ b/Online/Resource/WorldSession.cs
@@ -143,6 +143,10 @@ namespace RainMeadow
                     cycle.timer = rainCycleData.timer;
                     cycle.cycleLength = rainCycleData.cycleLength;
 
+                    cycle.waterCycle.timeInStage = rainCycleData.timeInStage;
+                    cycle.waterCycle.stageDuration = rainCycleData.stageDuration;
+                    cycle.waterCycle.stage = (WaterLevelCycle.Stage)rainCycleData.stage;
+
                     if (realizedRooms != null)
                     {
                         foreach (var index in realizedRooms.list)
--- a/Online/Resource/WorldSession.cs
+++ b/Online/Resource/WorldSession.cs
@@ -106,7 +106,8 @@ namespace RainMeadow
             [OnlineField(nullable: true)]
             public Generics.DynamicOrderedUshorts realizedRooms;
 
-            // Used to sync RainWorldGame.clock which is used for a few 
+            // Used to sync RainWorldGame.clock which is used for a few APOs introduced in the Watcher as well as a few
+            // other functions in the game. Should be safe to update this way but I would keep a watch on it for a little bit.
             [OnlineField]
             public int clock;
             public WorldState() : base() { }
--- a/Online/Resource/WorldSession.cs
+++ b/Online/Resource/WorldSession.cs
@@ -105,11 +105,17 @@ namespace RainMeadow
             public RainCycleData rainCycleData;
             [OnlineField(nullable: true)]
             public Generics.DynamicOrderedUshorts realizedRooms;
+
+            // Used to sync RainWorldGame.clock which is used for a few 
+            [OnlineField]
+            public int clock;
             public WorldState() : base() { }
             public WorldState(WorldSession resource, uint ts) : base(resource, ts)
             {
                 if (resource.world != null)
                 {
+                    clock = resource.world.game.clock;
+
                     RainCycle rainCycle = resource.world.rainCycle;
                     if (rainCycle.brokenAntiGrav == null)
                     {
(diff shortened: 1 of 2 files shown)


Online/Resource/WorldSession.cs
```diff
@@ -128,11 +135,18 @@ public override void ReadTo(OnlineResource resource)
                 if (resource.isActive)
                 {
                     var ws = (WorldSession)resource;
+
+                    ws.world.game.clock = clock;
```
invalidunits, line 140: Confirm that the clock never regresses
```suggestion
                    ws.world.game.clock = Mathf.max(clock, ws.world.game.clock);
```

Online/Resource/WorldSession.cs
```diff
@@ -128,11 +135,18 @@ public override void ReadTo(OnlineResource resource)
...
+
                     RainCycle cycle = ws.world.rainCycle;
                     cycle.preTimer = rainCycleData.preTimer;
                     cycle.timer = rainCycleData.timer;
                     cycle.cycleLength = rainCycleData.cycleLength;
 
+                    cycle.waterCycle.timeInStage = rainCycleData.timeInStage;
```
invalidunits, line 147:
```suggestion
                    cycle.waterCycle.timeInStage = Mathf.max(rainCycleData.timeInStage, cycle.waterCycle.timeInStage);
```

Online/Resource/WorldSession.cs
youbitchoc: @invalidunits does this affect devtools clock reset?

Online/Resource/WorldSession.cs
youbitchoc: @invalidunits does timeinstage reset for every stage? if so, this check would prevent that reset

Game/RainMeadow.GameHooks.cs
```diff
@@ -62,6 +62,25 @@ private void GameHooks()
             IL.Menu.SleepAndDeathScreen.GetDataFromGame += SleepAndDeathScreen_FixNullKarmaLadder;
 
             On.ProcessManager.CueAchievement += ProcessManager_CueAchievement;
+
+            On.GlobalRain.InitDeathRain += GlobalRain_InitDeathRain;
+        }
+
+        private void GlobalRain_InitDeathRain(On.GlobalRain.orig_InitDeathRain orig, GlobalRain self)
```
6fears7, line 69: We should early return if the lobby is null and call orig

Online/Resource/WorldSession.cs
```diff
@@ -106,11 +106,18 @@ public class WorldState : ResourceWithSubresourcesState
             public RainCycleData rainCycleData;
             [OnlineField(nullable: true)]
             public Generics.DynamicOrderedUshorts realizedRooms;
+
+            // Used to sync RainWorldGame.clock which is used for a few APOs introduced in the Watcher as well as a few
+            // other functions in the game. Should be safe to update this way but I would keep a watch on it for a little bit.
```
6fears7, line 111: Definitely need some playtesting first before merge if this is the case!

Online/State/RealizedBigMothState.cs
```diff
@@ -0,0 +1,87 @@
...
+namespace RainMeadow
+{
+    public class RealizedBigMothState : RealizedCreatureState
+    {
+        [OnlineFieldHalf]
+        public Vector2 moveDirection;
+        [OnlineField]
+        bool drinkingChunk;
+        [OnlineField(nullable = true)]
+        BodyChunkRef drinkChunk;
```
6fears7: I would think we should mark BodyChunkRef as nullable since the original methods assume its nullable iirc

Online/Resource/WorldSession.cs
Quaxly: tbh I think even if the value was being skewed all over the place it'd be totally fine. `RainWorldGame.clock` was added pretty recently for just the Watcher iirc.

Online/State/AbstractPhysicalObjectState.cs
```diff
@@ -61,6 +61,7 @@ protected virtual RealizedPhysicalObjectState GetRealizedState(OnlinePhysicalObj
             if (onlineObject.apo.realizedObject is SlimeMold) return new RealizedSlimeMoldState(onlineObject);
             if (onlineObject.apo.realizedObject is VultureGrub) return new RealizedVultureGrubState(onlineObject);
             if (onlineObject.apo.realizedObject is Hazer) return new RealizedHazerState(onlineObject);
+            if (onlineObject.apo.realizedObject is Watcher.BoxWorm.Larva) return new RealizedFireSpriteLarva(onlineObject);
```
6fears7: Please Remove this if its not ready, it causes crashes

6fears7 (dismissed):

6fears7:
If you were interesting in syncing SandGrubs, that would be a great task to complete so I can pursue other issues to t/s
[online_sav3.zip](https://github.com/user-attachments/files/23569354/online_sav3.zip)
Badlands save.

PR #1101, opened 2025-10-19
Allow SporePlants (Beehives) in Arena.

Quaxly:
Fixes the Gourmand desync bug when regurgitating and swallowing a beehive by allowing beehives to be synced in Arena Competitive. Beehives themselves will be synced while `SporePlant.AttachedBee` will remain client side. 

Beehives don't seem to cause any major performance issues anymore ever since `SporePlant.AttachedBee` was set to client-side. Will adjust and unblock beehives in other areas (like arena spawns) upon further feedback.


6fears7 (approved):
Just gonna send it

PR #1105, opened 2025-10-25
Fix piggybacking through gates and warps

invalidunits:


6fears7 (approved):
I like making the function, "RemoveEntityFromGame". In terms of arena and story those changes look fine, won't speak for the leaveImpl diff

PR #1106, opened 2025-10-25
Restart button for arena

invalidunits:


invalidunits:
Some of the fixes from #1105 could be applied here. It just depends on which one gets merged first.

invalidunits:
It's could useful if something goes wrong in a round, such as in a minigame lobby or when testing features. It keeps the round number, doesn't save kill credit, and won't send you to the final result screen.

6fears7:
I'll take a closer look in a few days when I return

6fears7 (dismissed):

PR #1107, opened 2025-10-28
Allow two-way echo warps

Dustylemny:
Currently forced one-way echo warps can softlock host from achieving 3rd ending. And two-way echo warps work pretty fine so we can remove it

also adds host to mark and save the echo a client just met because you wouldn't want to have both the portal and echo together.

also includes
- A prevention to allow players to join after warping to daemon
- isAscended check to prevent going to ripple karma screen whenever going near/to an echo room after already achieving 1st ending

Rn keeping an ilhook that overrides the saved portal's data found in room with spinning top's portal data on SpinningTop.SpawnBackupPortal to allow hosts to continue where they left off instead of them having to restart their watcher campaign when they find themself softlocked

### DIFF
--- a/Story/StoryRPCs.cs
+++ b/Story/StoryRPCs.cs
@@ -260,8 +260,11 @@ namespace RainMeadow
             //}
             if (!OnlineManager.lobby.isOwner)
             {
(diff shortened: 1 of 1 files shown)


6fears7 (approved):
It's all pain and I'm here for it

PR #1109, opened 2025-11-01
Lizzie's Sand grub fix

Dustylemny:


6fears7 (approved):
Throw it onto the stack of watcher jank

PR #1117, opened 2025-11-05
Prevent Warp Precast for clients

Dustylemny:
Prevents overworld from being unable to find the desired region if timed correctly.


Online/Entity/OnlinePhysicalObject.cs
```diff
@@ -426,9 +426,12 @@ public void RemoveEntityFromGame(bool onlineaware = true)
 
                 if (po is Creature creature)
                 {
-                    foreach (Creature.Grasp grabbing in creature.grasps.OfType<Creature.Grasp>())
+                    if (creature.grasps != null)
```
6fears7, line 429: the only real change to this file

PR #1121, opened 2025-11-06
Arena: update Inv's eg spin timer

6fears7:
@Dustylemny My IL sucks, please confirm.

May end up reducing this to 80

### DIFF
--- a/Arena/ArenaHooks.cs
+++ b/Arena/ArenaHooks.cs
@@ -725,14 +725,13 @@ namespace RainMeadow
                 cursor.GotoNext(x => x.MatchLdarg(0));
                 cursor.GotoNext(x => x.MatchLdfld<SingularityBomb>(nameof(MoreSlugcats.SingularityBomb.counter)));
                 cursor.GotoNext(x => x.MatchLdcR4(40));
-                cursor.Remove();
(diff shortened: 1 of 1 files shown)


Arena/ArenaHooks.cs
```diff
@@ -716,6 +716,31 @@ private void SingularityBomb_ctor(On.MoreSlugcats.SingularityBomb.orig_ctor orig
...
+                cursor.GotoNext(x => x.MatchLdcR4(40));
+                cursor.Remove();
+                cursor.EmitDelegate<Func<float>>(() =>
+                {
+                    if (RainMeadow.isArenaMode(out var _))
+                    {
+                        return 120f; 
+                    }
+                    return 40f; 
+                });
```
invalidunits: Don't use remove. Use the function to transform the value instead.
```suggestion
                cursor.EmitDelegate<Func<float, float>>((float eggtimer) =>
                {
                    if (RainMeadow.isArenaMode(out var _))
                    {
                        return 120f; 
                    }
                    return eggtimer; 
                });
```

invalidunits (changes requested):

6fears7:
Thank you

PR #1125, opened 2025-11-07
Watcher: fix index out of range chunks

6fears7:
Fixes out of index error when watcher enters ripplespace


henpemaz (approved):
Heres hoping that it works.

Still, would be really nice to figure out what's the source of these different chunk lengths in rare occasions.

PR #1147, opened 2025-11-19
Arena ripplespawn camometer

Dustylemny:


6fears7 (approved):
Looks good, yes we can hook ripplespace layer  as needed

PR #1148, opened 2025-11-20
Remove duplicate hook subscription in unrelated hook method.

One-Letter-Shor:
The method is correctly subscribed at RainMeadow.MenuHooks.cs:30. Without removal, the method is being subscribed to the hook each time the user enters a online arena.

If I did something wrong do tell me, this is my first time doing this.


6fears7:
Great work

PR #1163, opened 2025-12-02
Arena:  prevent grab corpse

6fears7:


Game/RainMeadow.PlayerHooks.cs
```diff
@@ -2057,10 +2057,13 @@ private bool Player_CanIPickThisUp(On.Player.orig_CanIPickThisUp orig, Player se
...
-                        if (pl.Stunned || pl.dead)
-                        {
+                        if (pl.dead || pl.Stunned) {
+                            if (pl.dead & !arena.piggyBack)// no grabbing period
+                            {
+                                return false;                  
+                            };
                             return orig(self, obj);
-                        };
+                        }
```
invalidunits, line 2066: Use logical AND instead of bitwise AND.
```suggestion
                        if (pl.dead || pl.Stunned) {
                            if (pl.dead && !arena.piggyBack)// no grabbing period
                            {
                                return false;                  
                            };
return orig(self, obj);
                        }
```

Alternatively


```suggestion
                        if (!arena.piggyBack) return false;
```

invalidunits (dismissed):
No major blockers. I'm not sure if the bitwise operator will cause any issues in the future, but it seems to be identical to the logical and operator just a bit unclear.

PR #1164, opened 2025-12-02
Menu: adds more music options

6fears7:
image

Dynamically get song names, also adds "None" option to preserve original music

### DIFF
--- a/Game/MeadowRemixOptions.cs
+++ b/Game/MeadowRemixOptions.cs
@@ -317,9 +317,6 @@ public class RainMeadowOptions : OptionInterface
                watcherWarning = new OpLabel(introroll.pos.x + 170, 70, Translate("Watcher DLC is not activated, vanilla intro will be used instead")),
 
               new OpLabel(10, 310, Translate("IntroRoll Music")),
-
-            
-
               music = new OpComboBox2(PickedIntroMusic, new Vector2(10, 280f), 160f,  MeadowMusicRemix.ConvertSongs()) { colorEdge = Menu.MenuColorEffect.rgbWhite },
 
             };
--- a/Game/MeadowRemixOptions.cs
+++ b/Game/MeadowRemixOptions.cs
@@ -72,6 +72,34 @@ public class RainMeadowOptions : OptionInterface
     public readonly Configurable<bool> EnableAchievementsOnline;
 
     public readonly Configurable<IntroRoll> PickedIntroRoll;
+    public readonly Configurable<IntroRollMusic> PickedIntroMusic;
+
+    public enum IntroRollMusic
+    {
+        Afio,
+        Cascen,
+        Cloudlayer,
+        DustAshWrong,
+        Establish,
+        Eyto,
+        Folkada,
+        Grasp,
+        Indufor,
+        MTC,
+        Me,
+        my,
+        
+        Ones,
+
+        Porls,
+        Significance,
+        slow,
+        Soup,
+        StepsSteps,
+        Trists,
+        Woodback,
+        None
+    }
 
     public enum IntroRoll
     {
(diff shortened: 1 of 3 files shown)


Game/MeadowRemixOptions.cs
```diff
@@ -162,6 +162,8 @@ public RainMeadowOptions(global::RainMeadow.RainMeadow instance)
 
 
         PickedIntroRoll = config.Bind("PickedIntroRoll", IntroRoll.Meadow);
+        PickedIntroMusic = config.Bind("PickedIntroMusic", "Woodback"); // Happy One Year, Meadow
+
```
henpemaz: A default value is saved to the config file when no config is found, but is ignored if the config is found and changing it in the future won't have an effect in existing users.

You already had this issue happen with the arena truce timer when you changed it from ticks to seconds and changed the default, people who played before were left with 5 minutes timers until they went into the settings and changed it.

If we want to temporarily change it to woodback and then back once its done we'll need quite a bit more logic.

Game/MeadowRemixOptions.cs
```diff
@@ -308,11 +311,14 @@ public override void Initialize()
...
 
-                new OpLabel(10, 370, Translate("Introroll")),
+            new OpLabel(10, 370, Translate("Introroll")),
                introroll = new OpComboBox2(PickedIntroRoll, new Vector2(10, 340f), 160f, OpResourceSelector.GetEnumNames(null, typeof(IntroRoll)).Select(li => { li.displayName = Translate(li.displayName); return li; }).ToList()) { colorEdge = Menu.MenuColorEffect.rgbWhite },
                downpourWarning = new OpLabel(introroll.pos.x + 170, 70, Translate("Downpour DLC is not activated, vanilla intro will be used instead")),
                watcherWarning = new OpLabel(introroll.pos.x + 170, 70, Translate("Watcher DLC is not activated, vanilla intro will be used instead")),
 
+              new OpLabel(10, 310, Translate("IntroRoll Music")),
+              music = new OpComboBox2(PickedIntroMusic, new Vector2(10, 280f), 160f,  MeadowMusicRemix.ConvertSongs()) { colorEdge = Menu.MenuColorEffect.rgbWhite },
+
```
henpemaz: I think you mean "Lobby Music"? 

IntroRoll is the thing that plays before you get to the main menu, and we're not changing the classic intro song that comes with it.

henpemaz (changes requested):

Game/MeadowRemixOptions.cs
6fears7: Your second comment gave me an idea that I just tested.

Arena's truce timer could've easily be avoided had I just thought to update the string representation in the config.Bind. 

The same could be done here. My test performed the following:

1. Alter selected song choice to "None".
2. Update the string rep in code to "PickIntroRollMusic2" (Will update properly later), set a new default song
3. New default song starts up
4. Update the string rep in code back to "PickIntroRollMusic". Old value stored is re-used. No custom music played in the lobby. 


What do you think about this solution? It would allow us to only change the string entry if wanted to trigger a change. So the new config bind could be "MeadowAnniversaryYearX", so if we really wanted to change the default every year, we can guarantee it'll be picked up.

henpemaz: Sounds plausible, has the slight downside that it'll revert people's pick to your new default when you update it to the new key but it's not too bad.

In my head I had a more complex setup with a flag for "manually picked" somewhere that we'd read to detect if the player ever changed from default, then manipulate the default.

Another option is to have a "Default" string as the default, then interpret that in code to set whatever we want. Might be the cleanest actually.

henpemaz (approved):

PR #1169, opened 2025-12-03
arena: fix gorm smash

6fears7:
fixes gorm attack from sending when neither attacker nor defender simulate the event 

https://github.com/user-attachments/assets/ac6d6e9c-526e-457d-8f9b-548b8eaa4599

### DIFF
--- a/Arena/ArenaHooks.cs
+++ b/Arena/ArenaHooks.cs
@@ -865,7 +865,7 @@ namespace RainMeadow
                         if (OnlineManager.lobby != null)
                         {
                             var onlineCreature = self.abstractPhysicalObject.GetOnlineObject();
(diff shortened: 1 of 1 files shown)


Arena/ArenaHooks.cs
```diff
@@ -865,9 +865,9 @@ private static void Player_Collide2(ILContext il, MethodBase original)
                         if (OnlineManager.lobby != null)
                         {
                             var onlineCreature = self.abstractPhysicalObject.GetOnlineObject();
-                            if (onlineCreature != null && !onlineCreature.isMine)
+                            if (onlineCreature != null && !onlineCreature.isMine && source.owner.abstractPhysicalObject.GetOnlineObject()!.isMine)
```
invalidunits: Preferably to prevent any null reference errors.
```suggestion
                            if (onlineCreature != null && !onlineCreature.isMine && source.owner.isLocal())
```

invalidunits (dismissed):
Nothing major blocking.

PR #1178, opened 2025-12-12
Steam Subscriber Agreement Legal Disclaimer

Quaxly:
- Gives an acknowledgement popup your first time using Rain Meadow's Steam online features. 

image

- Enables profanity filter by default for new users.

Will leave this as a draft for now to get more input.


6fears7:
I'm still not sure how I feel about this given that Steam already has agreements you implicitly agree to with its usage of the workshop, and that Akupara hasn't given us any kind of wrist slap. 

I'll defer to @henpemaz on this one

PR #1186, opened 2025-12-18
Handle meadow mode in gate transition

6fears7:
.


Story/StoryHooks.cs
```diff
@@ -1992,9 +1993,10 @@ private void RegionGate_Update(ILContext il)
                     if (isStoryMode(out var story))
                     {
                         story.storyClientData.readyForTransition = true;
-                        return true;
+                        
                     }
-                    return false;
+                    return true;
```
6fears7, line 1998: Meadow Mode needs to return true here for the gate to operate due to the OpCode.

henpemaz (approved):

PR #1195, opened 2025-12-23
Release Before Deactivate

invalidunits:
Fixes issues with lobbies deactivating resources before properly leaving them.

[LogOutput Server.log](https://github.com/user-attachments/files/24316172/LogOutput.Server.log)
[LogOutput Client.log](https://github.com/user-attachments/files/24316173/LogOutput.Client.log)

### DIFF
--- a/GameModes/ArenaCompetitiveGameMode.cs
+++ b/GameModes/ArenaCompetitiveGameMode.cs
@@ -442,7 +442,17 @@ namespace RainMeadow
                 {
                     if (entities[i] is AbstractPhysicalObject apo && OnlinePhysicalObject.map.TryGetValue(apo, out var oe))
                     {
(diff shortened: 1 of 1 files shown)


6fears7:
[LogOutput.zip](https://github.com/user-attachments/files/24317751/LogOutput.zip)

No success. Both logs in this zip

6fears7 (commented):
We have similar code logic in LoadingHooks for Arena Next Level & Overworld gate transitions where we examine the online aware entities. Should we strip all the code there too like you've done in arena competitive gamemode or leave those?

invalidunits:
> Should we strip all the code there too like you've done in arena competitive gamemode or leave those?

Seems like I messed up on the logic here. All of the other instances are fine, I just need to surround this one with an oe.isMine if block.

PR #1198, opened 2025-12-25
arena: improve visibility of extra tabs

6fears7:
https://github.com/user-attachments/assets/96aa8b21-0461-45fb-a15e-de11939c2000

@Dustylemny thoughts?


Dustylemny:
Rainbow is a bit off, if slow flash has not yet been explored I will like to see that. But I'm cool with it

6fears7:
will make to simple flow flash

PR #1205, opened 2025-12-31
Made blocklist public for ArenaOnlineGameMode

EdEnStonne:
Talked with UO about this, seems like it won't be an issue for Rain Meadow integrity, and it'd really help for my own mod's compatibility with Meadow so... here.
One line change.


6fears7:
I will likely make all gamemodes benefit from this change before merge

EdEnStonne:
I checked the other gamemodes and it doesn't seem like they have that problem.
I could check back again when I have time, but arena was the only one with no `public` keyword

PR #1207, opened 2026-01-03
feat: Chat Quality of Life updates.

Quaxly:
Gonna use this PR for a few quality of life improvements.

- [x] Chat Opacity (Makes chat semi transparent when a player is behind it)
- [x] Chat Inactivity (Makes chat semi transparent after a short period of inactivity (no new messages and no typing))
- [x] Enforce max message length for receiving messages.
- [x] Copy/Paste support
- [x] Recently Sent Messages (Up/Down arrows)
- [x] Sound when mentioned by name in chat.
- [x] Host icon in chat
- [x] Deprecate ChatTextBox2 and use just ChatTextBox

Misc Fixes
- [x] FIX Saints can ascend each other even if they're on the same team
- [x] FIX CustomPacket incorrect data sizing.

### DIFF
--- a/Arena/ArenaHooks.cs
+++ b/Arena/ArenaHooks.cs
@@ -579,7 +579,7 @@ namespace RainMeadow
 
                     if (tb.winningTeam != -1)
                     {
-                        self.headingLabel.text = self.Translate("<TEAMNAME> WIN!").Replace("<TEAMNAME>", tb.teamNames[tb.winningTeam].ToUpper());
+                        self.headingLabel.text = self.Translate("<TEAMNAME> WIN!").Replace("<TEAMNAME>", tb.displayTeamNames[tb.winningTeam].ToUpper());
                     }
                 }
             }
@@ -1985,7 +1985,7 @@ namespace RainMeadow
                         {
                             if (OnlineManager.lobby.clientSettings[currentName].TryGetData<ArenaTeamClientSettings>(out var td))
                             {
-                                self.playerNameLabel.text += $" -- {team.teamNames[td.team].ToUpper()}";
+                                self.playerNameLabel.text += $" -- {team.displayTeamNames[td.team].ToUpper()}";
                             }
                         }
                     }
@@ -2089,7 +2089,7 @@ namespace RainMeadow
 
                     if (tb.winningTeam != -1)
                     {
-                        self.headingLabel.text = self.Translate("<TEAMNAME> WIN!").Replace("<TEAMNAME>", tb.teamNames[tb.winningTeam].ToUpper());
+                        self.headingLabel.text = self.Translate("<TEAMNAME> WIN!").Replace("<TEAMNAME>", tb.displayTeamNames[tb.winningTeam].ToUpper());
                     }
                 }
             }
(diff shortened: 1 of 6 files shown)


Arena/ArenaHooks.cs
```diff
@@ -1017,22 +1017,21 @@ private void Player_ClassMechanicsSaint1(ILContext il)
...
-                                {
-                                    opo.owner.InvokeOnceRPC(RPCs.Creature_Die, opo, saint);
-                                }
-                                else
-                                {
-                                    opo.owner.InvokeOnceRPC(RPCs.Creature_Die, opo, null);
-                                }
-
+                                // Don't ascend friends!
+                                if (po is Creature c && self.FriendlyFireSafetyCandidate(c)) return;
```
6fears7, line 1044: can also do:


```suggestion
if (ArenaHelpers.CheckSameTeam(OnlineManager.mePlayer, saint.owner, self, c)) return;

```

could also rename saint var cuz its wrongly calling target  creatures saint but ces la vie

Menu/Objects/ChatTextBox.cs
iulh, line 1: i'm so happy about this information

Online/Matchmaking/MatchmakingManager.cs
```diff
@@ -172,10 +172,19 @@ public virtual OnlinePlayer BestTransferCandidate(OnlineResource onlineResource,
...
+                RainMeadow.Error($"Error: {player} tried sending a chat message longer than what is allowed. Message will not be displayed. {message.Length}/{ChatTextBox.textLimit}");
+                return;
+            }
             ChatLogManager.LogMessage($"{player.id.GetPersonaName()}", $"{message}");
         }
         public virtual void FilterMessage(ref string message) { }
+        public virtual string FilterTeamName(string name)
+        {
+            return name;
+        }
```
invalidunits, line 187: Any reason why we can't use FilterMessage? I don't see any situation where Team name would be acceptable but a message wouldn't be and the inverse.

Online/Matchmaking/MatchmakingManager.cs
```diff
@@ -172,10 +172,19 @@ public virtual OnlinePlayer BestTransferCandidate(OnlineResource onlineResource,
 
         public virtual bool canSendChatMessages => false;
         public virtual void SendChatMessage(string message) { }
-        public virtual void RecieveChatMessage(OnlinePlayer player, string message) { 
+        public virtual void RecieveChatMessage(OnlinePlayer player, string message) {
+            if (message.Length > ChatTextBox.textLimit)
+            {
+                RainMeadow.Error($"Error: {player} tried sending a chat message longer than what is allowed. Message will not be displayed. {message.Length}/{ChatTextBox.textLimit}");
+                return;
+            }
```
invalidunits, line 180: This might fit better in `ChatLogManager.LogMessage*`

invalidunits (approved):
No major blockers.

Online/Matchmaking/MatchmakingManager.cs
Quaxly: FilterMessage passes the string by reference and uses 
k_ETextFilteringContextChat in the Steam domain while FilterTeamMessage returns a new string and uses k_ETextFilteringContextName in Steam

Online/Matchmaking/MatchmakingManager.cs
Quaxly: Doing it in MatchmakingManager just makes sure the limit only applies to messages coming in from other players and not anything originating from your own client.

PR #1211, opened 2026-01-08
arena: stop bees

6fears7:


6fears7:
- [ ] Need to add bomb to the AllowedInMode check, and the arena hooks hook is redundant so can remove

PR #1239, opened 2026-01-29
fix: manage desync entity ID

6fears7:
image

Desync entity ID leads to failed state reads

### DIFF
--- a/Online/Entity/OnlineEntity.EntityId.cs
+++ b/Online/Entity/OnlineEntity.EntityId.cs
@@ -1,4 +1,6 @@
-﻿namespace RainMeadow
+﻿using System;
+
+namespace RainMeadow
 {
     public abstract partial class OnlineEntity
     {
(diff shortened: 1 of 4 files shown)


Online/Entity/OnlineEntity.EntityId.cs
```diff
@@ -1,4 +1,6 @@
-﻿namespace RainMeadow
+﻿using System;
+
```
invalidunits, line 2: make sure to hard reset this file since nothing was dchanged.

Online/Resource/OnlineResource.Entities.cs
```diff
@@ -115,8 +115,15 @@ public void OnNewRemoteEntity(OnlineEntity.EntityDefinition entityDefinition, On
             }
             if (oe.primaryResource == this)
             {
-                RainMeadow.Error($"Already registered: " + oe);
-                return;
+                if (oe.GetType() != entityDefinition.GetType()) {
```
invalidunits: EntityDefinition and OnlineEntity will always be different types. This needs a different implementation.

invalidunits (changes requested):
Looks good so far. I'm a bit concerned about how you're using GetType though.

PR #1245, opened 2026-02-07
fix: entity resources sending states while transitioning

6fears7:
- Fixes desynced entity ID
- Fixes state reads during regional transitions for any game mode 
- This solves state reading problems with high object counts that would normally lock us up

- [x] See if I can use one coroutine definition instead of separate params for same behavior 
- [x] Meadow mode needs a world release still since I removed that from the bottom


invalidunits (dismissed):
Looks good!

invalidunits (approved):

henpemaz (approved):

PR #1246, opened 2026-02-08
fix: reset client remix settings on new lobby owner

6fears7:
- [x] Call SetClientRemixSettings
- [x] We don't need RPC everyone runs this 
- [x] Need to call ModManager.Reset() to read from disk for host

### DIFF
--- a/Online/Resource/OnlineResource.cs
+++ b/Online/Resource/OnlineResource.cs
@@ -448,6 +448,16 @@ namespace RainMeadow
             {
                 RainMeadow.Debug($"Lobby owner {player} left!!!");
                 NewOwner(MatchmakingManager.currentInstance.GetLobbyOwner());
+                if (OnlineManager.lobby != null && OnlineManager.lobby.isOwner)
+                    {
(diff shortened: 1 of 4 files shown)


Online/Resource/OnlineResource.cs
```diff
@@ -443,11 +443,20 @@ protected void SanitizeSubresources()
 
         public void OnPlayerDisconnect(OnlinePlayer player)
         {
-            //RainMeadow.Debug(this);
+            RainMeadow.Debug(this);
             if (this is Lobby lobby && owner == player) // lobby owner has left
             {
                 RainMeadow.Debug($"Lobby owner {player} left!!!");
                 NewOwner(MatchmakingManager.currentInstance.GetLobbyOwner());
+                if (OnlineManager.lobby != null && OnlineManager.lobby?.state is Lobby.LobbyState ls && !OnlineManager.lobby.isOwner)
```
6fears7: We only read from lobby state remix on mod check. Maybe we could remove that check and just let state natively handle it

PR #1252, opened 2026-02-11
Fix port claiming issues.

invalidunits:


6fears7 (approved):
I like this impl better

PR #1278, opened 2026-02-26
feat: tournament features

Quaxly:
Tournament specific features to ensure they're fair. Will add onto this mostly based on features requested from RWSE
- [ ] Tournament Validation similar to Speedrun validation.
- [x] Anti Pipecamp option


6fears7:
I think we can close this. Tournament's level of enforcement was delegated to host mod ban

PR #1282, opened 2026-03-02
fix: timeouts leading to race conditions

6fears7:
Fixes scenarios in story/arena where a timeout is reached and next world is not handled


Online/Resource/WorldSession.cs
```diff
@@ -152,15 +175,20 @@ public class WorldState : ResourceWithSubresourcesState
...
             [OnlineField]
             public int clock;
-            public WorldState() : base() { }
-            public WorldState(WorldSession resource, uint ts) : base(resource, ts)
+
+            public WorldState()
+                : base() { }
+
+            public WorldState(WorldSession resource, uint ts)
+                : base(resource, ts)
```
henpemaz, line 191: holy formatting mess

henpemaz (approved):
Alright lets try this again

Online/Resource/WorldSession.cs
6fears7: Yeah csharpier sucks, I removed my usage of it

PR #1295, opened 2026-03-08
proof of concept

iulh:
tag spawned pearls with unrecognized attribute for a future removal.

why? they're easily duped


6fears7 (changes requested):
Removing the entity from the game should be enough, we shouldn't need the unrecognized attributes field used. 

Please test and report results

PR #1296, opened 2026-03-09
Fix potential ripple level crash

CptnRedstone:
Makes sure we never try to display a ripple level above 5. Potentially fixes some watcher crashes (definitely if you externally mess with karma level).


6fears7 (approved):
The future is now

6fears7:
Changelog could be nice, not necessary. I think you can merge it yourself so merge if you're good with this

PR #1297, opened 2026-03-09
Challenge-style arena dens

CptnRedstone:
If challenge mode is available, use its more aggressive player pushout logic in arena to dissuade camping.


6fears7 (approved):

6fears7 (approved):
maybe one day we'll triple the velocity to really add some spice to arena

PR #1303, opened 2026-03-10
Fix neuron glow's color not working

CptnRedstone:
This broke at some point and looking back I'm not sure how it ever worked. We weren't applying our custom color hack to PlayerGraphics.Update() which prevented glow's LightSource from pulling the correct color; the local instance defaulted to scug class color and online players all defaulted to white. Should be fixed now for story and arena, meadow seems to handle things differently and I'm not sure it ever worked.


Game/RainMeadow.CustomizationHooks.cs
```diff
@@ -193,6 +194,28 @@ private void PlayerGraphics_DrawSprites_SlugcatCustomization(On.PlayerGraphics.o
                 PlayerGraphics.customColors = cachedCustomColors;
             }
         }
+        private void PlayerGraphics_Update_SlugcatCustomization(On.PlayerGraphics.orig_Update orig, PlayerGraphics self)
```
6fears7, line 197: I'd prefer wrapping this in a lobby == null check at the top so we can early return for orig for ease of viewing the hook. Not a blocker, just for future consideration

CptnRedstone: I'd considered that but every other place we apply the hack is written identically, I decided following form was probably better in this case. Behaviorally it's already fine, the hack appears to imply a lobby != null somewhere. I could certainly track down every time we apply the hack and add a lobby != null if wanted.

PR #1307, opened 2026-03-12
fix: grabbedBy removing list item during iteration

ButterscotchV:
Not sure this matters much as I haven't directly linked it to any problems during gameplay, but the error in the logs is really annoying me.

If performance is a concern, then the list can also be manually iterated through backwards, but it seems unnecessary considering how infrequently it gets run anyways and this way is more readable.


henpemaz (approved):

henpemaz:
Thank you for the fix!

PR #1314, opened 2026-03-13
Acquire public libs via NuGet

BarackOBusiness:
Modifies the project file to use package references to acquire BepInEx core libs, UnityEngine modules, and Newtonsoft.Json, which mirrors BepInEx plugin template behavior so only references to game specific code are manually included.
This change was primarily motivated by Newtonsoft.Json not being bundled in rain world anymore making it a bit tedious to setup the project initially. That main qualm has been solved and the bepinex and unity modules are nice extras, other public libraries the game uses are Steamworks.NET and Unity.Mathematics but I was unable to determine the versions in use by rain world to add those references, so this is fine for now.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,34 +45,16 @@ Before you will be able to make a mod, you need the following:
 
 ### Assemblies
 
-Dependencies aren't included in the repo for several reasons. You will need to add the following assemblies to the `lib/` folder:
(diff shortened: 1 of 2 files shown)


cgytrus:
I noticed this PR also adds the `nuget.samboy.dev` NuGet repository alongside the BepInEx one which I don't think is used?

BarackOBusiness:
Yeah you're right, I added that from my bepinex projects without a second thought but it only contains Cpp2IL utilities which aren't used, I'll sync my branch and remove that

PR #1319, opened 2026-03-18
feat: selectable logging level

6fears7:
- To optionally alter the experience and potentially offer performane improvements. 
- Default level is INFO.
- Most bugs now come from resolving error level alerts
- I can add trace here if you want as well to match, or can just keep it with the input key.


henpemaz (approved):
We might need a pass in the future promoting some of the Debug logs to Info, since this ships with Info level by default, but overall ok change.

Looks like there's some commits in here for other stuff as well, I'll let you merge at your discretion

PR #1324, opened 2026-03-22
undo:  rollback scav animation sync

6fears7:
Problem: Delta tries to cast for animation state, fails when there's a animation state mismatch, this creates a major  network disruption.   

```
[Error  :RainMeadow] System.InvalidCastException: Specified cast is not valid.
  at (wrapper dynamic-method) 
...

  at RainMeadow.Generics.DynamicIdentifiablesPrimaryDeltaList4+<>c__DisplayClass7_0[T,U,W,Imp].<Delta>b__0 (T e) [0x00000] in /Online/Serialization/Generics.cs:432 
  at System.Linq.Enumerable+SelectListIterator2[TSource,TResult].MoveNext ()
```

This leads back to Generics.cs:

```
        public Imp Delta(Imp baseline)
        {
            if (baseline == null) { return (Imp)this; }
            Imp delta = new();
            delta.list = list.Select(e => baseline.lookup.TryGetValue(e.ID, out var b) ? (T)e.Delta(b) : e).Where(sl => !sl.IsEmptyDelta).ToList();
            delta.removed = baseline.list.Select(e => e.ID).Where(e => !lookup.ContainsKey(e)).ToList();
            delta.BuildLookup();
            return (delta.list.Count == 0 && delta.removed.Count == 0) ? null : delta;
        }
```

this bubbles up into AbstractCreatureStateState and RoomState and can lead to major game disruption.

I don't know how to fix this at the moment, so I'm proposing removing its implementation until a solution can be found while preserving invalid's work and restoring general functionality with Scavengers in the meantime.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 - ⚠️ BREAKING: Moved all arena `arena.Killing` to reside inside of `arena.ExternalGameMode` and removed the `playerIndex` param
 ## Story
 - Fixed the "Wait for others to rescue you" death prompt blocking pause inputs.
+- Rolled back Scavenger animation sync to prevent major desync scenarios
 ### Watcher
 - Impossibly high ripple levels (6+) no longer crash the game when viewed.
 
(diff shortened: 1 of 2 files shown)


6fears7:
#1343  is used to resolve and if merged, this needs to be closed

PR #1325, opened 2026-03-22
Overseer in Meadow and Arena

invalidunits:
- [x] Move between rooms
- [x] Synchronize Emotes


Arena/ArenaOnlineGameModes/BaseGameMode.cs
```diff
@@ -647,6 +650,27 @@ List<int> suggestedDens
             }
         }
 
+        public void SpawnPlayerOverseer(ArenaOnlineGameMode arena,
+            ArenaGameSession self,
+            Room room,
+            int randomExitIndex)
+        {
+            bool spawningAvatars = RainMeadow.sSpawningAvatar;
```
6fears7, line 869: I'm a bit confused as to why this local bool is used

6fears7 (commented):
Looks fine on arena side, just want to emphasize if the overseer dies, from when I tested previously, they will exit the resource and that'll not go over well as compared to a Player character

Arena/ArenaOnlineGameModes/BaseGameMode.cs
invalidunits: I usually do this static booleans to prevent functions from interfering with each others state. In this case it has no effect but I assume it's good practice.

invalidunits:
This PR is still not ready for merging, but I'd like @henpemaz to look over it.

6fears7 (approved):

6fears7:
changelog entry would be nice @invalidunits

henpemaz (approved):

PR #1342, opened 2026-03-25
feat: add customizable eye color

6fears7:
image


henpemaz (changes requested):
This will always override the skin's default eye color with no way to get the default behavior back.

I think it should be accompanied by a checkbox next to the picker (and the field be synched as nullable perhaps), OR it should be handled similarly to the tint feature, showing an eye-tint picker and a tint-amount slider

henpemaz (approved):

PR #1347, opened 2026-03-25
fix:  PlaySounds

6fears7:
Shifted to using re-usable functions with overrides


Game/RainMeadow.GameplayHooks.cs
```diff
@@ -710,15 +710,17 @@ private void ShelterDoorOnClose(On.ShelterDoor.orig_Close orig, ShelterDoor self
...
-                                        pearl.GetOnlineObject()?.RemoveEntityFromRoom(); // ugly: since pearls are destroyed after door closing it's very noticeable when are they destroyed
-                                        coinCount++;
-                                    }
+                                    pearl.GetOnlineObject()?.RemoveEntityFromRoom(); // ugly: since pearls are destroyed after door closing it's very noticeable when are they destroyed
+                                    coinCount++;
                                 }
+                            }
                         }
                         SpecialEvents.GainedMeadowCoin(coinCount);
+                        SpecialEvents.PlayMeadowCoinSound(self.room);
```
iulh: so it plays even on zero increase, right?

6fears7: good catch

henpemaz (approved):

PR #1373, opened 2026-04-02
fix: creature kill score realtime value

6fears7:
TODO: 
- [x] Score should be set for player ONCE
- [x] Just set player score before the break

### DIFF
--- a/Game/RainMeadow.PlayerHooks.cs
+++ b/Game/RainMeadow.PlayerHooks.cs
@@ -1922,7 +1922,6 @@ public partial class RainMeadow
         RainMeadow.Debug($"%%% DIE {onlineEntity}");
         if (isArenaMode(out var arena) && self.killTag == null && arena.emptyKillTagScore > 0 && self.room.game.session is ArenaGameSession s)
         {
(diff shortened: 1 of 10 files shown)


Game/RainMeadow.PlayerHooks.cs
```diff
@@ -1919,6 +1920,24 @@ private void PlayerOnDie(On.Player.orig_Die orig, Player self)
...
+                if (onlinePlayer == null || !arena.playerNumberWithScore.TryGetValue(onlinePlayer.inLobbyId, out _))
+                {
+                    continue;
+                }
+                if (onlinePlayer == OnlineManager.mePlayer)
+                {
+                    excludeMyScoreUpdate = s.arenaSitting.players[x].playerNumber;
+                    continue;
+                }
+                onlinePlayer.InvokeOnceRPC(ArenaRPCs.UpdateAllOtherPlayerScore, excludeMyScoreUpdate, arena.playerNumberWithScore[onlinePlayer.inLobbyId] + arena.emptyKillTagScore);
```
6fears7: add check if player is host and dying to update their stats right away

PR #1378, opened 2026-04-04
fix: pearl string duping 

Quaxly:
This PR attempts to fix pearl string duplication which is notorious for making Scavenger Outposts and many rooms in metropolis completely unplayable for large lobbies.


Online/Resource/RoomSession.cs
```diff
@@ -92,6 +129,13 @@ public RoomState(RoomSession resource, uint ts) : base(resource, ts)
                     {
                         FlameJetTime = firstJet.time;
                     }
+                    // Track these on a need by need basis (Severe dupes, etc)
```
6fears7: @Quaxly  should this code remain or can it be removed?

Quaxly: That's part of the placedobjectref code and this PR isn't using it right now. Since PlacedObjectRef isn't fully implemented I'll take it out and save it for when its actually needed.

henpemaz (approved):
Happy to try it out

PR #1383, opened 2026-04-09
feat: blocklist & last played with

Quaxly:


6fears7:
I think this is an interesting concept. Up to you to keep or close

PR #1397, opened 2026-04-13
fix: Added activateTimer sync to ch 70

6fears7:


henpemaz (approved):

Online/State/RealizedOracleState.cs
```diff
@@ -8,6 +8,9 @@ public class RealizedOracleState : RealizedPhysicalObjectState
         [OnlineField(nullable = true)]
         public MoreSlugcats.STOracleBehavior.Phase phase;
 
+        [OnlineField]
+        public int activateTimer;
```
henpemaz, line 12: a `short` would overflow in 13 minutes, and this timer resets every phase, wouldn't that maybe be enough?

Passable as is, there's at most one of these going around anyways.

PR #1426, opened 2026-04-24
Fix save state duplication on clients

ButterscotchV:
Naively calling `self.currentSaveState.LoadGame()` a second doesn't actually overwrite the previous save, some lists are not cleared and retain the values from the original client save. I noticed particularly `SESSIONRECORDS`, `CONSUMEDFLOWERS`, and `imagesShown` were problems, gaining over 200,000 entries each throughout my survivor campaign with a friend. Once a save gets that bad, the game starts taking a *long* time to load (we timed 15 minutes).

The easiest way to reproduce this problem is to play a campaign, alternating between two people hosting. Each time the game loads, the host's lists will be appended to the client's list. If "Sync Save" is disabled, then it at least doesn't save the problematic file and the other person can simply re-join.

Currently marked as draft because this is currently untested, plus I don't know exactly which fields should be loaded from the client, so any feedback on that would be appreciated. Alternatively, the host's save state could simply fully overwrite the client's save state, but I think what is copied here is sensible? I tried to copy the original behaviour of the filtered host save state.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Fixed clients not being sucked into warps
 ### Saint
 - Fixed some critical errors in Saint's ending
+- Fixed duplication of some save records when loading on clients. This should fix save bloating that caused load slowdown over time.
 ## General
 - Added Global Mute toggle to Meadow Remix's "Gameplay" tab
 - Fixed most cases of pearl strings duplicating. This should noticeably decrease latency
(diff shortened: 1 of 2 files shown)


6fears7:
This has worked both in LAN multiple save slot testing and with another online player. I'll run one more test today and then if good will merge it

6fears7 (approved):

PR #1427, opened 2026-04-24
russian translation for mod's name

iulh:
hear me out, i'm not sure it's a worth PR but there's not anything i can rely on, most of the mods from my subs don't even do translations, and those who does translate languages they're familiar with i believe.
you can check it yourself with `grep -r --include="string.txt" -- "-name|" *` (it's 4 somehow translated mods out of 53 for me)

now to the point: it seems weird to me that Russian lacks mod's name translation, so i'd like to bring up the question since it's a mystery for me. if you know the reason please let me know


phanerka:
Mod name translation wouldn't make a lot of sense. 
"Meadow" can't be translated on its own (since it's a reference to another game), and translating "Rain" alone would be weird. 

That was something that was internally briefly discussed before mod release

iulh:
> Mod name translation wouldn't make a lot of sense. "Meadow" can't be translated on its own (since it's a reference to another game), and translating "Rain" alone would be weird.
> 
> That was something that was internally briefly discussed before mod release

ty, the answer works. but why do other languages have it then if it's a reference to the game in a context

6fears7:
The answer really is just, "we didn't notice they did that."

PR #1439, opened 2026-04-28
Directly set 'didParry' to false

6fears7:
Don't defer and potentially leave protection on the player. Just set to false.


6fears7:
Closing in favor of parry fix PR: https://github.com/henpemaz/Rain-Meadow/pull/1441

PR #1445, opened 2026-05-03
Ghost Spear fix

EdEnStonne:


EdEnStonne:
So this wasn't supposed to happen, sorry

PR #1446, opened 2026-05-03
Ghost Spear Fix (hopefully)

EdEnStonne:
So, it's been 2 hours and a half, I have no clue why I even though about trying to fix that but here I am.
I've been recommended to not set `spear.addPoles` with an RPC, and since I can do quite limited testing on my own (I don't realy know how to compile Meadow), I've put every modification I did for my testings BUT that.
It shouldn't cause issues, the main issue with ghost spear is that `spear.addPoles` was set to true every frame of it being embedded, while it needed to be set only once. As long as `spear.addPoles` is not set twice or more, there shouldn't be ghost spear issues.
I've put an RPC to remove the pole added by the spear anyway. It might be actually useless in practice when the above bug is remove, but we never know. Should be 100% safe even if the spear is long gone by the time the RPC reach the client, but of course it's Meadow, we never know.

So here is my code that I couldn't test fully but could test enough to actually make a PR about it. Enjoy !

### DIFF
--- a/Game/RainMeadow.GameplayHooks.cs
+++ b/Game/RainMeadow.GameplayHooks.cs
@@ -59,6 +59,7 @@ namespace RainMeadow
             IL.VultureGrub.InitiateSignal += PhysicalObject_Trigger;
 
             On.Spear.Spear_makeNeedle += Spear_makeNeedle;
+            On.Spear.resetHorizontalBeamState += Spear_resetHorizontalBeamState;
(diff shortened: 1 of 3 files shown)


EdEnStonne:
Seems like InvalidUnit got a better solution to it :
https://github.com/henpemaz/Rain-Meadow/pull/1447

Feel free to close this PR

PR #1452, opened 2026-05-05
Remove sSpawningAvatar

invalidunits:
Needs extensive testing, don't merge for now.

### DIFF
--- a/Arena/ArenaHooks.cs
+++ b/Arena/ArenaHooks.cs
@@ -469,6 +469,8 @@ namespace RainMeadow
                 return;
 
             var room = self.room;
+
+            RainMeadow.sSpawningAvatar = true;
             AbstractPhysicalObject apo = new(
                 room.world,
                 Watcher.WatcherEnums.AbstractObjectType.RippleSpawn,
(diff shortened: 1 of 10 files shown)


6fears7:
Putting this here as a reminder: If this gets added, add a chglog so modders are aware how to benefit from the changes

Arena/ArenaOnlineGameModes/BaseGameMode.cs
```diff
@@ -498,10 +497,9 @@ CreatureTemplate.Type templateType
             abstractCreature.pos.abstractNode = room.ShortcutLeadingToNode(
                 randomExitIndex
             ).destNode;
+            OnlinePhysicalObject.RegisterPhysicalObject(abstractCreature, OnlinePhysicalObject.RegisterFlags.Avatar);
             abstractCreature.Room.AddEntity(abstractCreature);
             RainMeadow.Debug("assigned ac, registering");
-            self.game.world.GetResource().ApoEnteringWorld(abstractCreature);
```
6fears7, line 503: I think we need this line back

Arena/ArenaHooks.cs
```diff
@@ -487,6 +488,7 @@ public void Player_SpawnDynamicWarpPoint(
             };
```
6fears7, line 488: we need to ensure the amoeba is not transferable. Im not sure if the below will work but you get the drift: 

```suggestion
OnlinePhysicalObject.RegisterPhysicalObject(apo as AbstractCreature, OnlinePhysicalObject.RegisterFlags.Avatar);
```

invalidunits:
The refactor is nice, but there's no general need for this.

PR #1455, opened 2026-05-08
Pr/1441 plus gourmand arti spear parry fix

EdEnStonne:
Ok so the Evil RPC gods stops me from receiving `Weapon_CreatureDeflect`.
I still don't know why. 
It should work. 
There's no reason it shouldn't work.

This is supposed to be a fix for Gourmand and Artificer's parry being abit wonky, with the help of the new parry fix

> I'm not sure of what I'm doing, but ima try to lay it out :
>1. At the end of `spear.HitSomething`, if the weapon doesn't met any embedding criteria, the weapon will be set to `Mode.Free` with an opposite velocity and a random rotation, in addition to the classic "you hit the lizard on the head" sound. It's notably used by gourmand, who's `creature.SpearStick()` is set to false if Gourmand is rolling or sliding.
>  - I think marking the spear with the `"parry"` lock and act out the bounce off with an RPC, similar to the parry fix.
>2. Artificer's parry grabs all the weapons in a 300 unit range and... also put the weapon on `Mode.Free` and gives it a new velocity. Like cmon guys THERE'S LITERALLY A FUNCTION THAT DOES EXACTLY THAT IN WEAPON CLAS-
>  - Same here, lock it as `"parry"` and give it the weapon states.
>  - Also probably putting an option to tune the range of artificer's parry cause WHAT THE HELL DO YOU MEAN 15 TILES ???


EdEnStonne:
So I forgot to push the version where I did some initial fixes.
Should be good now, hopefully, needs testing

EdEnStonne:
Ok so, after abit of testing, it works wonderfully well.
I'm so glad I got that long lasting issue fixed (I can finally parry as Artificer now :monksillius:).
Will leave that fix for review

6fears7:
I'd recommend merging from main to clean up any changes for the diff. 

Few other things:
- Do we have room in the MSC Settings for the arti parry distance text field? A screenshot would be nice.
- When watching the tournament, we noticed what looks like locally simulated parries but outdated spear states where spears continue to travel after the parry. I believe the spear owners saw things correctly, but we (via the overseer) did not. It may be good to test the arti parry RPC by getting the OnlineCreature and calling BroadcastRPCInRoom, or if the OC is not available, iterating over room participants.
- Update the CHANGELOG.md with your changes

EdEnStonne:
Hello UO !

- Yes ! There is, since the vanilla Artificer's parry range is plainly ridiculous. The default is set to an arbitrary 150 (half the vanilla range) and I'm pretty sure it needs abit more testing from some Artificer mains to find the sweet spot. 
- (Also wait you can include screenshots here ? Neat. Here's teh screenshot, I only have MSC enabled in there.)
image

- ...I'm not sure of what you're describing to me here, but in any case, this is not a bad idea, will probably test this out. Though I want to note that anotehr bug that I found with the parry fix (and had fixed here by accident) is that the spear position was set *before* the mode was changed, which is not an issue unless the spear already hit the player in the other client before the parry RPC came. In that case, the spear would... *fall* from the hit player instead for getting flinged away. If that was the bug mentionned, the fix was simply setting the state before the pos.
- Woops, didn't know that existed, will do !

Menu/Components/OnlineSlugcatAbilitiesInterface.cs
```diff
@@ -147,10 +147,26 @@ public MSCSettingsPage(Menu.Menu menu, MenuObject owner, Vector2 spacing, string
...
-                blockArtiStunCheckBox = new(menu, this, this, positioner - spacing, textSpacing, Translate("Disable Artificer Stun:"), DISABLEARTISTUN, false, Translate("Prevent Artificer from stunning other players"));
-                sainotCheckBox = new(menu, this, this, positioner - spacing * 3, textSpacing, Translate("Sain't:"), SAINOT, false, Translate("Disable Saint ascendance ability, but allow it to throw spears"));
-                saintAscendDurationTimerTextBox = new(new Configurable<int>(RainMeadow.rainMeadowOptions.ArenaSaintAscendanceTimer.Value), positioner - spacing * 4 + new Vector2(-7.5f, 0), 40)
+                artiParryDistanceTextBox = new(new Configurable<int>(RainMeadow.rainMeadowOptions.ArtificerParryDistance.Value), positioner - spacing * 1 + new Vector2(-7.5f, 0), 40)
+                {
+                    alignment = FLabelAlignment.Center,
+                    description = Translate("How many unit Artificer can parry from (300 units in MSC). 20 units is a tile. Default : 150."),
```
henpemaz: "How far Artificer can parry from. Vanilla default: 300. Online default: 150."

EdEnStonne: Will change that. The default value will also be lowered to 80-100

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# Release 1.15.0
+
+## Arena
+- Made Artificer and Gourmand's parry defencer side.
```
6fears7: Grammar: "defender", not defencer. No slugcats are taking away ANYONES fences

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
...
+- Added a MSC setting to change Artificer's parry range.
```
6fears7: "Added configurable setting to change..."

Game/RainMeadow.GameplayHooks.cs
```diff
@@ -68,13 +68,112 @@ public void GameplayHooks()
             IL.Player.TerrainImpact += Player_TerrainImpact;
             On.Weapon.HitThisObject += Weapon_HitThisObject;
             On.Weapon.Thrown += Weapon_Thrown;
+            IL.Spear.HitSomething += Spear_HitSomething;
+            IL.Player.ClassMechanicsArtificer += Player_ArtificerParry;
```
6fears7: Move to PlayerHooks.cs please

Online/State/RealizedWeaponState.cs
```diff
@@ -30,10 +30,10 @@ public RealizedWeaponState(OnlinePhysicalObject onlineEntity) : base(onlineEntit
 
         public override void ReadTo(OnlineEntity onlineEntity)
         {
-            base.ReadTo(onlineEntity);
```
6fears7, line 33: Was this moved due to manage the deflection state update?

6fears7 (changes requested):

Online/State/RealizedWeaponState.cs
EdEnStonne: Yep, the position being set before the Weapon Mode made the spear just drop at the defender's feet instead of getting deflected normally

Arena/ArenaHooks.cs
```diff
@@ -184,6 +184,163 @@ public void ArenaHooks()
...
+            IL.Player.ClassMechanicsArtificer += Player_ArtificerParryNStunRange;
+        }
+        
+        // Restrict Artificer's parry and stun range in arena
+        private const float VANILLA_ARTI_PARRY_RANGE = 300f;
+        private const float VANILLA_ARTI_STUN_RANGE = 200f;
+        private const float VANILLA_ARTI_STUNTHROUGH_RANGE = 60f;
+        private const float ARTI_PARRY_MAX_COOLDOWN = 40f;
+        private const int ARTI_PARRY_EXTRA_FRAMES = 5;
+        private bool IsPlayerLocal(Player player) =>
```
invalidunits: Use `player.IsLocal()` instead

Game/MeadowRemixOptions.cs
```diff
@@ -42,8 +42,10 @@ public class RainMeadowOptions : OptionInterface
     public readonly Configurable<bool> FriendlyFire;
 
 
+    public readonly Configurable<int> ArtificerParryDistance;
+    public readonly Configurable<int> ArtificerStunDistance;
```
invalidunits: I'd prefer these all be multipliers.

Arena/ArenaHooks.cs
```diff
@@ -184,6 +184,163 @@ public void ArenaHooks()
...
+        private const int ARTI_PARRY_EXTRA_FRAMES = 5;
+        private bool IsPlayerLocal(Player player) =>  
+            player.abstractCreature?.GetOnlineCreature() is not OnlineCreature onlineCreature 
+            || onlineCreature.isMine;
+        private bool IsArtiInExtraParryFrames(Player player) => isArenaMode(out var arenaOnline)
+            && IsPlayerLocal(player)
+            && arenaOnline.artiParryDistance > 0
+            && player.pyroParryCooldown < ARTI_PARRY_MAX_COOLDOWN 
+            && player.pyroParryCooldown > (ARTI_PARRY_MAX_COOLDOWN - ARTI_PARRY_EXTRA_FRAMES);
+        private void Player_ArtificerParryNStunRange(ILContext il)
```
invalidunits: Hook function names should include the function they are hooking

Arena/ArenaHooks.cs
```diff
@@ -184,6 +184,163 @@ public void ArenaHooks()
...
+                    x => x.MatchStloc(11)
+                )) { throw new KeyNotFoundException("Couldn't find the key 1 of IL hook"); }
+                
+                
+                toParryLoop = cursor.MarkLabel();
+                cursor.GotoNext(moveType: MoveType.After, x => x.MatchStloc(11));
+
+                cursor.Emit(OpCodes.Ldarg_0);
+                cursor.EmitDelegate(IsArtiInExtraParryFrames);
+                cursor.Emit(OpCodes.Brfalse, cursor.Next);
```
invalidunits, line 221: Run the next instruction if true, otherwise run the next instruction...?  This seems unnecessary to me

6fears7 (dismissed):

EdEnStonne:
I might need alittle more time to push another update from player's feedback.

The last feedback I got was :
- lower leniency, 5 frames might be too much (the player saying that seeemed WAY more skilled at parrying than the other, so I'm taking this comment with a grain of salt)
- add some audio cue to parrying (surpirsed tehre isn't in vanilla, but I do agree with this statement, pings makes it hard to see if the spear was parried or something weird happened)

I got feedback from a vaste range of player, some being WAY more consistent at parrying than others.
I might also add another checkbox to make the leniency of parry toggleable, as some people seems to prefer the vanilla Artificer's parry.
In the end, it's mostly about player's taste and colors : I do think that parry being short-range but lenient is the most fair option, but in the case that alot of people disagree, I'll give the option to opt-out.

In terms of bugs and testing, everything works well ! 
Gourmand got barely any visible changes from this, except the spear bouncing off instead of staying impaled inside em.
Artificer's parry is completely usable and you can train yourself to parry consistently without worrying about ping ! It's really hard to use mid-match, but it always has been like that in local : one doesn't just parry like that.

6fears7:
In my opinion, you should set what you think is the best default value for range and leniency. We can include it, and if we see player feedback after release requesting changes, we can revisit. We don't have a lot of room left in the arena UI for more toggleable features and I'm hesitant to add more tabs since we already have like 4. 

We may want to consider inside the MSC settings another nested tab for each character. But that can be revisited another time.

invalidunits (dismissed):
Looks good! No major issues.

Arena/ArenaHooks.cs
EdEnStonne: Oh woops yeah will do

Arena/ArenaHooks.cs
EdEnStonne: Oh, not at all ! The cursor.Next points to the next instruction, and then I add instruction in between.
Since I skip a part of the code if Artificer is in the extra leniency frames, which includes the initialization of a local variable, I have to initialize it back to allow the rest of the code to work properly.
If I'm not in those extra leniency frame, I skip this redunding code.

Here's what the IL looks like at this place spefically :

image

Game/MeadowRemixOptions.cs
EdEnStonne: Could do, except for the fact that I've never seen float in meadow settings.
Were you thinking of it in pourcent ? (0-999% converts to 0 to 9.99 multiplier)
Or do you have an idea on how to add float Text boxs ?

EdEnStonne: Also, I feel like multipliers are less appealing than the pixel radius, which seems easier to visualize than a pourcent of some kind of base range.

6fears7: most users won't understand a hard set pixel radius. They haven't seen your annotated screenshot. A multiplier makes it a little easier I think to visualize the radius size.  Except nobody knows arti's range is insanely busted atm

Here's what we do for ripple level:

```cs
        public float SetRippleLevel(Func<Player, float> orig, Player self)
        {
            if (isArenaMode(out var arena))
            {
                return (arena.watcherRippleLevel - 1) * 0.5f + 1; // take the configurable int and convert to float
            }
            return orig(self);
        }

```

but I do see how we'd still need a configurable float. Let me look.

6fears7: `textbox.accept = OpTextBox.Accept.Float`

EdEnStonne: Oh neat !
Will see how I can arrange that.

6fears7 (dismissed):

invalidunits (dismissed):

EdEnStonne:
At the request of UO, I've merged [Drown](https://github.com/henpemaz/Rain-Meadow/pull/1448) on this branch to set up `ExportLocalSettings` correctly. 
So this branch depends on Drown now ig.

EdEnStonne:
So I fixed that tiny bug which made gourmand parry twice.
It feels really clean now, Glad it works.
Though, I did add a way to add a lock manually to an OnlineEntity (with a dummy RPC attached to it), so @invalidunits I hope you don't mind that.

Online/Entity/OnlineEntity.cs
```diff
@@ -523,6 +523,18 @@ public void Lock(string key, RPCEvent @event)
...
+            RainMeadow.Debug($"{this}: Locked {key} manually");
+            if (!locks.ContainsKey(key))
+            {
+                locks.Add(key, [ new() ]);
+            }
+            else
+            {
+                locks[key].Add(new());
+            }
+        }
```
EdEnStonne, line 537: Here is the thing I added. The only way to remove this lock realistically is `ClearLock`, and I use it to see if a spear was already deflected.
It gets removed when the spear is not in free mode anymore, so it won't linger around for the next throws.

6fears7 (dismissed):

Online/Entity/OnlineEntity.cs
invalidunits: Looks good to me. As long as clearlock is actually called this should be fine.

invalidunits (dismissed):

henpemaz:
I'll review this once the diff is readable, hopefully it's just a matter of merging Drown to main

henpemaz (approved):

invalidunits (approved):

PR #1458, opened 2026-05-14
Ext enum desync fix

EdEnStonne:
This is a LONG beef of mine.
The concept is pretty simple :
- allows some ExtEnum to be mapped in a separate place by instanting an `CompressedExtEnumBase` in `MeadowExtEnumSync.SyncedExtEnumList`
- the synced ExtEnum will be send by the host as a compressed string. Compression/Decompression method is set by CompressedExtEnumBase
- every instance of ExtEnum.Index has been replaced to use that mapping if the ExtEnum is synced.

I had to add something else to make it work :
- Dictionaries are considered non-nullable when caching the serialization method from the RPC registery.


EdEnStonne:
- ExtEnum are now mapped in alphabettical order to minimise order mismatch

EdEnStonne:
I'm getting abit paranoid.
Putting it back to draft till I did 1 more bajilion tests on it

EdEnStonne:
So the tests were absolutely necessary, but the desync by enum order is finally. fixed.
I did added `SerializeNullableShort<T>`, and it might need to be reviewed as well as the dictionnary issue since I'm not the best as serialization.
Theoretically, this *could* allow Jolly Coop to not be sync required anymore (that's even how I tested desync), but I feel like it'd still be a bad idea to put something that adds enums as non-sync required. Your call.

Anyway, this PR is ready.

EdEnStonne:
There's one optionnal improvement this could have, and it's to handle Enums with more than 256 entries.
But that'd be a hell of a job and I'm alrady happy that this bug is done for.

Online/ExtEnumSync.cs
```diff
@@ -0,0 +1,864 @@
...
+                    // Uh have fun ig ? Not much to do there
+                    // LogTestCompression();
+                }
+                else
+                {
+                    OnlineManager.lobby.owner?.InvokeRPC(RequestCompressedExtEnums);
+                }
+            }
+        }
+    }
```
invalidunits: Since this isn't a hook. It should be named `OnLobbyJoined`

Online/ExtEnumSync.cs
```diff
@@ -0,0 +1,864 @@
...
+            (x,y) => x.Split(separator).Last() == y.Split(separator).Last().Substring(1));
+    }
+}
+
+public static class MeadowExtEnumSync
+{
+    // --------------------- Hooks
+    internal static void ApplyHooks()
+    {
+        MatchmakingManager.OnLobbyJoined += MatchmakingManager_BTWVersionChecker_OnLobbyJoined;
```
invalidunits: From what I understand this is the origin of all ext-enum synchronization. After the lobby is available to us.

However we can't safely modify state enums after joining the lobby. Since it might invalidate our lobby state.

You should completely synchronize ext enums before lobby is available is requested and made available by `Lobby.RequestLobby`

Extensions.cs
```diff
@@ -339,137 +339,160 @@ public static void TryMutualBind(this Menu.Menu? menu, MenuObject? first, MenuOb
...
         }
-        //I've been told these could go in Menu/MenuHelpers.cs, but separating them from the functions they're based off of seems wrong. If someone else wants to though, go for it.
-        ///<summary>Effectively deletes a UI element's directional keybind(s) by binding the button to itself.</summary>
-        public static void TryDeleteBind(MenuObject menuObject, bool left = false, bool right = false, bool top = false, bool bottom = false)
-        {
-            menuObject.nextSelectable[0] = (left ? menuObject : menuObject.nextSelectable[0]);
-            menuObject.nextSelectable[1] = (top ? menuObject : menuObject.nextSelectable[1]);
-            menuObject.nextSelectable[2] = (right ? menuObject : menuObject.nextSelectable[2]);
-            menuObject.nextSelectable[3] = (bottom ? menuObject : menuObject.nextSelectable[3]);
+        //I've been told these could go in Menu/MenuHelpers.cs, but separating them from the functions they're based off of seems wrong. If someone else wants to though, go for it.
```
invalidunits, line 342: There right, no need to hide this here.

invalidunits (changes requested):
Looks good, here's a few of my thoughts!

Online/ExtEnumSync.cs
EdEnStonne: Woops, completely forgot to change that name from BTW.
Will do

Online/ExtEnumSync.cs
EdEnStonne: If the state isn't correctly synced, meadow will error on the type cast and ask the host to send the data again, which from my testing didn't cause any significant issues.
Nevertheless, your approach is way cleaner, so I'll attempt to do that.

Extensions.cs
EdEnStonne: ...I don't remember adding those lines..?
Hold on

EdEnStonne: Those aren't my changes. I could move them, but I have no clue what this function is

Online/ExtEnumSync.cs
EdEnStonne: Well that hook was deleted in the end since I've completely changed the way to trigger the sync ping pong

Online/ExtEnumSync.cs
EdEnStonne: The enum sync check has been put before the lobby request. It seems to work flawlessly.
Might need some review on error handling, but this *should* be fine, there's close to no reason the compression algorithm should error.

invalidunits (dismissed):

EdEnStonne:
So I've debated myself long enough and since Jolly Coop worked perfectly fine before it was put as high impact (except the enum desync thingy), there literally 0 reason to not put it back as non-high impact except underconfidence.
Plus it should help with extensive testing sooo...

Yep, welcome back Jolly

Online/Entity/OnlineEntity.Data.cs
```diff
@@ -54,7 +54,7 @@ public EntityData() { }
             [DeltaSupport(level = StateHandler.DeltaSupport.NullableDelta)]
             public abstract class EntityDataState : OnlineState, IIdentifiable<byte>
             {
-                public byte ID => (byte)handler.stateType.index;
+                public byte ID => handler.StateTypeIndex;
```
henpemaz, line 57: This is a hot-path in the netcode, it **can't** be a lookup, needs caching, probably at the handler level.

Online/State/OnlineState.cs
```diff
@@ -267,6 +267,13 @@ public class StateHandler
             public Func<OnlineState, OnlineState, OnlineState> delta;
             public Func<OnlineState, OnlineState, OnlineState> applydelta;
             public int ngroups;
+            public byte StateTypeIndex
+            {
+                get
+                {
+                    return this.stateType.MeadowIndex();
```
henpemaz: cache this, and figure out if the cache needs invalidating at some point.

henpemaz (changes requested):
It's a nice to have but if it runs heavy I'd rather leave it out

Online/State/OnlineState.cs
EdEnStonne: Will do. There shouldn't be any change in that number when in the same lobby

Online/Entity/OnlineEntity.Data.cs
EdEnStonne: Oh yeah- didn't realize it. I'll probably mimic how index are cached in rain world.
Is there anywhere else where I should set up a cache system ? I don't know all the hot-path, and don't want to cause any heavy lag due to enum mapping.

Also, I haven't noticed some heavy (oblivious) perfomance loss in the testing, so that's good news I'd guess

Online/State/OnlineState.cs
EdEnStonne: Should be all good now

henpemaz (approved):

PR #1461, opened 2026-05-18
Remove OnlinePlayer directly from resource.

invalidunits:
Adds a new function `Discharge` which removes a player from a resource with a single RPC


invalidunits:
This is functional with vanilla, but I need to figure out better watcher warping and the MSC warps before I merge this.

6fears7 (approved):

6fears7:
Warping has been improved due to some bug fixes found during 1.15. Keep or close?

invalidunits:
Succeeeded by #1635

PR #1467, opened 2026-05-23
Fix slug on back null error

invalidunits:


henpemaz (approved):

6fears7:
This is already in main, closing.

PR #1470, opened 2026-05-25
Don't injure creatures in Meadow mode

Ombekende:
oaoa im not super familiar w github so apologies if something is incorrect ^^;

Returns Creature.Violence and Lizard.Violence in Meadow mode lobbies, just like in Creature.Die
Otherwise players can be injured, for example by explosions, but not _die_, which leaves them unable to move until they lobby.


invalidunits (dismissed):
Looks good, I'll leave the merging to @henpemaz

henpemaz (dismissed):

henpemaz:
Thank you for the fix. I'm just wondering though, if there's a "normal" way of running into explosions that isn't oneself cheating in bombs? @Ombekende

6fears7:
@henpemaz Inv timeline in Meadow Mode will spawn explosive snails

Ombekende:
@henpemaz as far as i know no, but its very easy to hook Player.Pyrodeath() in other mods to explode slugcats (which will leave the scug unharmed but injure other creatures)
thats how i found out about the issue ^^; we blew up poor Foo a bunch of times

PR #1473, opened 2026-05-28
fix specator menu go outside of the screen

iulh:
so previously it behaved like this on "actual size" 1920x1080. you see it's unusable..
RainWorld_926kqCsZvN
this is an actual cause of the PR :3

previously i had no clue why is it like that, but now i have a theory the page is centered with "fixed size" of original game (1366x760). i have no guaranteed proof of it, but here are images from my duck taping:

1. "bright zone" is 1366x760 vs "actual size" 1600x900, `pos.x` is 0
image

2. "bright zone" is 1366x760 vs "actual size" 1920x1080, `ScreenWidth` is fixed 1366
2-wawa

all the two examples show that the zone might be centered with "fixed size" of 1366x760

fuck formatting

### DIFF
--- a/OnlineUIComponents/SpectatorOverlay.cs
+++ b/OnlineUIComponents/SpectatorOverlay.cs
@@ -12,67 +12,67 @@ namespace RainMeadow
         public static float ButtonSize => 30;
         public static Vector2 MenuShift => new(-186f, 171f);
         private Vector2 ScreenSize => manager.rainWorld.options.ScreenSize;
-        public List<PlayerButton> PlayerButtons => playerScroller.GetSpecificButtons<PlayerButton>();
+        // public List<PlayerButton> PlayerButtons => playerScroller.GetSpecificButtons<PlayerButton>();
         public SpectatorOverlay(ProcessManager manager, RainWorldGame game, RoomCamera camera) : base(manager, RainMeadow.Ext_ProcessID.SpectatorMode)
         {
             this.game = game;
             this.camera = camera;
             selectedObject = null;
 
-            playerScroller = null!;
-        }
-        private bool UpdateList()
-        {
-            List<PlayerButton> playerButtons = PlayerButtons;
-            List<OnlinePlayer> newPlayers = [.. OnlineManager.players.OrderBy(onlineP => onlineP.isMe ? 0 : 1)]; // will keep this logic for LAN
-            List<OnlineCreature> realizedPlayers = [.. OnlineManager.lobby.playerAvatars.Select(kv => kv.Value.FindEntity(true)).OfType<OnlineCreature>().OrderBy(opo => opo.isMine ? 0 : 1)];
-            if (newPlayers.Count == playerButtons.Count) return false; // race condition that will Never Happen(TM)
-
-            for (int i = playerButtons.Count - 1; i >= 0; i--)
-            {
-                PlayerButton button = playerButtons[i];
-                if (newPlayers.Contains(button.player))
-                {
-                    newPlayers.Remove(button.player);
-                    button.avatars = realizedPlayers.Where(x => x.owner == button.player).ToList();
-                    continue;
-                }
-                playerScroller.RemoveButton(button, false);
-            }
-            foreach (OnlinePlayer player in newPlayers)
-            {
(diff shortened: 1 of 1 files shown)


iulh:
maybe it'd be more nicer to take the size from the page instead of rain world's settings.
yet it'd look ugly still, but correct

wait i might find something
image

iulh:
all i've done is duplicated change of `pos` to `Update` loop.
i didn't touch the Y axis also, so it's not centered. easy to fix

the changes are have to be reviewed, since i'm not satisfied with some of my decisions and i need another perspective

## 1920x1080
image

## 1024x768
image

OnlineUIComponents/SpectatorOverlay.cs
```diff
@@ -87,7 +97,8 @@ public void SpectatePlayer(AbstractCreature? aC)
         public AbstractCreature? spectatee;
         public RainWorldGame game;
         public RoomCamera camera;
-        public ButtonScroller playerScroller;
+        public MenuLabel label;
+        public ButtonScroller scroller; // previously "playerScroller"
```
iulh: i changed this field's name, i'm aware of it being used anywhere else. please take in mind

OnlineUIComponents/SpectatorOverlay.cs
```diff
@@ -66,6 +71,11 @@ public override void Update()
                 selectedObject = null;
 
             }
+
+            var page = pages[0];
+            var pos = new Vector2(ScreenWidth, 0) + MenuShift - page.pos;
+            label.pos = pos;
+            scroller.pos = new(pos.x, pos.y - 38 - ButtonScroller.CalculateHeightBasedOnAmtOfButtons(MaxVisibleOnList, ButtonSize, ButtonSpacingOffset)); // ugly: repeating the code :3
```
iulh: also i'd like to highlight this part, since it's a duplicate of another line of code and runs every frame which might be changed in time?

OnlineUIComponents/SpectatorOverlay.cs
```diff
@@ -56,6 +47,21 @@ private bool UpdateList()
             scroller.ConstrainScroll();
             return true;
         }
+        public override void Init()
```
iulh: turns out you should NOT start building a UI hierarchy inside `ctor`, instead do it after `Init` method is complete. 
in fact it removes the flickering effect and unnecessary calculation on every frame i tried while prototyping.

i'm glad i decided to try this approach instead, and it's probably can be applied to other interfaces for Rain Meadow.

OnlineUIComponents/SpectatorOverlay.cs
```diff
@@ -49,11 +49,11 @@ private bool UpdateList()
         }
         public override void Init()
         {
-            base.Init();
-
             Page page = new(this, null, "spectator", 0);
             pages.Add(page);
 
+            base.Init();
```
iulh: oh my fucking god, it has to be called after the page is added. i believe i just didn't look for it in logs, since it was checked on Update if not initialized

OnlineUIComponents/SpectatorOverlay.cs
iulh: don't listen to me. it's complicated and only 3 classes use `Init` in general

OnlineUIComponents/SpectatorOverlay.cs
```diff
@@ -12,67 +12,67 @@ public class SpectatorOverlay : Menu.Menu
...
-        public override void Init()
-        {
             Page page = new(this, null, "spectator", 0);
             pages.Add(page);
 
-            base.Init();
-
             var pos = new Vector2(ScreenSize.x, ScreenSize.y / 2) + MenuShift - page.pos;
 
             MenuLabel label = new(this, page, Translate("PLAYERS"), pos, new(110, 30), true);
```
iulh: commiting this just to show the overlay doesn't flicker without `ButtonScroller`. Dusty just mentioned Remix elements caused issues before and this overlay is the simplest thing i could test the theory on

yet there's also a shift with using `ButtonScroller`

iulh:
my theory is the `Init` method was implement after Downpour, since their elements started to take way more time so race conditions started happening. and the check for initialized page and `Init` calling on `Update` hold the process nicely

6fears7:
Looks good, have you tested this with other players in your game?

iulh:
> Looks good, have you tested this with other players in your game?

not really, but it's tested on compiled build by myself on different lobbies. they were modded so hard to distinguish the causes. one of the bugs i've found was [about the order where i add the page too late](https://github.com/henpemaz/Rain-Meadow/pull/1473#discussion_r3327381479), nothing than that i saw no issues if this works

PR #1479, opened 2026-06-10
Fixed Slugslot pseudo-random generator

EdEnStonne:
I didn't test it yet (my rw is... not available rn) but I mean it's 2 lines, there's no way it just breaks for no reason. 
Here, pseudo-random generator as a static variable, so slugslot isn't rigged anymore


EdEnStonne:
Woops changelog hold on

EdEnStonne:
Tested with no issues

EdEnStonne:
So someone else also tested it and they got it first try somehow LMAO.
Pretty sure it works 👍

PR #1480, opened 2026-06-10
Allow players to spectate their own slugcats when dead

wonky2:
I've seen people supporting the idea of spectating dead slugcats, and I've found that as long as the slugcat of which you're the owner exists, you can always spectate it, even if it's dead and in an unloaded room. The same doesn't happen with others' slugcats; if they're dead and not in a loaded room, you can't spectate them even if they're not destroyed yet.

Technically, it should be fine to completely remove `GetSpectableAvatars()` and allow spectating of all slugcats, and things will still work, but the inconsistency of who can be spectated and who can't (due to rooms getting loaded and unloaded) may be too confusing.

Anyway, I think spectating one's own dead slugcat should be allowed for two reasons:
- we already spectate it by default when we die until we switch to someone else
- it's already possible in jolly co-op, so people might be more used to having that option


6fears7 (dismissed):

6fears7:
@wonky2 please resolve the changelog merge issue

@invalidunits any thoughts or you cool with this?

invalidunits (approved):

PR #1481, opened 2026-06-12
feat: cheats setting and dev tools improvements

Quaxly:
- Adds the `Cheats` setting to lobby creation which allows the host to allow or disallow other players using Dev Tools cheats.
   - `OnlineManager.CheatsAllowed` will return true if you're offline, if cheats are allowed in your lobby or if you're the host. Otherwise they'll be disabled. This property can also be read by other mods. (Will provide a reflection method to do this safely and easily in 1 line without the need to use Rain Meadow as a dependency)
   - Players will be able to open dev tools but won't be able to use most functionality or open the Dev Tools UI.
   - Disables Dev Tool functionality within every UpdateableAndDeletable.
   - "Q" to unload rooms and a few visual debug tools will still work.
- Completely disables restarting the cycle using Dev Tools when online which would cause major desyncs.
- Fixed the dev tools numpad keys trying to teleport everyone.


henpemaz (dismissed):
Good QoL

henpemaz (approved):

PR #1482, opened 2026-06-14
fixed watcher warp not working when warping from a world for the second time in the cycle

r3nsen:
So, if you warp out and back into any world, the second time you try to warp will fail

In watcher campaign, Overworld_Loaded_WaitLoop is synchronous because WaitAndExecuteSession doesnt hit the yield, so in the end, worldsession.transitionInProgress will be set as true and no other warp will be allowed from that world until the world state is reset. 

Also, there is no need to have worldsession.transitionInProgress set as true after the coroutine call, as it will be set internally before any yield

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,13 @@
 
 ## General
 - Fixed irrelevant rooms not being unloaded while spectating other players that led to higher network throughput
-# Arena
+## Arena
 - Switched the input for banning slugcats from pckup to Shift+Click when using the mouse.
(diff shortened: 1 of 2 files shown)


6fears7:
Looks good @r3nsen , can you add an update to the CHANGELOG.md in this PR? 

I would also ask that you check and make sure normal gate behavior is not impacted from this change.

r3nsen:
I checked locally Watcher, Survivor, and Monk gate, and everything seems to be working perfectly.

6fears7:
Great work

PR #1485, opened 2026-06-22
Replaced dropdown menu for Cape Colors with textbox to input color hex code

wonky2:
What I messed with:
- Replaced the dropdown menu for cape colors with a textbox to input a hex code for any color, as well as a color preview rectangle for convenience.
- Added a checkbox for rainbow cape next to the textbox.
- Removed now-redundant code and translation strings related to the color presets of the dropdown menu.
- Replaced "Cape Colors" with "Cape Color" in translation strings, so the actual translations may need updating.

The color preview fails to update for certain colors, and I can't figure out why, but it works mostly okay for now. I'll look more into it in the future. An alternative option could be to just use the good ol' OpColorPicker, but I thought it looked too clunky. I could also add sliders for hue/saturation/lightness if that's okay, I'd probably just need to move some things around to make enough space.
The textbox only accepts number 0-9 and letters a-f and A-F, and has to have exactly 6 characters.
If it's too short or has invalid characters, it reverts to the previous valid string.
If it's too long, it gets trimmed.

The Rainbow Cape checkbox is only visible during events and defaults to `true`. The rainbow cape only works when an event is active, otherwise the cape will just be the desired color even if enabled. `SlugcatCustomization.eventCape` was modified to make that possible.

capecolor

The reason for this PR is because I've seen other non-dev players running around with custom cape colors, and I saw the trick to manually edit the config file being mentioned by someone else, so I figured it's not much of a secret anymore. May as well make it accessible.
I know it was supposed to be a dev-only privilege, so I think it's okay if you'll want to go the opposite way and just fix the custom color trick instead. Though, now that the slugcat is out of the bag, it could be a bit disappointing for those who like and have been using that trick.


6fears7:
You have my vote. @invalidunits ?

invalidunits (dismissed):

PR #1488, opened 2026-06-25
fix/sync prince

r3nsen:
- spawn princebulb only once
- sync princebulb idealPos
- sync prince lookPoint, GetToDir and GetToPos
- delay spawnPrinceBulb until the room has an owner to fix another case of princedup


r3nsen:
just to add some more context:
- not sure if the coroutine approach is the best, but I had trouble with princebulb being spawned too early
- shortcutHooks line 43: not sure if excluding prince here is the best either, I saw that this was to fix moon's wakeup in hunter, but it is duplicating some physicalObjects in worldstate, at first I thought of making it only work for SLOracleSwarmer, but didnt know yet if it would break anything else

Game/RainMeadow.ShortcutHooks.cs
```diff
@@ -40,12 +40,13 @@ private void RoomOnAddObject(On.Room.orig_AddObject orig, Room self, UpdatableAn
             orig(self, obj);
             if (OnlineManager.lobby != null)
             {
-                if (obj is PhysicalObject po && po.abstractPhysicalObject is AbstractPhysicalObject apo && !apo.GetOnlineObject(out _))
-                {
-                    self.world.GetResource()?.ApoEnteringWorld(apo);
-                    self.abstractRoom.GetResource()?.ApoEnteringRoom(apo, apo.pos);
-                }
-            }
+                if (obj is not Watcher.Prince)
```
invalidunits: Try moving this to `ShouldSyncAPOInWorld`

Online/State/RealizedPrinceBulbState.cs
```diff
@@ -0,0 +1,57 @@
...
+        [OnlineField]
+        PrinceState prince;
+        [OnlineField]
+        Vector2 idealPos;
+
+        public RealizedPrinceBulbState() { }
+        public RealizedPrinceBulbState(OnlinePhysicalObject onlineEntity) : base(onlineEntity)
+        {
+            PrinceBulb bulb = onlineEntity.apo.realizedObject as PrinceBulb;
+            prince = new(bulb.prince);
```
invalidunits: `prince` might be null, You'll need to consider than when constructing the object

Online/State/RealizedPrinceBulbState.cs
```diff
@@ -0,0 +1,57 @@
+﻿using UnityEngine;
+using Watcher;
+
+namespace RainMeadow
+{
+    public class RealizedPrinceBulbState : RealizedPhysicalObjectState
+    {
+        [OnlineField]
+        PrinceState prince;
```
invalidunits: Set this field as nullable
```suggestion
        [OnlineField(nullable=true)]
        PrinceState prince;
```

invalidunits (dismissed):

Game/RainMeadow.ShortcutHooks.cs
r3nsen: I moved it to ShouldSyncAPO, only InWorld it loads prince for each player in roomSession, only InRoom it loads for each player in worldSession, now it doesnt load in either of those, but it works

henpemaz (approved):
Data stuff looks ok

PR #1491, opened 2026-07-01
refactor: Add leaving lobby event

One-Letter-Shor:
Currently there is no way to know if you are leaving a lobby without hooking the overrides of LeaveLobby.


One-Letter-Shor:
Actually shouldn't this event be invoked after leaving to match the joined event? (and be renamed to OnLobbyLeft)

invalidunits (dismissed):

invalidunits (approved):

PR #1492, opened 2026-07-10
fix: Arena stats and API extensibility

One-Letter-Shor:
Fixes bugs with arena stats not registering properly and improves Arena API extensability


One-Letter-Shor:
Not complete yet, only basic changes have been made. When developing Hide and Seek I realized that the scoring logic is overcomplicated/inconsistent so I want to try to simplify it.

One-Letter-Shor:
@6fears7 Does arena online handle adding and removing `OnlinePlayer`s as they join and leave or is everything ushort InLobbyIds that may or may not belong to someone?

6fears7:
> @6fears7 Does arena online handle adding and removing `OnlinePlayer`s as they join and leave or is everything ushort InLobbyIds that may or may not belong to someone?

You'll have to narrow the scope of your question. When it comes to the lobby presence, yes we have to mange users who join/leave based on general stuff like the Arena player card. But the "`arenaSittingOnlineOrder`", the heart of ArenaOnline, that manages players who "request" to join and who quit an ongoing session:

## Qutting the gamesession
```cs
// ArenaRPCs.cs
        public static void Arena_RemovePlayerWhoQuit(OnlinePlayer earlyQuitterOrLatecomer)
        {
            if (RainMeadow.isArenaMode(out var arena))
            {
                if (arena.arenaSittingOnlineOrder.Contains(earlyQuitterOrLatecomer.inLobbyId))
                {
                    arena.arenaSittingOnlineOrder.Remove(earlyQuitterOrLatecomer.inLobbyId); // you'll add them in NextLevel
                }
            }
        }
```

## Readying Up
```cs
// Menu/ArenaOnlineLobbyMenu.cs
                Arena.arenaSittingOnlineOrder.AddRange(
                OnlineManager
                    .players.Where(x => ArenaHelpers.GetArenaClientSettings(x)?.ready == true)
                    .Select(x => x.inLobbyId)
            );
```

One-Letter-Shor:
I think I was unclear with what I needed. I need to add OnlinePlayers when they first join the lobby (not the arena sitting), and stop referencing them once they disconnect.

It seems like this is not possible without invoking events from [HandleJoin and HandleDisconnect](https://github.com/henpemaz/Rain-Meadow/blob/main/Online/Matchmaking/MatchmakingManager.cs#L193-L218). (Hopefully I'm wrong?)

One-Letter-Shor:
The alternative is keeping `OnlinePlayer` instances alive longer than they should and not having them added to dictionaries fast enough (which would require a bunch of redundant checks)

One-Letter-Shor:
Ok, I'm just being stupid. ~~The events in MatchmakingManager work for this.~~ OnlineGameMode has methods for this. (though it probably should be an event rather than manual calls)

One-Letter-Shor:
@6fears7 Thoughts on this?
<br>

Some of the things I did:
- Fixed many stat bugs
- Made all `ArenaPlayer` stats and lobby data stats be synced. `ArenaPlayer` stats are fast but potentially wrong (RPC), lobby data stats are reliable but a bit slower (lobby data).
- Made food points be awarded immediately instead at the end of the round. This helps with the score display being up to date, and generally makes more sense.
- Moved all stat logic into `ExternalArenaGameMode` so it is easy to override
- Removed subclass logic in `ExternalArenaGameMode`. (subclass logic will be moved into it's respective class later, dw)
- Refactored hook methods in `ExternalArenaGameMode`
  - Renamed to indicate they are hook methods
  - Pass in orig to all instead of only some of them
  - Fully give the responsibility of the hook to the methods. (conditions are not run before the call, and orig must be called if needed.)
- Refactored `ArenaPlayer` and lobby data stat methods in `ArenaOnlineGameMode`
  - Renamed to be more consistent with each other
  - Improved readability

One-Letter-Shor:
I'll add inline docs and a dedicated arena stats file when I finish everything up, but I wanted to make sure that everything currently being worked on is good.

6fears7:
Sounds good. I'll take a look at your PR in your repo to see exactly how it changed but overall what you said make sense

One-Letter-Shor:
Progress is going to be a bit delayed as I'm going on vacation for a week

One-Letter-Shor:
Alright, I think all that's needed is some play testing later today and this will be done.

Game/RainMeadow.EntityHooks.cs
```diff
@@ -302,7 +302,7 @@ private void AbstractRoom_AddEntity(On.AbstractRoom.orig_AddEntity orig, Abstrac
             orig(self, ent);
             if (OnlineManager.lobby != null && apo is not null && apo.pos.room == self.index) // skips apos being apo.Move'd
             {
-                self.world.GetResource().ApoEnteringWorld(apo);
+                self.world.GetResource()?.ApoEnteringWorld(apo);
```
One-Letter-Shor, line 305: @henpemaz Does this look reasonable? Clients were sometimes getting NREs in Drown mode (where batflies are enabled) when the they were moving on to the next arena session.

Note: Arena spawning batflies was bugged before so it never happened in the past

6fears7: adding this should be of no consequence. If the nullability removes NREs for Drown, that's an added bonus. It shouldn't impact existing behaviors because the world session should be ready to go in order for Add_AbstractEntity to even run.

6fears7 (approved):
testing scope to see if invalid is requierd for approval

6fears7 (changes requested):
waiting for sorting bug fix to be pushed

Game/RainMeadow.EntityHooks.cs
One-Letter-Shor: Is more about how it could be hiding a bug a than if it will break the game directly

6fears7 (approved):

6fears7:
Many people will benefit from your work for a long time, you should be proud 🙂

PR #1493, opened 2026-07-11
Parry Sound/Visual Fix

EdEnStonne:
A very simple fix to parry sound/visual effect.
I've come to notice that despite the Weapon Parry Fix, parries sometimes just... don't have any effects or sound cue when they happen, which is very noticable and weird.

Found out that the issue comes from `Weapon.HitAnotherThrownWeapon`, which checks the lastPos of weapon to see if the weapon go toward the same direction.

The issue is that lastPos is NOT synced by Meadow, so this check becomes inconsistent, which can cause :
- weapons to still hit despite a parry happening.
- the sound effect not happening in other's screen.
Basically it could cancel the parry sometimes.

So I put everything in a IL hook and send the lastPos in the RPC for parries.
Should work fine, I mean didn't crash or stopped parries RPC from happening when i tested it.
The test I've done was very basic but the changes are minimal.


EdEnStonne:
Aaaand I forgot CHANGELOG hold on...

6fears7:
@invalidunits first glance to me makes sense, what about you?

invalidunits (approved):

EdEnStonne:
Noticed I maybe should've put that in the realizedWeaponState, but hell if it works it works

PR #1500, opened 2026-07-15
Fixed Gourmand not being exhausted remotely

EdEnStonne:
This one is surprisingly tricky, but was fixed easily.
From the code, it really looked like gourmand exhaust should have been synced fine, but it can be clearly proven false by throwing a spear as gourmand and observing that the otehrs will not see a tired beast.
Since `aerobicLevel` is not synced in Meadow, Gourmand was always resetting its `gourmandExhausted` to false (unless you also got your aerobic level to go down remotely, by jumping or drowning for example), so my fix was to simply... skip this part of the code if it's a remote player.

I've done minimal testing, the hook didn't crash and seems to show the expected behaviour

Cheers !

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@
 - Fixed parried spears not being deflected on late parries.
 - Fixed explosive spears still damaging the player when parried.
 - Fixed sound/visual cue of parrying being inconsistant.
+- Fixed gourmand not being shown as exhausted when throwing a spear
 ## Meadow
 - Fixed creatures being able to get injured.
 ### Modders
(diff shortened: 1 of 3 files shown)


6fears7:
why don't we sync aerobicLevel in the RealizedPlayer?

EdEnStonne:
> why don't we sync aerobicLevel in the RealizedPlayer?

I figured it wasn't synced for a reason...?
Could as well do that if there's no objection to it

EdEnStonne:
Well, lemme do that then, it's abit cleaner anyway

EdEnStonne:
Well, did this, tested it, result is good, Gourmand is fixed 🎉

henpemaz (approved):

PR #1503, opened 2026-07-17
French translation

EdEnStonne:
No French traslation yet ? I could do that.


EdEnStonne:
Based off the spanish translation, i got it done all the way !
...surely there's some string missing, and i'll put them as I find them.

PR #1504, opened 2026-07-19
Diverse Arena Watcher fix

EdEnStonne:
Diverse Arena Watcher fix, including :
- Disabling Ripple Trail in online arena
- Fixed summoned Amoeba getting annihilated for diverse unwanted reasons
- Amoeba is now going in the right layer when summoned
- Summoned amoebas have idle behaviors now
- Summoned amoeba's friendly fire has been adjusted to teams and spear hit
- Summoned amoebas can stun non-player creatures now
- Amoeba won't slow down for dead players or friendlies
- Distortion effect fades away on player's death
- Watcher will have the same effect as a failed portal opening when trying to summon an Amoeba in countdown or without enough charge
- Ripple 9 makes Watcher fully invisible, leaving only a faint glow behind
- Watcher will always have while eyes in camo
- Other Watchers can see Watcher's eyes when they are in ripple space, at a lower opacity
- Watcher will see only the eyes of other Watchers in ripple space, instead of the full body
- Dev skin doesn't show in ripple space anymore
- Hands or mud doesn't show in ripple space anymore
- Watcher camo effect has been reduced for all ripple level
- Player Tag now interacts well with ripple layer change
- Added arena option to make ripple 9 Watcher actually fully invisible for everyone

Wow, that ended up being alot


EdEnStonne:
woops forgot Changelog again..

6fears7:
Ping me in discord when this is ready for review

EdEnStonne:
Need more testing, don't push in yet

EdEnStonne:
It's good to go ! Lemme check out changelog now...

6fears7 (dismissed):

Online/State/RealizedRippleSpawnState.cs
```diff
@@ -11,12 +11,15 @@ public class RealizedRippleSpawnState : RealizedPhysicalObjectState
         private int timer;
         [OnlineField]
         private bool startFadeOut;
+        [OnlineField]
```
6fears7, line 14: @henpemaz This is the only file needing your review

6fears7 (dismissed):

EdEnStonne:
Might add abit in this PR from player feedback

Arena/ArenaHooks.cs
henpemaz, line 1: Hey this doesn't seem right.

Is your IDE changing line-endings or something like that? Wrangle it into submission and see if you can get a clean commit.

Online/State/RealizedRippleSpawnState.cs
```diff
@@ -11,12 +11,15 @@ public class RealizedRippleSpawnState : RealizedPhysicalObjectState
...
+        private int rippleLayer;
```
henpemaz: Do you need 2^32 values for ripple, or do you think 2^8 would do? this could probably be a `sbyte`

Arena/ArenaHooks.cs
henpemaz: And just to clarify, there's a handful of reasons to keep commits clean, biggest one in my mind is to keep the git-blame view usable and useful. Git-blame tells you who was the last person who touched some part of the code, when, and why. If you go touching everything just because your IDE is throwing a fit, that information becomes useless.

Online/State/RealizedRippleSpawnState.cs
EdEnStonne: Actually I need 2 so this will be a bool

Arena/ArenaHooks.cs
EdEnStonne: I have no clue either on why it decided to mark everything as changed, probably hit a key stroke I shouldn't have.
I'll check on how to roll back and forth this.

EdEnStonne: Welp, a back and forth completely anihilated the issue.
Unsure why GitHub freaked out, will be more attentive to that type of issue in the future

6fears7 (dismissed):
works for me on the arena stuff impacted, thanks for keeping me in the loop

6fears7 (dismissed):

6fears7 (approved):

henpemaz (approved):

PR #1505, opened 2026-07-19
Sync LethalThunderStorm

Quaxly:
Syncs the lightning storms introduced in watcher. Lightning strikes are controlled by the room owner and replicated to all participants.


Game/RainMeadow.GameplayHooks.cs
```diff
@@ -75,6 +75,27 @@ public void GameplayHooks()
...
+            }
+            else
+            {
+                var roomSession = self.room.abstractRoom.GetResource();
+                if (roomSession != null && roomSession.isOwner)
+                {
+                    orig(self, pos, branchingChance, size, effectRadius, killRadius, delayFrames);
+                    foreach(var player in roomSession.participants)
+                    {
+                        player.InvokeRPC(RPCs.LightingMakerStrike, roomSession, pos, branchingChance, size, effectRadius, killRadius, delayFrames);
```
6fears7, line 96: Will we be invoking this on ourselves?

Quaxly: This RPC is sent from the room owner to all participants and will spawn a lightning strike. Peers will not spawn their own strikes but will instead get them from the room owner.

Edit: Just looked at it again, I'll fix it

Online/Resource/RoomSession.ThunderStorm.cs
```diff
@@ -0,0 +1,43 @@
...
+
+public partial class RoomSession
+{
+    [DeltaSupport(level = StateHandler.DeltaSupport.NullableDelta)]
+    public class LethalThunderStormData : OnlineState
+    {
+        [OnlineField(group = "counters")]
+        int lethal;
+        [OnlineField]
+        int lethalMax;
```
henpemaz: If we're synching `lightningMaker.Strike` through an RPC, then why are we synching this? clients shouldn't be ticking those counters on their own and spawning their own lightnings, it should come from room-host.

There's maybe a point in keeping it sync for when a room-transfer happens, but honestly that's very minor.

You can keep it somewhat in sync for much cheaper if you just send the new LethalMax on the RPC and reset the timer there (see how the timer is reset in TrySpawnLethalStrike)

Same for the cosmetic one imo.

Synchronizing counters is expensive and makes the delta-system work overtime, so try and be smarter about it.

henpemaz (approved):

PR #1509, opened 2026-07-20
Initialize some extenums early on

Dustylemny:
Fixes issues with region menu as slugcatname can sometimes register midway and cause an bug
Also sceneIDs are included so it will fix a bug that causes our custom bgs to not appear on main menu on restarting rain world


henpemaz (approved):
It's a fine change but half of it is unnecessary. If you want static fields to initialize, just access one of them and the entire class loads, like this https://github.com/henpemaz/Rain-Meadow/blob/main/Meadow/MeadowProgression.cs#L22

Your `InitializeExtEnums()` could literally just be
```
_ = Ext_SlugcatStatsName.OnlineSessionPlayer;
_ = Ext_SceneID.Slugcat_MeadowSquidcicada;
```

and that's all you need to load all those static fields.

One-Letter-Shor:
A dedicated static method for initialization (that does nothing) is better than discarding a static field's value because it is easy to discover

henpemaz:
The method is called "InitializeExtEnums", there isn't much question about what it does

One-Letter-Shor:
From the extenum class itself you could not tell that it is statically initialized early if you simply discard something. If it was more than one line I could see wanting to avoid it, but it would just be a singlar line per class

PR #1510, opened 2026-07-20
Gave Amoeba the player's color

EdEnStonne:
I really though it'd take a TON of time and destroy alot of stuff to have something semi-functional. (hence the separated branch)
Turns out... took a few hours to get that thing working.

I gave this decent testing in arena, seems to work fine in and out of the ripple world.

Note : the asset bundle was modified, since it uses a new shader to change the color, so in playtest the players will have to also get that file to be able to run the build correctly


6fears7 (dismissed):

6fears7:
Tested this myself, works great. Awesome work.

EdEnStonne:
Did not check team color, will do

henpemaz (dismissed):
Cool, highly optional, and not good for compatibility.

Part of me thinks it shouldn't be part of the mod, but hey it's a mess already, up to your judgement

EdEnStonne:
> Cool, highly optional, and not good for compatibility.

How bad is it in compatibility @henpemaz  ? 
I could try to make things more compat friendly, but I'm... unsure on what you're refering to rn.

henpemaz (approved):

henpemaz:
> > Cool, highly optional, and not good for compatibility.
> 
> How bad is it in compatibility @henpemaz ? I could try to make things more compat friendly, but I'm... unsure on what you're refering to rn.

Loading new shaders is not the greatest for compatibility with other mods. Can't really hook a shader or stack changes to it, only overwrite it.

But if it's really just the amoeba shader I guess that's so specific it's unlikely it will cause issues.

henpemaz:
Also my train of thought is that "if you're fixing something for both local-play and online-play, maybe it shouldn't be in the online-mod. Not critical in this case but food for thought.

6fears7 (approved):

EdEnStonne:
> Also my train of thought is that "if you're fixing something for both local-play and online-play, maybe it shouldn't be in the online-mod. Not critical in this case but food for thought.

I do fully understand that line, and I do agree to.
However, I'm not really sure what you are referencing here...?
Watcher cannot summon Amoebas in local, only in Meadow, so this is online exclusive. The colored Amoeba is a feature that has been asked for awhile, that's why I thought it'd be appropriate to make a PR out of it.

PR #1512, opened 2026-07-24
fix: Tinnitus fix attempt

Quaxly:


6fears7:
Sir this says mushrooms

PR #1517, opened 2026-07-25
Rain Meadow - The 4:3 Update

T1mbits:
fixes direct connect UI components in the lobby select menu to not be perpetually fixed relative to the right side of the screen and also modified the introroll a bit so that it works properly on smaller aspect ratios (like 4:3). The placement of logos is near identical in the introroll (the akupara and videocult logos are a bit higher than before) and the direct connect objects were pushed to the left by 20px to fit on the screen properly but are otherwise completely normal on the usual 16:9 aspect ratio (as well as other ratios).


iulh:
i really want to highlight my contribution to 4:3 support, since [spectate hud was unusable back then](https://github.com/henpemaz/Rain-Meadow/pull/1473) and it was the only critical flaw in my opinion.
i really glad to see you've found out the bugginess of other ratios besides 16:9

henpemaz (approved):

PR #1520, opened 2026-07-25
Complete Stowaway sync

r3nsen:
- tendrils synced
- bite and killer bite synced - and clients will not bite on their own
- behavior synced - and clients will not change behavior on their own
- calling homepos instead of bodyChunk.pos in bodySetup() because bodyChunk.pos will be different for clients due to sync


Online/State/RealizedStowawayState.cs
```diff
@@ -6,104 +6,169 @@
...
     {
-        [OnlineFieldHalf]
-        float retractFac;
+        const int chunksToSync = 3;
+
         [OnlineField]
         bool fired;
+
+        [OnlineField(group = "counters")]
+        int scooldown;
```
henpemaz: spitCooldown is only set to Range(40, 60) and only tested for <0, this field could probably be a `sbyte` with some care on the negative handling (if negative just set to -1)

Online/State/RealizedStowawayState.cs
```diff
@@ -6,104 +6,169 @@
...
+        
+        [OnlineFieldHalf]
+        float retractFac;
         [OnlineFieldHalf(group = "counters")]
-        float cooldown;
+        float hcooldown;
+        [OnlineFieldHalf]
+        float idealLength;
+
+        [OnlineField(nullable = true)]
```
henpemaz: this could probably be half-precision as well, it will be off by a couple pixels in the game's biggest rooms, but so are the other synched positions.

Game/RainMeadow.CreatureHooks.cs
```diff
@@ -52,6 +52,107 @@ private void CreatureHooks()
...
+            On.MoreSlugcats.StowawayBugAI.Update += StowawayBugAI_Update; // clients will not change behavior on their own
+            IL.MoreSlugcats.StowawayBug.Update += StowawayBug_Update; // clients will not bite on their own
+            IL.MoreSlugcats.StowawayBug.bodySetup += StowawayBug_bodySetup; // calling homepos instead of bodyChunk.pos because bodyChunk.pos will be different for clients due to sync
+        }
+
+        private void StowawayBug_bodySetup(ILContext il)
+        {
+            var c = new ILCursor(il);
+
+            Func<Instruction, bool>[] predicates = {
```
henpemaz, line 118: Would be nice if you included a commend on IL matches to say which lines of code it's supposed to match. Not line number, just a ctrl-c ctrl-v of the line where you're putting the cursor so it can be found easily in the future if needed.

Same for the other match.

henpemaz (changes requested):
Overall good pr, a couple things to address on the individual comments.

And if you're willing to put extra time on it, I think the better strategy would be to sync attacks through an RPC and prevent clients from running the attack logic. Basically have the block in `else if (this.AI.behavior == StowawayBugAI.Behavior.Attacking || this.placedDirection.y > 0.3f)` be behind an owner check and replicated to clients via RPC, then a bunch of these fields no longer need synching, and clients don't randomly fire on their own. But this is optional.

Game/RainMeadow.CreatureHooks.cs
r3nsen: I'm adopting this, ty

Online/State/RealizedStowawayState.cs
r3nsen: done

Online/State/RealizedStowawayState.cs
r3nsen: counters removed due to rpc

r3nsen:
> And if you're willing to put extra time on it, I think the better strategy would be to sync attacks through an RPC and prevent clients from running the attack logic. Basically have the block in `else if (this.AI.behavior == StowawayBugAI.Behavior.Attacking || this.placedDirection.y > 0.3f)` be behind an owner check and replicated to clients via RPC, then a bunch of these fields no longer need synching, and clients don't randomly fire on their own. But this is optional.

StowawayHeadAttack RPC  and StowawayBite RPC created

henpemaz (approved):

PR #1521, opened 2026-07-25
fix: watcher warping and serialization angry noises

6fears7:
Henpe some of these changes may no longer be required post-compression fix will test later, just check DeflateStream.cs for now and lmk thoughts. I originally thought it was all linked but I'll split these into multiple PRs as needed 

- Improves watcher warping integrity by updating some logic in MoveOnly. 
- Fixed sub-optimal region loading. Overworld loads all regions by default. We store that in our overworld session. That led to a lot of unnecessary information being passed around. Instead OnlineGameMode just grabs the regions relative to the lobby (timeline, campaign) instead of adding in everything Watcher DLC adds (like 52 regions). 
- As a result, Meadow mode was updated to accommodate this. 
- Fixes serialization poisoning in the DeflateStrea's Compress().

TL;DR the Compress() of DeflateStream was using CopyTo and passing in the buffer size, not length. So every compressed packet had stale buffer data and was too big to send. So fixed to only copy what we wrote. 

The symptom of this was you'd join something that'd push the Serializer to the edge (Watcher) story, then you'd jump into a different story mode or even another game mode and the Serializer got poisoned by the massive 192Kb buffer story watcher savestate string and filled it with ~142Kb of data.

 And changing lobbies doesn't clean up the raw bytes, so other sessions kept hanging onto it, meaning even a low-impact gamemode like Meadow got hit.

These changes have been tested across all three game modes for 3 hours.
- Meadow had tokens and ghosts present across 3 different regions
- Story mode worked as expected
- Arena worked as expected

### DIFF
--- a/Extensions.cs
+++ b/Extensions.cs
@@ -102,9 +102,6 @@ namespace RainMeadow
                 }
             }
         }
-        /// ChangeRooms dereferences world.GetAbstractRoom for both the old and the new coordinate, and
-        /// either resolves to null once a coordinate names a room outside this world - which is routine
-        /// when a state arrives from a player who has already warped to another region.
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static bool CanChangeRoomsTo(this AbstractPhysicalObject apo, WorldCoordinate newCoord)
         {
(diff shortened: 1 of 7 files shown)


6fears7:
closing, will push separate branches

PR #1522, opened 2026-07-25
fix: compress only the bytes the state wrote

6fears7:
Fixes painful bug Watcher exposed where a large state dirties the serializer so any state over zipTreshold could never be sent. A dropped state never becomes the acked baseline, so the subscription fell back to sending a full state every tick, leading to a death loop. 

This was tested across all modes for many hours


henpemaz (approved):
Good catch!

PR #1525, opened 2026-07-26
Add runtime DLL hash verification to prevent mod ID spoofing bypasses

XuanzeBufeng:
Currently, bans rely on mod IDs, which can be easily changed to bypass restrictions. This PR adds a runtime hash check: when a banned mod is uploaded, its DLL hash is recorded. When a player joins, the server checks both the local mod list AND the hashes of running assemblies. If a banned hash is detected, the join is blocked.

### DIFF
--- a/Menu/LobbyInfo.cs
+++ b/Menu/LobbyInfo.cs
@@ -13,8 +13,8 @@ namespace RainMeadow
         public string requiredMods;
         public string bannedMods;        
         public bool pinned;
-
-        public LobbyInfo(string name, string mode, int playerCount, bool hasPassword, int? maxPlayerCount, string highImpactMods = "", string bannedMods = "")
+        public string bannedHashes;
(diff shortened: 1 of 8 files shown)


One-Letter-Shor:
Is it possible to use the BepInEx guid (that is passed into the BepInPlugin attribute) instead of hashing dlls?

I'm not sure about the implementation, but it seems like it would be simpler. Just trying to give you ideas.

One-Letter-Shor:
Additionally, the guid would not need to be updated per update (newer OR older version) which means that the banned mods do not necessarily need to be installed

XuanzeBufeng:
This is indeed a simpler and more reliable approach. Perhaps we could use reflection to scan the running assemblies, find all classes that inherit from BaseUnityPlugin, and retrieve their properties and GUIDs. I might give that a try. If it works, this PR would likely become unnecessary, since its stability isn't as solid as the solution you proposed.

6fears7:
@XuanzeBufeng you could update this PR with the changes or close and make a new one. You're call.

One-Letter-Shor:
(You should mark this as a draft if you are working on changes)

PR #1528, opened 2026-07-28
Minor UI/SFX fixes

cgytrus:
- fixed a bug in arena where after transitioning to MultiplayerResults, the result box bump sound effect would play for each player instead of just once, so the more players there are in the lobby the louder the sound, being potentially disturbing or dangerous in large lobbies
  - this is actually a vanilla bug but there the limit is 4 players so it isnt that much of a problem, this PR fixes it only for rain meadow
- fixed a bug in arena where making an input on the results overlay screen would call PlayerPressedContinue for each player, causing the ready sfx to play for each player as well instead of just once
  - the ui itself doesnt affect anything here since non-host players cant ready or unready when already in a match anyway (though it looks like it was intended to actually be implemented at some point?)
- fixed a bug in story where on the sleep screen if a non-host player readies up, then the host continues to the game, the sleep screen of the non-host player would try to switch to the game process every update, causing sfx spam
- fixed the "to lobby" button in MultiplayerResults in arena drawing behind the result boxes
  - visible if you scroll up enough or if there are enough players
- fixed a bug in story and arena where nametags can visibly "jump" instead of instantly teleporting when the camera moves or the player enters a pipe, etc.
  - by just setting lastPos/lastAlpha in spots where changes must be instant
  - this type of bug and this type of fix for it are very common in vanilla rain world too because of the way its programmed

### DIFF
--- a/Arena/ArenaRPCs.cs
+++ b/Arena/ArenaRPCs.cs
@@ -328,24 +328,29 @@ namespace RainMeadow
 
 
         [RPCMethod]
-        public static void Arena_ReadyForNextLevel()
+        public static void Arena_ReadyForNextRound(RPCEvent rpcEvent)
         {
-            if (RainMeadow.isArenaMode(out var arena))
+            if (!RainMeadow.isArenaMode(out ArenaOnlineGameMode arena))
+                return;
+            if (Custom.rainWorld.processManager.currentMainLoop is not RainWorldGame game)
+                return;
+            if (game.manager.upcomingProcess is not null)
+                return;
+            List<ArenaSitting.ArenaPlayer> players = game.GetArenaGameSession.arenaSitting.players;
+            if (rpcEvent.from == OnlineManager.lobby.owner)
             {
-                var lobby = (RWCustom.Custom.rainWorld.processManager.currentMainLoop as ArenaLobbyMenu);
-                var game = (RWCustom.Custom.rainWorld.processManager.currentMainLoop as RainWorldGame);
-                if (game.manager.upcomingProcess != null)
-                {
-                    return;
-                }
-                for (int i = 0; i < game.arenaOverlay.resultBoxes.Count; i++)
-                {
-                    game.arenaOverlay.result[i].readyForNextRound = true;
-                }
-                game.arenaOverlay.nextLevelCall = true;
-
+                foreach (ArenaSitting.ArenaPlayer player in game.GetArenaGameSession.arenaSitting.players)
+                    player.readyForNextRound = true;
+                game.arenaOverlay.PlaySound(SoundID.UI_Multiplayer_All_Players_Ready);
             }
-
+            else
+            {
+                int index = ArenaHelpers.FindOnlinePlayerNumber(arena, rpcEvent.from);
+                if (index < 0 || index >= players.Count)
+                    return;
+                players[index].readyForNextRound = true;
+                game.arenaOverlay.PlaySound(SoundID.UI_Multiplayer_Player_Result_Box_Player_Ready);
+            }
         }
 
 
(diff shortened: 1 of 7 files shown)


6fears7:
@One-Letter-Shor @EdEnStonne any thoughts on this?

EdEnStonne:
Read the code, compiled it, tested in arena, does what it says, all good !

One minor set-back, which seems intended on those fixes but I'm unsure I like : when pressing any key to be ready for next round, it will behave like local and only ready your player, and nobody else.
Which is fine imo for non-host, they don't decide if they should move to next round or not, but for the host it does look abit odd to only have them marked as ready and move on, as in local every box has to be pushed right to be able to move to the next arena.
It... barely matters ngl, I'm being perfectionist, the rest behave as it should and I'm glad this was tackled. 

Here's a picture on how it looks like when one person gets ready for the other pals that haven't tested it :
image

_Would love to see the non-host marking themselves as ready being synced with other players, but that'd require some more juggling with States and a RPC, so if it isn't in your plans, I can definitely tackle that on after this PR has been approved._

cgytrus:
yeah i agree it looks a bit weird, can try to figure something out with that if you want but if youre gonna implement syncing the ready state i think this implementation might be a better setup for that

and yeah as i mentioned it does seem like theres some code for syncing the ready state so i think it was either already implemented but then removed or intended to be implemented but never finished at some point?

6fears7:
It was explored. However I think it's not a great feature because online players can go AFK regularly, so waiting for everyone to ready up would get old quick.

EdEnStonne:
> It was explored. However I think it's not a great feature because online players can go AFK regularly, so waiting for everyone to ready up would get old quick.

Oh not waiting for them, just as an indication.
Host can still make everyone move on like before.

cgytrus:
oh i think thatd be nice actually, i could implement that

6fears7:
Go for it

cgytrus:
added another fix that i forgot about earlier: in story and arena nametags can visibly "jump" instead of instantly teleporting when the camera moves or the player enters a pipe, etc.

Arena/ArenaRPCs.cs
```diff
@@ -328,24 +329,34 @@ public static void Arena_InitialSetupTimers(int setupTime, int saintMaxTime)
...
+                game.arenaOverlay.PlaySound(SoundID.UI_Multiplayer_All_Players_Ready);
+                if (readyPlayer.isMe)
                 {
-                    return;
-                }
-                for (int i = 0; i < game.arenaOverlay.resultBoxes.Count; i++)
-                {
-                    game.arenaOverlay.result[i].readyForNextRound = true;
+                    game.arenaOverlay.countdownToNextRound = game.arenaOverlay.countdownToNextRound == -1 ? 10 :
+                        Math.Min(game.arenaOverlay.countdownToNextRound, 10);
```
6fears7, line 349: If I'm reading this diff right, we are adding a 10s timer to the time to load between rounds? 

If so, I'd recommend we drop it to 5. It gives time for other players to see the scoreboard, but it won't infuriate people who are wanting to push for the next level too much.

cgytrus: this is in ticks so like 0.25s, i just copied this from the vanilla implementation of PlayerPressedContinue

the way this works in vanilla is every time a player presses continue it sets their readyForNextRound to true and calls PlayerPressedContinue, which starts the countdown to the next round and if all the players are ready sets it straight to 10

the way this works here is when the local player presses continue it calls this rpc method for everyone which sets readyForNextRound, and if the local player is the host it also starts the countdown already at 10 since the host forces readyForNextRound = true for every player

6fears7:
Good to merge if you are once a changelog is added

cgytrus:
should note that one of the bugs fixed here is listed as "When clients are readied then continue from the hibernation screen, they hear the continue button sound many times in a row." in the bugs spreadsheet so should probably be removed from there once this is merged

6fears7 (approved):

PR #1529, opened 2026-07-28
Fit the main menu exit button in the first column

cgytrus:
increases the limit of the vertical button columns on the main menu by 1 to fit in the meadow button without pushing exit to the second column
i'm not sure whether this is change desirable at this point but this patch is simple enough that its fine if this doesnt get merged
also not sure whether the hook should go in EssentialMenuHooks or MenuHooks, lmk if i should move it

| before | after |
|--------|--------|
| image | image  |

### DIFF
--- a/Menu/RainMeadow.MenuHooks.cs
+++ b/Menu/RainMeadow.MenuHooks.cs
@@ -691,15 +691,15 @@ namespace RainMeadow
             try
             {
                 ILCursor cursor = new(il);
-                int maxCount = 0;
-                if (!cursor.TryGotoNext(code => code.MatchStloc(1)) ||
-                    !cursor.TryGotoPrev(MoveType.After, code => code.MatchLdcI4(out maxCount)))
-                {
-                    Warn("failed to increase main menu button limit");
-                    return;
-                }
-                cursor.Emit(OpCodes.Pop);
-                cursor.Emit(OpCodes.Ldc_I4, maxCount + 1);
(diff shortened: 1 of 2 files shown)


Menu/RainMeadow.MenuHooks.cs
```diff
@@ -685,6 +686,28 @@ private void ProcessManager_PostSwitchMainProcess(On.ProcessManager.orig_PostSwi
...
+                ILCursor cursor = new(il);
+                int maxCount = 0;
+                if (!cursor.TryGotoNext(code => code.MatchStloc(1)) ||
+                    !cursor.TryGotoPrev(MoveType.After, code => code.MatchLdcI4(out maxCount)))
+                {
+                    Warn("failed to increase main menu button limit");
+                    return;
+                }
+                cursor.Emit(OpCodes.Pop);
+                cursor.Emit(OpCodes.Ldc_I4, maxCount + 1);
```
One-Letter-Shor: I would recommend making constants for local vars and args rather than using literals (even if it's name is just `locIndex1`) and add a small comment describing what type it is and what it's roughly used for.

This can help with debugging when the hook breaks.

cgytrus: good point, also decided to slightly refactor the hook as well, i think it makes more sense that way

6fears7:
@T1mbits do you care where this menu hook goes?

T1mbits:
Not particularly. All good in my opinion

PR #1530, opened 2026-07-28
Improve arena timers

cgytrus:
fixes the timers sometimes starting off a bit too fast by slightly changing the timer logic

before this change the timers were based off the host's system time and advanced every time the second changed, which means that if the host starts the match at, say, 10:42:15.500, the match would start at 10:42:20.000 (in 4.5 seconds) instead of 10:42:20.500 (in 5 seconds, like intended)

after this change the timers are based off the lobbys update cycle, advancing every time the tick counter counts past `framesPerSecond` and resetting the counter every time a timer is started, which means timers properly count starting from the time they were started at (unless there are somehow a lobby countdown and a setup timer running at the same time)

this solution isnt perfect but this is way easier to implement into the current system than a solution based off syncing timer start/end timestamps instead of incrementing/decrementing timer variables every second manually

### DIFF
--- a/Arena/ArenaTimer.cs
+++ b/Arena/ArenaTimer.cs
@@ -32,6 +32,7 @@ namespace RainMeadow
             arena.arenaPrepTimer = this;
             session = arenaGameSession;
             arena.trackSetupTime = arena.externalArenaGameMode.SetTimer(arena);
+            arena.timerTicks = 0;
             matchMode = TimerMode.Waiting;
 
             timerLabel = new FLabel(Custom.GetFont(), FormatTime(0))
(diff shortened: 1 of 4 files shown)


6fears7:
@One-Letter-Shor @EdEnStonne any remarks on this?

One-Letter-Shor:
That looks good to me

EdEnStonne:
Read the code, seems ok, not everyday that I see people using uint, but for how you are using it, it doesn't matter.
Tested it, seems to do exactly what it says ! 
Looks all good for me :D

cgytrus:
the uint was an accident that i only noticed after pushing but it doesnt matter here so i left it as is lol

the reason it ended up like that is it was originally interacting with the tick values that are also uint

6fears7 (dismissed):

invalidunits (dismissed):

cgytrus:
just a note here, i will probably make a PR that properly refactors arena's timers but it *might* take a while for me to figure out how to do it right so this is still fine to merge for now

6fears7 (approved):

PR #1532, opened 2026-07-30
refactor: Simplify external arena game mode checks

One-Letter-Shor:
- Changed method names to PascalCase.
- Removed requirement of passing `ArenaOnlineGameMode`.
- Renamed out'd external arenas to be consistent & correct.
  - tb, team, teamBattleMode -> teamBattle
  - d, drownData -> drown
  - chMode -> challenge
- Updated directly related documentation.

I still need to test that RM isn't bugged in a way that would result in the `InvalidOperationException` being thrown.

### DIFF
--- a/Arena/ArenaHelpers.cs
+++ b/Arena/ArenaHelpers.cs
@@ -182,7 +182,6 @@ namespace RainMeadow
 
 
             OnlinePlayer? deadPlayer = ArenaHelpers.FindOnlinePlayerByFakePlayerNumber(arena, excludedPlayerNumber);
-            bool isTeamBattle = TeamBattleMode.IsTeamBattleMode(out _);
             int[] allAlivePlayers = [];
 
             for (int i = 0; i < session.arenaSitting.players.Count; i++)
@@ -208,7 +207,7 @@ namespace RainMeadow
                 }
 
                 // EXCLUSION 3: Not on the same team
-                if (isTeamBattle && deadPlayer != null)
+                if (TeamBattleMode.IsTeamBattleMode(out _) && deadPlayer != null)
                 {
                     if (ArenaHelpers.CheckSameTeam(alivePlayer, deadPlayer)) continue;
                 }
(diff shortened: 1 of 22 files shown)


One-Letter-Shor:
@6fears7 Actually, should the checks be `Is<GameModeName>()` instead of `Is<GameModeName>Mode()`?
for example: `IsTeamBattle` instead of `IsTeamBattleMode`

(And yes I am aware there are a few mistakes I need to fix, I'm working in it rn)

One-Letter-Shor:
The naming for external arenas is `externalArena` and `<gameModeName>` (ex: `teamBattle`, NOT `teamBattleMode`)

(the type names are `<GameModeName>Mode` (ex: `TeamBattleMode`))

One-Letter-Shor:
Actually nvm, `Is<GameModeName>Mode` is probably better

6fears7:
What about using gametype instead?

- FFAGametype
- TeamBattleGametype
- DrownGameType

The reason i say that is because we seem cursed with recursive naming problems (HUD.HUD, RainMeadow.RainMeadow) and we'd be falling into the same trap with IsArenaMode having "TeamBattleMode".

Which, i understand it's not really a game type in some respects because all the extern modes are Competitive game type, but we do always leverage custom "InitGameType", so to me it makes sense to call them gametypes

One-Letter-Shor:
I think that the real problem with the naming is due to not just relying on `ArenaOnlineGameMode.externalGameMode`. We should really just use pattern matching on that field.

I'm not sure if there's a valid reason for that, and I didn't want to completely change how the logic works without asking (plus external modders probably won't want it to change so much)

6fears7:
If we're going to do healthy updates to the API, this is the time to do it. 

I'm fine with IsGameModeName<Mode>

One-Letter-Shor:
I forgot to mention this, but other changes are needed to properly use pattern matching (or at least, safely). `ArenaOnlineGameMode.currentGameMode` is not always synced with `ArenaOnlineGameMode.externalArenaGameMode`, and I believe switching to pattern matching will introduce a bug.

In the future I will fix the desync between the fields

6fears7 (approved):

6fears7:
https://youtu.be/rwI_7tDAbUU?is=ly8zOsO2nQVj82_n

PR #1533, opened 2026-07-30
Makes people wait others to load in Arena

EdEnStonne:
So another random beef of mine I wanted fixed.
In FFA or Teams, the countdown for the grace period is extended to make sure everyone can load in before counting down.
This is a great system ! It stops people will slow pc to get instantly killed. However, while the arena is waiting for every player, the ones that loaded in (mainly the host) can move freely and already take crusial items and spots.

I do find it quite unfair, expectially in lobby with low countdown that have no regard on spawn killing. It also puts people with slow pc at a constant disadvantage (I mean, one of many) that could be easily be solved.

My solution is simple : while the countdown is waiting for everyone, the players will... wait in their pipe !
This doesn't take in account the time for players to load all the items and creatures in the room, but it's already a big plus from the host getting a free 1-5s on anyone in the grace period.

The countdown has a safety catcher in case not everyone makes it. This was already there, I've just made it configurable if some people don't wanna deal with waiting to a maximum of 7.5sfor everyone to join in.

**TLDR :** 
- Fixed host being able to shelter immediatly before anyone loaded in
- Locked people in their shelter at the start of the round until everyone loaded in

**Note :**
This branch needs playtest with real players. If this happen to have underlying issues, I'd rather it not be pushed for 1.15 release so I can take my time to tackle everything.

### DIFF
--- a/Arena/ArenaHooks.cs
+++ b/Arena/ArenaHooks.cs
@@ -226,7 +226,7 @@ namespace RainMeadow
                             {
                                 startBump.startGameCounter = 10;
                             }
-                            return isArenaOnlineNOTReady;
+                            return isArenaOnlineWaitingForPlayers;
                         }
                     );
                     cursor.Emit(OpCodes.Brtrue, label);
--- a/Arena/ArenaHooks.cs
+++ b/Arena/ArenaHooks.cs
@@ -222,7 +222,7 @@ namespace RainMeadow
                                 && arenaOnline.externalArenaGameMode.HoldFireWhileTimerIsActive(arenaOnline)
                                 && (arenaOnline.arenaPrepTimer is null 
                                     || arenaOnline.arenaPrepTimer.showMode == ArenaPrepTimer.TimerMode.Waiting);
-                            if (isArenaOnlineNOTReady)
+                            if (isArenaOnlineWaitingForPlayers)
                             {
                                 startBump.startGameCounter = 10;
                             }
--- a/Arena/ArenaHooks.cs
+++ b/Arena/ArenaHooks.cs
@@ -218,15 +218,15 @@ namespace RainMeadow
                     cursor.Emit(OpCodes.Ldarg_0);
                     cursor.EmitDelegate((ArenaBehaviors.StartBump startBump) => 
                         {
-                            bool isOnlineArenaNotReady = isArenaMode(out var arena) 
-                                && arena.externalArenaGameMode.HoldFireWhileTimerIsActive(arena)
-                                && (arena.arenaPrepTimer is null 
-                                    || arena.arenaPrepTimer.showMode == ArenaPrepTimer.TimerMode.Waiting);
-                            if (isOnlineArenaNotReady)
+                            bool isArenaOnlineNOTReady = isArenaMode(out var arenaOnline) 
+                                && arenaOnline.externalArenaGameMode.HoldFireWhileTimerIsActive(arenaOnline)
+                                && (arenaOnline.arenaPrepTimer is null 
+                                    || arenaOnline.arenaPrepTimer.showMode == ArenaPrepTimer.TimerMode.Waiting);
+                            if (isArenaOnlineNOTReady)
                             {
                                 startBump.startGameCounter = 10;
                             }
-                            return isOnlineArenaNotReady;
+                            return isArenaOnlineNOTReady;
                         }
                     );
                     cursor.Emit(OpCodes.Brtrue, label);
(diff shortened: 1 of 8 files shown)


OnlineUIComponents/PlayerSpecificOnlineHud.cs
EdEnStonne, line 193: I did have to change the way the playerHUD detects a player "in shortcut" to make this work. 
I don't think that should cause any underlying bugs down the line, but I still want that change reviewed carefully in case it does.

Arena/ArenaHooks.cs
```diff
@@ -197,9 +197,84 @@ public void ArenaHooks()
...
+                if (cursor.TryGotoNext(MoveType.After,
+                    i => i.MatchLdarg(0),
+                    i => i.MatchLdfld<ArenaBehaviors.StartBump>(nameof(ArenaBehaviors.StartBump.startGameCounter)),
+                    i => i.MatchBrtrue(out label)))
+                {
+                    // Don't trigger the bump till everyone is ready
+                    cursor.Emit(OpCodes.Ldarg_0);
+                    cursor.EmitDelegate((ArenaBehaviors.StartBump startBump) => 
+                        {
+                            bool isOnlineArenaNotReady = isArenaMode(out var arena)
```
One-Letter-Shor: Just as an FYI, `arenaOnline` is a better name because `arena` is ambiguous with other types such as `ArenaGameSession` and `ExternalArenaGameMode`.
Also, for consistency `isOnlineArenaNotReady` should probably be renamed to `isArenaOnlineNotReady` then negated to `isArenaOnlineReady` as negative bool names are harder to read in conditions. (use `!` instead of "Not" in the name)

EdEnStonne: Fair enough stuff, I do use `arenaOnline` when there is a possibility of ambiguity but ig I copied abit too on the last PRs...
Unsure about the bool stuff, it's used only here, shouldn't be much of a bother.
Anyway, gonna check that tmr

OnlineUIComponents/PlayerSpecificOnlineHud.cs
EdEnStonne: Seems to not be an issue, all good

EdEnStonne:
Done some testing and it looks all good !
This PR is ready for review

Arena/ArenaHooks.cs
```diff
@@ -196,9 +196,84 @@ public void ArenaHooks()
...
+                if (cursor.TryGotoNext(MoveType.After,
+                    i => i.MatchLdarg(0),
+                    i => i.MatchLdfld<ArenaBehaviors.StartBump>(nameof(ArenaBehaviors.StartBump.startGameCounter)),
+                    i => i.MatchBrtrue(out label)))
+                {
+                    // Don't trigger the bump till everyone is ready
+                    cursor.Emit(OpCodes.Ldarg_0);
+                    cursor.EmitDelegate((ArenaBehaviors.StartBump startBump) => 
+                        {
+                            bool isArenaOnlineNOTReady = isArenaMode(out var arenaOnline)
```
6fears7:
```suggestion
                            bool isArenaOnlineWaitingForPlayers = isArenaMode(out var arenaOnline) 
```

Arena/ArenaHooks.cs
```diff
@@ -196,9 +196,84 @@ public void ArenaHooks()
...
+                {
+                    // Don't trigger the bump till everyone is ready
+                    cursor.Emit(OpCodes.Ldarg_0);
+                    cursor.EmitDelegate((ArenaBehaviors.StartBump startBump) => 
+                        {
+                            bool isArenaOnlineNOTReady = isArenaMode(out var arenaOnline) 
+                                && arenaOnline.externalArenaGameMode.HoldFireWhileTimerIsActive(arenaOnline)
+                                && (arenaOnline.arenaPrepTimer is null 
+                                    || arenaOnline.arenaPrepTimer.showMode == ArenaPrepTimer.TimerMode.Waiting);
+                            if (isArenaOnlineNOTReady)
```
6fears7:
```suggestion
                            if (isArenaOnlineWaitingForPlayers)
```

Arena/ArenaHooks.cs
```diff
@@ -196,9 +196,84 @@ public void ArenaHooks()
...
+                            {
+                                startBump.startGameCounter = 10;
+                            }
+                            return isArenaOnlineNOTReady;
```
6fears7:
```suggestion
                            return isArenaOnlineWaitingForPlayers;
```

EdEnStonne:
Note to self : Don't put # in comit names

6fears7 (approved):

PR #1400, opened 2026-04-14
feat: slugicon

Quaxly:
Adds the SlugIcon class which lets you render Rain World Wiki style icons using a slugcat class and it's colors.

PR #1547, opened 2026-08-03
feat(ui): lobby select menu overhaul redux

T1mbits:
New lobby select menu.

image

New features:
- Tabs for lobby metadata, filters, and networking options
- show new lobby data such as lobby timeline and mods needed to join the lobby directly in the lobby select menu
- SlugIcon: a re-implementation of #1400 that should be usable in HUDs and allows mods to add their own icons, with fallback sprites and colours.
- activeTimeline in LobbyInfo, showing the active timeline in lobbies where that is relevant (Meadow mode and in-game story lobbies)

note: I drastically altered the git commit history to make it cleaner, and cannot guarantee all commits build properly. I can however guarantee that it builds as of the latest


6fears7 (approved):
You know, for someone who complains about their lack of will to do things, when you lock in, you get a lot of impressive work done well.

That panel on the side looks good and provides some opportunities for future bells and whistles as needed, perhaps such as future lobby tags. 

You are right that the ModManager could probably use a facelift at some point.

.vscode/tasks.json
henpemaz, line 1: Can you just double-check if you wanted to commit this?

henpemaz (dismissed):

.vscode/tasks.json
T1mbits: yeah it’s intentional. The old tasks didn’t really do anything different so I just removed them

invalidunits (dismissed):
I want to merge this with the backend router changes in 1.16.0.  But it can be merged whenever @T1mbits wants to.

T1mbits:
I don’t particularly see any reason to withhold the pr

6fears7:
@T1mbits it creates more work for invalid due to the LAN changes you're making here, router PR significantly changes the matchmaking layer. 

I'll let you read this then ping me if you want to merge or wait

T1mbits:
@6fears7 If it’s matchmaking changes that’d cause the issue I can undo the changes in matchmaking, unless LobbyInfo is also significantly altered by router pr. The new lobbyinfo/matchmaking stuff is completely optional and the menu works without those changes

6fears7 (dismissed):
```
[Info   :RainMeadow] 23:14:20|316959|RequestLobbyPacket.Process
[Error  :RainMeadow] 23:14:20|316959|LANNetIO.RecieveData:System
eption: Value cannot be null.
Parameter name: value
  at System.IO.BinaryWriter.Write (System.String value) [0x00003
ca45069c528c15c9fdd749>:0
  at RainMeadow.InformLobbyPacket.Serialize (System.IO.BinaryWri
0064] in /Online/Matchmaking/Packets/InformLobbyPacket.cs:39
```

when creating a LAN Arena lobby, clients cannot see the lobby. Only happens in Arena gamemode

6fears7 (approved):

henpemaz (approved):

invalidunits (approved):

PR #1548, opened 2026-08-04
update japanese translation

solo-snuggles:
hello :o i'm on git now
a few things to note, all important and directly relevant elements are translated.
with it, i will also forward an updated MISSING_jap.txt to the discord (#playtesting, Translator thread)
All context elements in Arena Mode's main tab (most checkbox elements) left untranslated due to Rain World's lovely font.


Mod/text/text_jap/strings.txt
```diff
@@ -1,695 +1,829 @@
...
+Peach Lizard|ピーチトカゲ
+Slugcat NPC|ナメクジネコ NPC
+MothGrub|モスグラブ
+Angler|アングラー
+TowerCrab|タワークラブ
+
+//misc
+
+/lacking translate string?
+Current points:|現在のポイント
```
6fears7, line 829: @One-Letter-Shor if you think of it during the refactor, please add the Translate() method around menu strings

PR #1554, opened 2026-08-07
feat(hud): SlugIcon

T1mbits:
Replace the in-game slugcat indicator with SlugIcon, showing the player's slugcat class and custom colours. The change only replaces the "Kill_Slugcat" element, so the host symbol, arena symbols, shelter symbols, etc. will still take priority over the new icon. An option has been added to the remix menu to use the old icon if people prefer.

Modded slugcats are supported through the exact same API and fallbacks, though I haven't tested modded slugcats or stress-tested the icons in all the arena gamemodes. Should work though since most of the heavy lifting is done by SlugIcon.

This completely replaces #1400.

TODO:
- [x] Slugpup icon
- [x] have fallback hardhat be uncoloured, colour using in-game tinting
- [x] test watcher aura
- [x] verify (and potentially patch) bugs pertaining to death and refreshes
- [x] test (modded slugcats & colours)
- [x] minimalist slug icon option

image
image


Arena/ArenaOnlineGameModes/BaseGameMode.cs
```diff
@@ -446,7 +446,6 @@ OnlinePlayer player
             if (SpecialEvents.EventActiveInLobby<SpecialEvents.AprilFools>() || playerGotSlots)
             {
                 SpecialEvents.LoadElement("meadowcoin");
-                if (display.slugIcon != null) display.slugIcon.scale = 0.08f;
```
6fears7, line 449: This needs to stay. The normal sprite icon for meadowcoin is massive

OnlineUIComponents/OnlinePlayerDisplay.cs
```diff
@@ -233,32 +214,20 @@ public override void Update()
...
-                    }
-                    else if (owner.PlayerInAncientShelter) slugIcon.SetElementByName("ShortcutAShelter");
-                    else if (owner.PlayerInShelter) slugIcon.SetElementByName("ShortcutShelter");
-                    else if (owner.PlayerInGate) slugIcon.SetElementByName("ShortcutGate");
-                    else if (owner.PlayerConsideredDead) slugIcon.SetElementByName("Multiplayer_Death");
-
-                    else slugIcon.SetElementByName(iconString);
-                    if (slugIcon.element.name == "meadowcoin")
-                    {
-                        this.slugIcon.scale = 0.08f;
```
6fears7, line 256: Scaling needs to stay for meadow coin

OnlineUIComponents/OnlinePlayerDisplay.cs
```diff
@@ -127,25 +122,11 @@ public OnlinePlayerDisplay(PlayerSpecificOnlineHud owner, SlugcatCustomization c
                 }
 
             }
-            if (RainMeadow.isArenaMode(out var arena))
-            {
-                if (arena.reigningChamps != null && arena.reigningChamps.list != null && arena.reigningChamps.list.Contains(player.id))
-                {
-                    this.iconString = "Multiplayer_Star";
```
6fears7, line 134: Might be the diff but I'm not saying this added anywhere

6fears7 (commented):
Just leaving these comments here even though it's a draft for posterity on things that need to remain

Arena/ArenaOnlineGameModes/BaseGameMode.cs
T1mbits: meadowcoin scaling is handled in playericon directly now

OnlineUIComponents/OnlinePlayerDisplay.cs
T1mbits: meadowcoin scaling is handled in playericon directly now

OnlineUIComponents/OnlinePlayerDisplay.cs
T1mbits: pretty sure I removed this because it’s handled in Update() anyways but I’ll double check

T1mbits: correct me if I'm wrong but I'm 99% sure the iconString values set in this block would never be used since the else if block on L219 runs with the exact same conditions and completely overrides iconString

6fears7 (approved):

henpemaz (approved):

PR #1555, opened 2026-08-07
Added Jolly and Exp icon to Lobby Card

EdEnStonne:
Saw the Expedition and Jolly-Coop counted as "MOD" in the lobby card, which looked odd imo, so uh I fixed it.
If @T1mbits doesn't object, I think it's good to go


6fears7:
picture would be good for henpe to see @EdEnStonne

T1mbits:
I’m not the biggest fan of adding jolly coop and expedition to the list. MSC, remix, and watcher are all major content dlcs, hence why they get their own indicators. Expedition in my eyes is more of a gamemode (akin to menagerie) and jolly coop is apparently being marked as safe for client side only in 1.15, so I think there’s really no reason to have them highlighted when they are present in high impact.

EdEnStonne:
It's abit late to show but here's how that looks :
image

EdEnStonne:
> I’m not the biggest fan of adding jolly coop and expedition to the list. MSC, remix, and watcher are all major content dlcs, hence why they get their own indicators. Expedition in my eyes is more of a gamemode (akin to menagerie) and jolly coop is apparently being marked as safe for client side only in 1.15, so I think there’s really no reason to have them highlighted when they are present in high impact.

Oh. I thought it was appropriate for any DLC mods.
Ima get this to draft for now, most likely gonna close if it's not useful.

PR #1559, opened 2026-08-09
Improved Lobby Loading Screen

EdEnStonne:
Simply put, the "Joining Lobby" handling was a mess, and if something went wrong, you have ALOT of chances to end in those weird UI mush of doom and despair.
Timbits already fixed the exit button removing you from the joining dialog box, I went out to fix the rest.

In short :
- Fixed getting still send to meadow/story/arena menu after the joining process errored.
- Changed how lobby joining errors are handled to support more error messages.
- Added a "CANCEL" button to the "Joining Lobby" dialog
- Made the password input send the password when pressing enter
- Put a maximum attempt to join a lobby
- Added a timeout system to OnlineEvents
- Lobby joining will now timeout if the client gets no response (configurable)
- Lobby joining will now stop past 10 attempts
- Added abit more text on the loading screen

image


EdEnStonne:
New errors example :
image

Game/MeadowRemixOptions.cs
```diff
@@ -400,11 +402,17 @@ public override void Initialize()
                 new OpLabel(310, 400f, Translate("Pointing Key")),
                 new OpKeyBinder(PointingKey, new Vector2(310f, 370f), new Vector2(150f, 30f)),
 
-                new OpLabel(10f, 340, Translate($"Player Menu Scroll Speed for Spectate, Story menu, Arena results.  Default: ${ScrollSpeed.Value}"), bigText: false),
+                new OpLabel(10f, 340, Translate($"Player Menu Scroll Speed for Spectate, Story menu, Arena results.  Default: 10"), bigText: false),
```
6fears7: Honestly this is redundant because you can reset your remix file to default anyways. But I'm sure there's other instances where the default is explicitly stated so we can change it later

Online/Resource/Lobby.cs
```diff
@@ -2,6 +2,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
```
6fears7: I assume this is unused

Game/MeadowRemixOptions.cs
```diff
@@ -400,11 +402,17 @@ public override void Initialize()
...
                 new OpTextBox(ScrollSpeed, new Vector2(10, 310), 160f)
                 {
                     accept = OpTextBox.Accept.Float
                 },
+
+                new OpLabel(10f, 270, Translate($"Lobby Joining Timeout (in seconds)"), bigText: false),
+                new OpTextBox(JoiningTimeout, new Vector2(10, 240), 160f)
+                {
+                    accept = OpTextBox.Accept.Int
+                },
```
T1mbits: I'd recommend moving this to General instead of Gameplay since this doesn't have much to do with anything in-game

Game/MeadowRemixOptions.cs
EdEnStonne: I'd remove it, since it's in the label and not in the description like it should be.
It's odd to see in the menu.

Online/Resource/Lobby.cs
EdEnStonne: oh yeah it is, an attempt of mine

Game/MeadowRemixOptions.cs
EdEnStonne: Lemme find some space for it...

Online/OnlinePlayer.cs
```diff
@@ -180,6 +180,10 @@ public void Update()
             bytesOut[nextSnapshotIndex] = 0;
             bytesSnapIndex = nextSnapshotIndex;
 
+            // handle timeout events
+            var timeoutEvents = new Queue<OnlineEvent>(OutgoingEvents.Where(e => e.timeout > 0 && --e.timeout == 0));
```
henpemaz: More reviewing later but an important logical step as a side-effect inside a query is a bit too easy to miss.

invalidunits (commented):
I don't think there's any need for a new GenericResult, especially If we are not sending it.
Set `OnlineEvent.aborted` instead if the event never reaches it's target.

I'm not too sure about timeouts in general, Since If the matchmaking layer hasn't told us that they left the connection, we should still attempt to communicate with them. The cancel button does everything a timeout would need to.

EdEnStonne:
> I don't think there's any need for a new GenericResult, especially If we are not sending it. Set `OnlineEvent.aborted` instead if the event never reaches it's target.

Was about to type how Generic.Error would count as a connection fail until you typed about directly setting `OnlineEvent.aborted`. That sounds like it could work.
Does this bool being true automaticaly dismiss incoming GenericResults from the RPC ? I'm not too sure, but I'm also not too sure with self-sending Timeout.

> I'm not too sure about timeouts in general, Since If the matchmaking layer hasn't told us that they left the connection, we should still attempt to communicate with them. The cancel button does everything a timeout would need to.

I... do kinda agree and disagree with this. Timeout is always nice to have to not feel like waiting forever, but that could allow me to remove all my sketchy system and only leave that.

Online/OnlinePlayer.cs
EdEnStonne: if there's no potential to it, I can revert it.

Game/MeadowRemixOptions.cs
T1mbits: @EdEnStonne I also saw the translation problem and fixed it over [here](https://github.com/henpemaz/Rain-Meadow/pull/1554/changes/9c970889dc6b25ce14229c6fd77e3df22a2961dd#diff-13b93ad8fb8f004136610ea49787da5e31f29b72ba82483694b46929e7d7e517R398), I think this is the preferred method for this kind of thing (though I'm not sure why this is the only label that gets a default value indicator)

EdEnStonne: oh alr, reverting it then

Menu/Menus/LobbySelectMenu.cs
```diff
@@ -313,18 +321,65 @@ public void OnLobbyJoined(bool ok, string error)
...
+            int attempts = OnlineManager.lobby.enumsChecked 
+                ? OnlineManager.lobby.joiningAttempts 
+                : OnlineManager.lobby.enumSyncAttempts;
+            int step = !OnlineManager.lobby.enumsChecked
+                ? 0
+                : OnlineManager.lobby.isRequesting
+                    ? 1
+                    : 2;
+
+            string text = Translate("Joining lobby...") + $" {step}/2";
```
henpemaz: this looks like a debug thing, do we really need to tell the user that?

Menu/Menus/LobbySelectMenu.cs
```diff
@@ -313,18 +321,65 @@ public void OnLobbyJoined(bool ok, string error)
...
+                {
+                    joiningTimeoutCount = TimeoutTicks - FastTimeoutCount; // Making the timeout shorter
+                }
+
+                if (++joiningTimeoutCount >= TimeoutTicks)
+                {
+                    if (OnlineManager.lobby.joiningEvent is not null)
+                    {
+                        RainMeadow.Warn("Joining process it taking too long, timing it out !");
+                        OnlineManager.lobby.joiningEvent.timeout = 1; // safely timeout
```
henpemaz: So you added a whole timeout system for the events and this is all you do with it? All of your timeout logic is in the UI anyway, unless I'm missing something.

Needlessly over-engineered, just abort the event and check for aborted event in the lobby join retry attempt (the result should have a reference to the event that got aborted somewhere iirc)

No need for new result type etc.

henpemaz (changes requested):

Menu/Menus/LobbySelectMenu.cs
EdEnStonne: It always feels nice to have somewhat of a progression indicator.
I could hide it in a Remix setting, but I'm adamant to add more info on the joining steps to the user.

Menu/Menus/LobbySelectMenu.cs
EdEnStonne: it's also in the Lobby.cs file
`joiningEvent = supervisor.InvokeRPC(RequestedLobby, key).Then(ResolveLobbyRequest, TimeoutTicks);`
To give a timeout to the RPC

But honestly ? Yeah, it's overengineered, and I'm afraid it might cause underlying issues as I'm not too sure of what I'm doing.

Will revert it and use the Abort system. I just need to do some more testing about the new, simpler method.

invalidunits (dismissed):

Menu/Menus/LobbySelectMenu.cs
```diff
@@ -313,18 +321,68 @@ public void OnLobbyJoined(bool ok, string error)
...
-        manager.ShowDialog(new NotifyDialog(manager, errorMessage, UIUtils.DIALOG_SIZE));
+        if (error != ERROR_Cancelled) manager.ShowDialog(new NotifyDialog(manager, errorMessage, UIUtils.DIALOG_SIZE));
         RainMeadow.Error(errorMessage);
+
+        // Stop any process/menu switch when an error occur 
+        if (OnlineManager.instance.manager._processSwitchQueue.Count > 0)
+        {
+            RainMeadow.Warn("Found process(es) in queue, clearing it !");
+            OnlineManager.instance.manager._processSwitchQueue.Clear();
+        }
```
henpemaz, line 332: This feels like a hack, What was causing a process switch if not our own lobby handling?

The lobby only goes Active once its Available (onlinemanager.cs:150) and the gamemode only switches to its menu once the lobby is Active (onlinegamemode.cs:282).

Is something else causing process switching when we join lobbies?

Menu/Menus/LobbySelectMenu.cs
```diff
@@ -313,18 +321,68 @@ public void OnLobbyJoined(bool ok, string error)
...
+                if (++joiningTimeoutCount >= TimeoutTicks)
+                {
+                    if (OnlineManager.lobby.joiningEvent is not null)
+                    {
+                        RainMeadow.Warn("Joining process it taking too long, timing it out !");
+                        OnlineManager.lobby.joiningEvent.Abort(); // safely timeout
+                    }
+                    else
+                    {
+                        RainMeadow.Error($"No process running ? What's happening ? Timing it out ! (enumChecked <{OnlineManager.lobby.enumsChecked}>, isRequesting <{OnlineManager.lobby.isRequesting}>, isAvailable <{OnlineManager.lobby.isAvailable}>)");
```
henpemaz: L'espace entre les môts e le ?/! c'est qu'en français, en englais on le met pas!

EdEnStonne: Oh my God the mail being in french jumpscared me.
Force of habit, my bad. I paid attention to that rule for any text that goes in the UI, so there shouldn't be some issues for the english average user.

Will fix it.

Menu/Menus/LobbySelectMenu.cs
EdEnStonne: This bug took me AWHILE to track, so I'll try to break it down with what I remember :

Oftentimes when exiting the "joining lobby" dialog (for example, when it wouldn't stop loading), the process would switch to the arena/story/meadow menu, even though we aren't correctly joined in. The menu would be quite glitchy, often due to a null lobby error, and the game could crash at about any point.
You can still test that bug in 1.14.1, it's fairly easy to replicate.

From what I remember, it was something to do with us recieving the Lobby state but still having no response on the LobbyRequest RPC. The state would mark the lobby as available and queue the new process, bypassing the lobby request entirely. The queued process would then trigger as soon as you closed the dialog box, and making the game believe you joined the lobby when you didn't.

The best solution I came up for this was to clear the queue when the lobby request fails. It's hacky, I agree, but there shouldn't be any reason you'd want to switch process when the lobby request fails, so it shouldn't cause any unforseen issues.
I also don't want to touch the Ressource system too much as this isn't my domain of expertise - this is a quite complex bug that runs deeper that I'm comfortable to reach in.

EdEnStonne: So ok I'm tracing back my steps.

The proces switch is in `OnlineGameMode.ResourceActive`, which is called by `OnlineRessource.Activate`, which is called by the Update of the OnlineManager on lobby when `isAvailable` is true, which is set true by only `lobby.Available`, which can be called by
- `NewOwner` if you are the owner
- `OnlineRessource.State.ReadState` if `isWaitingForState` is true
- `RessolveLobbyRequest`
- `OnlineRessource.Need` to `OnlineRessource.Request`

There is soemthing going wrong over there, although I realized that `OnlineRessource.State.ReadState` is the *normal* route, so I'm unsure how, but the bug is definitely here.

henpemaz (dismissed):

henpemaz (approved):

invalidunits (approved):

PR #1566, opened 2026-08-11
fix: Add null check in Creature.Update hook

One-Letter-Shor:


One-Letter-Shor:
My assumption is that RW is being weird not RM. Hopefully this isn't just covering up a RM bug

6fears7 (approved):

PR #1567, opened 2026-08-12
Made Watcher VFX normal for local player

EdEnStonne:
There UO, that's it. 
Going to sleep now, will test later, feel free to test it before me.


EdEnStonne:
Also I lied again it's actually 3 lines

6fears7 (approved):

PR #1576, opened 2026-08-17
Update CHANGELOG.md so i shouldn't (probably) complain after the release

iulh:
while my changes are made according to my honest opinion, different Point of View, accounting the context i could find and my first-read misunderstandings. i'd like to say that [UO made a very good job by polishing the changelog before the update](https://github.com/henpemaz/Rain-Meadow/pull/1575) so NOBODY can criticize his grammar!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Release 1.15.0
 
 ## Engine
-- Fixed a serialization issue that would break lobbies when Watcher was enabled
+- Fixed a serialization issue that would break lobbies when The Watcher was enabled
 - Added more failsafes to lobby joining, including a configurable timeout
 - Improved some sources of desync (enums)
 - Improved Noodle Fly sync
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -89,14 +89,13 @@
 - Amoebas no longer slow down for dead players or teammates
 - The distortion effect now fades on player death
 - Watcher now shows the same effect as a failed portal opening when attempting to summon an Amoeba during countdown or without enough charge
-- Ripple 9 now makes Watcher fully invisible, leaving only a faint glow behind
+- Ripple 9 now makes Watcher fully invisible to everyone, leaving only a faint glow behind; can be disabled in the Meadow Arena Remix page
```
iulh, line 92: this is a bit tricky spot i'd highlight..

i found nothing better than combine 2 changes as 1, but "...can be disabled in the Meadow Arena Remix page" sounds off to me. where's "Arena Remix page" located i don't get it, did i save the context?

iulh:
other than that i found nothing weird with the changelog, so praise of UO is actually deserved

6fears7 (changes requested):
I like some of the changes, but the indentations were made on purpose to organize sub comments from a parent topic, and that should stay

6fears7:
1.15 is released, thanks for trying though!

iulh:
> the indentations were made on purpose to organize sub comments from a parent topic, and that should stay

i see. couldn't think of the vision so i made what i thought right.

> thanks for trying though!

np really, the focus was on a different view on the text.
the final text is on you anyway

PR #1578, opened 2026-08-17
fix: byte overflows on risky dataset

6fears7:
Arena has a serious risk of exceeding 255 values for map queue (`List<string>`), player scores (`Dict<int, int>`), and trophies (`Dict<int, List<string>`)

I would propose the exchange of adding an additional byte to guard against this.

### DIFF
--- a/Online/Serialization/Serializer.Data.cs
+++ b/Online/Serialization/Serializer.Data.cs
@@ -753,7 +753,7 @@ namespace RainMeadow
 #endif
             if (IsWriting)
             {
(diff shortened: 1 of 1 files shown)


One-Letter-Shor:
We tested it just now and it works great :+1:

henpemaz:
Not happy with blanketing "bigger arrays" all around, let me see if I can make it specific to the fields that need it real quick

6fears7:
opened #1581

PR #1584, opened 2026-08-19
refactor: Use event instead of IChatSubscriber

One-Letter-Shor:
This simple change does a few things:
- Minor styling fixes.
- Uses a proper event instead of a weirdly implemented interface.
- Removes a bug where subscribers are non-deterministically unsubscribed when `IChatSubcriber.Active` is false.
  - This is non-deterministic because it only removes subscribers when a new message is received, which is directly based on user input. (And even if it was deterministic this behavior is not expected)
- Removes pointless enumeration (`.Any()`) and allocation (`.Where().ToList()`) in `ChatLogManager.LogSystemMessage` and `ChatLogManager.LogMessage`.
- Removes the confusing interface mapping in `StoryOnlineMenu` where `Menu.Active` (the base class) satisified `IChatSubscriber.Active` implicitly.

I still need to test tomorrow or the day after, so this is marked as a draft. I'll also keep this as a draft until 1.15 releases so the changes can be added to the changelog.


One-Letter-Shor:
@EdEnStonne You really should get your editor to automatically remove trailing white space. Also I think you over use expression bodied members and nested ternaries a little (Used in places that just makes the code harder to read).

EdEnStonne:
> @EdEnStonne You really should get your editor to automatically remove trailing white space. 

Found the setting, it's all good now.

One-Letter-Shor:
I forgot the changelog lmao

6fears7 (approved):

PR #1585, opened 2026-08-19
EnumSync edge case fix

EdEnStonne:
This is quite the urgent fix, as I've left a big oversight in how ambiguous and missing enums are handled.
Enums send back in full could be counted as missing still, and ambiguous were counted as extra when they were not.

This has to go before 1.15 release. It is a simple fix and shouldn't have unintented side effects.

_Also Shor recommended me to toggle that extra white space fix, since I kept leaving some white speace everywhere, and yeah it was definintely true... So yeah my commit looks abit weird._

### DIFF
--- a/Online/ExtEnumSync.cs
+++ b/Online/ExtEnumSync.cs
@@ -248,10 +248,10 @@ public abstract class CompressedExtEnumBase
             }
             else if (search.Count == 0)
             {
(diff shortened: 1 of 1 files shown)


EdEnStonne:
I'm gonna let @r3nsen test it before it goes out of draft, since they seem to have a consistent setup where it breaks the system.

EdEnStonne:
Confirmed fixed !

henpemaz (approved):

PR #1588, opened 2026-08-20
hotfix: Remove thrown exception, add todo

One-Letter-Shor:


One-Letter-Shor:
This is just a bandaid. UO is looking into the problem and will only merge this if he can't find a solution quickly

PR #1590, opened 2026-08-20
feat(slugicon): support nightcat

T1mbits:
haha whoops definitely didn't forget about this guy.

Just a note, the default colours mirror watcher's default colourscheme but nightcat actually has black eyes in-game by default (which is a little odd). The default colours provided here aren't ever used though since there's no campaign with nightcat and in-game slugicons just borrow SlugcatCustomization colours, so they're there for completeness more than anything


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# Release 1.15.1
+
+## General
+- Removed nightcat's stylish hat in SlugIcon (now uses watcher sprites)
```
6fears7, line 4: and what if watcher isn't enabled?

6fears7 (changes requested):
no changes may be required, but please answer the question

CHANGELOG.md
T1mbits: the sprites are all provided by rain meadow in slugicons.png, watcher dlc is irrelevant to the availability of the slugicon sprites

6fears7 (approved):

PR #1196, opened 2025-12-24
engine: release overworld session

6fears7:
Addresses an issue where the overworld resource is not properly released.

PR #1591, opened 2026-08-20
hotfix: missing online counterparts during transitions

6fears7:
Ok, let me give some background on this before you think "wtf is this? "

In `PlayerOnDie` recently, we used to log an error when the Online counterpart was missing. It swapped to being a thrown exception, and that was causing crashes in arena. 

I found that weird, so I went looking. 

During the WaitLoop we use for Story & Arena, there's a period of time where we're waiting for participants to exit the world and room resources but are still running simulations. That's what the `Room_Preparer` and `Room_Update` freeze hooks address: an inactive world but a room still wanting to simulate while we're transitioning, which pisses things off. 

After that, I noticed that the host never attempts a second Release correctly when participants leave. It'd just try once during a world tear down, give up because participants were still leaving, then would depend on the deadlock timer to escape.

As a result, `Release`() was added to `ParticipantLeft` to re-try. 

Then, when I was testing across all modes, I noticed Watcher was throwing some weird errors about Room states failing updates during warp transitions to another world. 

After investigating, I realized `NotNeeded`() and `Deactivate`() were in the wrong spots. `Deactivate` nulls out resources and was happening first, so room states weren't being properly cleaned up. That's why they're now swapped. 

After doing *that*, I tested Meadow mode and realize the final client in a resource could not transition gates. Just kept getting a failure to release things *we* own.

After more investigation, I realized that with the `NotNeeded` and `Deactivate` properly swapped it exposed a bug where the client was spinning in place trying to address a `Release` they would never complete because we had an `super.isReleasing` check during the Release, but because participants being removed is safe we didn't need this and could defer with a retry outside the Update frame to reduce pressure and stop re-filling the `ProcessSelfEvents` that created a "livelock" scenario where we're constantly screwing ourselves over. 

Benefits of this PR:
- Gate / Warp transitions are happier
- I don't see any repeated room failure releases we used to get in #1196 

==
- Tested across all modes with 3 users in LAN. 
- Included in multiple "real" playtests
- Supersedes #1588 

If this is too much for a hotfix then we can go with the #1588

### DIFF
--- a/Game/RainMeadow.ObjectHooks.cs
+++ b/Game/RainMeadow.ObjectHooks.cs
@@ -24,8 +24,14 @@ namespace RainMeadow
         private void Room_Update_FreezeDeactivatedWorld(On.Room.orig_Update orig, Room self)
         {
             if (OnlineManager.lobby != null
-                && self.abstractRoom.GetResource() is RoomSession rs
-                && !rs.worldSession.isActive) return;
+                && self.abstractRoom.GetResource() is RoomSession rs)
+            {
+                if (!rs.worldSession.isActive)
+                {
+                    RainMeadow.Error($"froze room update for its world, it's never coming back");
+                    return;
+                }
+            }
             orig(self);
         }
 
--- a/Game/RainMeadow.ObjectHooks.cs
+++ b/Game/RainMeadow.ObjectHooks.cs
@@ -28,7 +28,7 @@ namespace RainMeadow
             {
                 if (!rs.worldSession.isActive)
                 {
-                    RainMeadow.Error($"froze room update for its world, it's never coming back");
+                    RainMeadow.Debug($"froze room update for its world, it's never coming back");
                     return;
                 }
             }
--- a/Game/RainMeadow.ObjectHooks.cs
+++ b/Game/RainMeadow.ObjectHooks.cs
@@ -14,9 +14,19 @@ namespace RainMeadow
         private void ObjectHooks()
         {
             IL.Room.Update += Room_Update;
+            On.Room.Update += Room_Update_FreezeDeactivatedWorld;
 
             // IL.ScavengerOutpost.ctor += ScavengerOutpost_ctor1; ;
             On.LobeTree.ValidSpawnPos += LobeTree_ValidSpawnPos;
+        }
+
+        // A deactivated world deregistered all its entities, but its rooms update until the process switches.
+        private void Room_Update_FreezeDeactivatedWorld(On.Room.orig_Update orig, Room self)
+        {
+            if (OnlineManager.lobby != null
+                && self.abstractRoom.GetResource() is RoomSession rs
+                && !rs.worldSession.isActive) return;
+            orig(self);
         }
 
         private bool LobeTree_ValidSpawnPos(On.LobeTree.orig_ValidSpawnPos orig, Room room, IntVector2 pos, List<Vector2> lobeTreeSpawnLocsSoFar)
--- a/Game/RainMeadow.PlayerHooks.cs
+++ b/Game/RainMeadow.PlayerHooks.cs
@@ -2087,6 +2087,7 @@ public partial class RainMeadow
         if (OnlineManager.lobby.gameMode is MeadowGameMode) return; // do not run
 
         OnlineCreature? onlineCreature = self.abstractCreature.GetOnlineCreature();
+        // player quit while a death event was running, or wait loop is processing
         if (onlineCreature is null)
         {
             orig(self);
--- a/Game/RainMeadow.PlayerHooks.cs
+++ b/Game/RainMeadow.PlayerHooks.cs
@@ -139,7 +139,7 @@ public partial class RainMeadow
         try
         {
             ILCursor cursor = new ILCursor(il);
-            if (cursor.TryGotoNext(MoveType.Before, 
+            if (cursor.TryGotoNext(MoveType.Before,
                 x => x.MatchCallvirt(typeof(FNode).GetProperty(nameof(FNode.isVisible)).GetSetMethod())))
             {
                 cursor.MoveAfterLabels();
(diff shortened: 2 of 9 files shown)


Game/RainMeadow.ObjectHooks.cs
```diff
@@ -14,11 +14,21 @@ public partial class RainMeadow
...
         {
             IL.Room.Update += Room_Update;
+            On.Room.Update += Room_Update_FreezeDeactivatedWorld;
 
             // IL.ScavengerOutpost.ctor += ScavengerOutpost_ctor1; ;
             On.LobeTree.ValidSpawnPos += LobeTree_ValidSpawnPos;
         }
 
+        // A deactivated world deregistered all its entities, but its rooms update until the process switches.
+        private void Room_Update_FreezeDeactivatedWorld(On.Room.orig_Update orig, Room self)
```
6fears7, line 24: Prevents an NRE with AbstractEntity_Destroy and online counterpart error with PlayerOnDie during those transitions. No online counter parts are available during the transition so they'd throw otherwise

GameModes/ArenaCompetitiveGameMode.cs
```diff
@@ -710,10 +710,6 @@ public void RestartGame(RainWorldGame game, ArenaGameSession arenaSession)
 
             AbstractRoom absRoom = game.world.abstractRooms[0];
             Room room = absRoom.realizedRoom;
-            WorldSession worldSession = WorldSession.map.GetValue(
```
6fears7, line 713: Let me quit to the Arena menu you bastard

Online/Resource/OnlineResource.cs
```diff
@@ -383,6 +383,13 @@ private void ParticipantLeft(OnlinePlayer participant)
                 if (isWaitingForState) isWaitingForState = false;
                 PerformRequests();
             }
+            else if (!isNeeded && isAvailable && canRelease)
```
6fears7: specifically a waitloop requirement. Watcher & Meadow Mode don't use the waitloop, just arena and other story modes. Because we spin while we wait for others to leave in order to assist with processing those temporarily subscribed but missing when moving across worlds, we need another opportunity to Release beyond the original NotNeeded call that kicks off the waitloop

Online/Resource/WorldSession.cs
```diff
@@ -60,11 +60,6 @@ System.Action onComplete
                         var remainingPlayers = session
                             .participants.Where(x => x != player)
                             .ToList();
-                        if (session.overworldSession != null)
```
6fears7, line 63: I'm not sure why I originally added this because dropping the overworld ownership can only lead to misery

Game/RainMeadow.ObjectHooks.cs
```diff
@@ -14,11 +14,27 @@ public partial class RainMeadow
...
+        {
+            if (OnlineManager.lobby != null
+                && self.abstractRoom.GetResource() is RoomSession rs)
+            {
+                if (!rs.worldSession.isActive)
+                {
+                    RainMeadow.Error($"froze room update for its world, it's never coming back");
```
6fears7: there should never be a scenario where we're trapped with a room never updating after the first frame of a worldSession no longer active and no longer moving into a new world, but if so, log it so it's not silent

Online/Resource/WorldSession.cs
```diff
@@ -29,7 +29,7 @@ System.Action onComplete
             bool cleanupTriggered = false;
 
             session.transitionInProgress = true;
-            bool canSkipWaitLoop = OnlineManager.lobby.gameMode is MeadowGameMode || RainMeadow.isStoryMode(out var story) && story.currentCampaign == Watcher.WatcherEnums.SlugcatStatsName.Watcher;
```
invalidunits, line 32: Any reason to remove this?

Online/Resource/OnlineResource.cs
```diff
@@ -385,6 +387,16 @@ private void ParticipantLeft(OnlinePlayer participant)
...
+
+            // If NotNeeded was  turned down because a
+            // participant was still here or hasn't ackd the latest lease nothing ever tries
+            // again and whoever is waiting for this resource to empty out waits forever. Truly doom
+            // Runs after so nobody gets handed an already-unavailable resource
+            // with no supervisor (during wait loops)
+            if (!participant.isMe && !isNeeded && isAvailable && canRelease)
+            {
+                Release();
+            }
```
invalidunits: Check if we're pending first, otherwise it'll throw an exception.
```suggestion
            if (!participant.isMe && !isNeeded && !isPending && isAvailable && canRelease)
            {
                Release();
            }
```

invalidunits (dismissed):

Online/Resource/WorldSession.cs
6fears7: The changes made here enable the watcher wait loop to work correctly and drastically speeds up warp times. It was exempt originally because of release handling being funky

One-Letter-Shor:
I playtested this with many players and it seems to completely have removed the problem.

One thing: I do think the log message "froze room update for this world, it's never coming back" should be info or debug rather than error.

Game/RainMeadow.PlayerHooks.cs
```diff
@@ -2087,9 +2087,13 @@ private void PlayerOnDie(On.Player.orig_Die orig, Player self)
         if (OnlineManager.lobby.gameMode is MeadowGameMode) return; // do not run
 
         OnlineCreature? onlineCreature = self.abstractCreature.GetOnlineCreature();
+        // player quit while a death event was running, or wait loop is processing
         if (onlineCreature is null)
-            throw new InvalidProgrammerException("Player doesn't have an OnlineCreature counterpart!!");
-
+        {
```
6fears7, line 2092: not really useful to throw here if we can't find the counterpart, just run local sim and return

henpemaz (approved):
Good catch! Happy to include in the hotfix if we get some playtesting in

invalidunits (approved):

PR #1592, opened 2026-08-20
fix(extenumsync): Load All ExtEnum left in static constructors

EdEnStonne:
This is an attempt at also getting rid of the enum desync issues that are left (or most of it), this one being abit more forceful.

The first time you join a lobby, Rain Meadow will scan all assemblies from mods and run the cctor of any classes that have some ExtEnum in a static variable (if it wasn't run already).

This is to prevent lazy cctor loading to mess up with the enum mapping system. If an enum is added too late (post lobby joining), it won't be added to the map unless you rejoin the lobby, and any RPC/State that references this object will throw an exception (most likely a NullException).

The scan is restained to ExtEnum that are in a static field/propriety, and are mapped by Rain Meadow. The classes with types that cannot be loaded (soft-dependancies) will be ignored.

The process takes around ~25ms with my i9-119900H on modpacks with big mods, so the joining time shouldn't be affected by much.


EdEnStonne:
Still a draft till a get more testing and judgement on it.

EdEnStonne:
It's functional, so all it's left is to jugde if it's a good idea

One-Letter-Shor:
I think this PR has some major problems

Firstly, I launched the game using Stopwatch to measure how long the loop took, and the result was ~220ms with 17 mods, and ~260ms with 60 mods on a 7700x. (This is not a proper test but it should give you an idea of how long it takes)
Unless this isn't possible, you should only be enumerating mod assemblies. I'd assume this can be done by reading from enabledMods.txt or getting data from BepInEx.

Secondly, even if this wasn't slow, this is very likely to cause issues with mods that aren't doing anything wrong because you are changing the behavior of their types. And anyway, if other mods *were* doing something wrong, that's not Rain Meadow's responsibility to fix.

EdEnStonne:
> Firstly, I launched the game using Stopwatch

Oh geez, I did not know it was this slow. Seemed fast enough on my PC, so I didn't bother mesuring loading time.

> Unless this isn't possible, you should only be enumerating mod assemblies.

If it's doable, I'll definitely try it.

> Secondly, even if this wasn't slow, this is very likely to cause issues with mods that aren't doing anything wrong because you are changing the behavior of their types.

This is the part where I'm not fully sure if that feature is a good idea, and I'm glad you commented on it.
I did make sure to put it on PostModsInit to avoid messing with the startup of a mod, and from what I could read, it's quite unlikely to mess with other mods.
The mods would need to have a class that has an ExtEnum with already registered elements, which happens to be statically initialised, which happens to not be initialized already after all mods loaded, which happens to have statically initialized element that have to be lazy initialized at a precise point otherwise it'd do something weird.
Any other class that don't fit that requirement should be left alone, or be completely unaffected by it, since it calls the cctor on classes that haven't called it yet and will call it sooner or later.

> And anyway, if other mods were doing something wrong, that's not Rain Meadow's responsibility to fix.

Meadow has a big sensibility on ExtEnum being loaded and sorted, and errors of that sort are quite complicated to track down. Statically initializing stuff works well for those mods, but Rain Meadow's sensibility can mess it up unexpectedly.
In my opinion, if it's possible to be fixed, that's worth a try. I'm not sure expecting every mod to have it loaded specifically when *we* want is the good approach.

HOWEVER, it's all my thoughts, and if it's that horrible it's better to drop it all now.

EdEnStonne:
Testing time on the hook, with only mods assemblies loaded (minus Rain Meadow) :
- **Empty** (Remix and Modpack) : 2ms, no assembly touched.
- **All DLC** : 2ms, no assembly touched.
- **Region Kit** : 20ms, 2ms for POM, 14ms for RegionKit.
- **27 mods** (including RegionKit and ALOT of new slugcats) : 28ms, 11 assemblies that has some types cctor run pre-emptively.

My processor is an i9-119900H. 
For reference, the Rain Meadow hook `RainWorld_OnModsInit` (which doesn't include the assembly load in my PR) took 21.05s in my last test to run.

I'll clean that up and push this in my PR.

Edit : will reinclude Rain Meadow in that. Rain Meadow takes about 15ms to check all types.

EdEnStonne:
After talking with Shor, I wanna fix multiple problems before getting it to review again :
- Loading cctor early is a hell of a risk. I'm going to delay it further, but it still doesn't leave the risk out.
- Doing nothing would still be an issue, as late ExtEnum added will mess up the ExtEnum mapping (RegionKit has that issue for example).
- I need to find a way to not load cctors of types with unloaded dependancies.

EdEnStonne:
After some research and testing :
- The scanning has been delayed to the first time you join/create a lobby, which is the latest I can make it.
- The scanning has been restrained to only ExtEnums that are mapped by Rain Meadow.
- The potential risk is still there, but at this point in time I don't see any reason why a cctor would need to be loaded even later. All my test have shown no issues.
- While I cannot check if a file/class is dedicated for soft compatibility, Meadow has already a function to load safely the classes that have all their type defined. This is the best I can go in terms of ignoring classes serving soft-compatibility.
- Scaning time is around 25ms for big modpacks ! (which is already insane to run a meadow lobby with, but hell)
- To respond to @One-Letter-Shor, I recognize that the ideal solution would be that every mods on the workshop doesn't use cctor to register ExtEnum, but I do find it idealistic to bet on that idea when Meadow's ExtEnum to byte system is the one being sensible to ExtEnum order change, especially when big mods like RegionKit do this (it's probably worth contacting them about it though).

I'll also updated the description.

henpemaz:
I would rather trust load order and mods correctly being added to high-impact to determine the order of enums. On top of that we already have an enum-sync layer that can detect problems. On-top-of-on-top-of-that, this? Seems redundant, plus modifying the loading behaviour of other mods is problematic and can cause hard to debug issues.

EdEnStonne:
> I would rather trust load order and mods correctly being added to high-impact to determine the order of enums. On top of that we already have an enum-sync layer that can detect problems. On-top-of-on-top-of-that, this? Seems redundant, plus modifying the loading behaviour of other mods is problematic and can cause hard to debug issues.

It's not to change load order. Some mods loads their enums after joining a lobby (issue rose with RegionKit having an object enum registered only after a story session was started), and that can cause errors with the enum mapping and order.

This will guarantee all ext enum that we care about and left in cctor will be loaded when you first join a lobby.

henpemaz:
Then submit a PR to regionkit

EdEnStonne:
I've already send an issue report about this to them.
Wherever they care to make a fix is their choice, I have enough on my plate already.

PR #1595, opened 2026-08-20
hotfix: Chat Clear Option (1.15.1 / 1.6)

EdEnStonne:
Now that chat finally stays after every round, people complain about the clutter.
So I added a Remix option to fix it.

Tested on public lobbies and it work.


OnlineUIComponents/ChatHud.cs
```diff
@@ -51,6 +51,30 @@ public ChatHud(RoomCamera camera) : base(RMOverlayHUD.GetOverlay())
...
+            else if (RainMeadow.rainMeadowOptions.ClearChatEveryRound.Value == RainMeadowOptions.ChatClear.System)
+            {
+                ClearMessages(x => ChatLogManager.IsUserSystemSignature(x.Item1));
+            }
+            else if (RainMeadow.rainMeadowOptions.ClearChatEveryRound.Value == RainMeadowOptions.ChatClear.All)
+            {
+                ClearMessages(x => true);
+            }
+        }
+        public void ClearMessages(Predicate<(string, string)> match)
```
One-Letter-Shor: FYI, you should be giving these tuple items names: `(string User, string Message)`

One-Letter-Shor: To be clear: this isn't a horrible place to not give them names but it does confirm to the caller that they are passing in the correct thing

EdEnStonne: I didn't even know you could do that.
This is gamechanger for me.

6fears7 (dismissed):

6fears7:
Lmk if you want merge

EdEnStonne:
@6fears7 should be all good now !

6fears7 (approved):

PR #1601, opened 2026-08-21
hotfix : Added RPC owner check (1.15.1/1.16.0)

EdEnStonne:
A certain XuanZeBuFeng notified us on some security issues in our RPCs, so I fixed them the best I could.

- Added owner check to RPC.ExitToGameModeMenu
- Added owner check (room or object) to OnlinePhysicalObject.Explode

Was able to explode as Arti and exit to arena menu in local testing

### DIFF
--- a/Arena/ArenaRPCs.cs
+++ b/Arena/ArenaRPCs.cs
@@ -305,9 +305,9 @@ namespace RainMeadow
             }
         }
         [RPCMethod]
-        public static void Arena_EndSessionEarly()
+        public static void Arena_EndSessionEarly(RPCEvent rpc)
         {
-            if (RainMeadow.isArenaMode(out var arena))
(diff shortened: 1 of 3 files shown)


6fears7:
This might be a use case for RPCSecurity property: https://github.com/henpemaz/Rain-Meadow/blob/main/Online/Resource/Lobby.cs#L159

EdEnStonne:
oh jolly that's a great idea

EdEnStonne:
> This might be a use case for RPCSecurity property: https://github.com/henpemaz/Rain-Meadow/blob/main/Online/Resource/Lobby.cs#L159

After some digging, it seems that RPCSecurity.Owner only works if the RPC is from a ressource or an online entity, not from a static method.

Opening it to review !

6fears7 (approved):

henpemaz (approved):

PR #1586, opened 2026-08-19
hotfix: collection modification error OnlinePhysicalObject.ResolveRequest

Quaxly:
will tack on hot fixes as I make them, feel free to merge whenever.

- [x] Fixed a collection modification error that would occur if OnlinePhysicalObject.ResolveRequest failed.

PR #1602, opened 2026-08-21
hotfix : Fixed more collection modification error (1.15.1)

EdEnStonne:
Those are not the same as #1586

- Fixed RemoveEntityFromGame throwing when modifying the list count
- Fixed AbstractCreatureFromString (probably) throwing when modifying the list count

Was able to run arena and leave with an item in hand without LeaveImp throwing.

### DIFF
--- a/Online/Entity/OnlineCreature.cs
+++ b/Online/Entity/OnlineCreature.cs
@@ -130,9 +130,10 @@ namespace RainMeadow
             EntityID id = EntityID.FromString(array[1]);
             AbstractCreature abstractCreature = new AbstractCreature(world, StaticWorld.GetCreatureTemplate(type), null, pos, id);
 
-            foreach (var item in abstractCreature.stuckObjects.ToArray()) // Some (dropbug) creatures spawn with random items attached
+            var stuckObjects = abstractCreature.stuckObjects.ToArray();
+            for (int i = stuckObjects.Length - 1; i >= 0 ; i--)
             {
-                item.Deactivate();
+                stuckObjects[i].Deactivate(); // Some (dropbug) creatures spawn with random items attached
             }
 
             abstractCreature.state.LoadFromString(Regex.Split(array[3], "<cB>"));
@@ -305,7 +306,7 @@ namespace RainMeadow
                 RainMeadow.Error($"{this} is to early to spit out of shortcut.");
                 return;
             }
-            
+
             if (this.realizedCreature is null) {
                 this.creature.Realize();
             }
--- a/Online/Entity/OnlinePhysicalObject.cs
+++ b/Online/Entity/OnlinePhysicalObject.cs
@@ -211,7 +211,7 @@ namespace RainMeadow
                 creatingRemoteObject = oldCreatingRemoteObject;
                 throw;
             }
-            creatingRemoteObject = oldCreatingRemoteObject; 
+            creatingRemoteObject = oldCreatingRemoteObject;
 
 
             realized = initialState.realized;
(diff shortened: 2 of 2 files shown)


Online/Entity/OnlinePhysicalObject.cs
```diff
@@ -446,18 +447,19 @@ public void RemoveEntityFromGame(bool onlineaware = true)
...
+                    po.grabbedBy[i].Release();
                 }
 
                 if (po is Creature creature)
                 {
                     if (creature.grasps != null)
                     {
-                        foreach (Creature.Grasp grabbing in creature.grasps.OfType<Creature.Grasp>())
+                        var grabbedObjects = creature.grasps.OfType<Creature.Grasp>().ToArray();
+                        for (int i = grabbedObjects.Length - 1; i >= 0; i--)
```
One-Letter-Shor, line 460: This looks like unnecessary allocation (.ToArray()) and a redundant OfType<>() call (it behaves as a null check)

does this not work?
```cs
for (int i = creature.grasps.Length - 1; i >= 0; i--)
    creature.grasps[i]?.Release();
```

One-Letter-Shor: And if you do use .ToArray() you can just use a regular foreach loop because it's a copy of creature.grasps (won't be modified by Grasp.Release()

One-Letter-Shor: Also the other loops have the same problem, I think .ToArray() shouldn't be used in any of these

EdEnStonne: will check that then, the OfType threw me off but I figured it had a use.
if it doesn't, your solution would absolutely work

One-Letter-Shor: If you install resharper (integrated into rider) it should warn you about these types of problems
image

Online/Entity/OnlineCreature.cs
```diff
@@ -130,9 +130,10 @@ public static AbstractCreature AbstractCreatureFromString(World world, string cr
             EntityID id = EntityID.FromString(array[1]);
             AbstractCreature abstractCreature = new AbstractCreature(world, StaticWorld.GetCreatureTemplate(type), null, pos, id);
 
-            foreach (var item in abstractCreature.stuckObjects.ToArray()) // Some (dropbug) creatures spawn with random items attached
+            var stuckObjects = abstractCreature.stuckObjects.ToArray();
+            for (int i = stuckObjects.Length - 1; i >= 0 ; i--)
```
henpemaz, line 134: Any reason why you're duplicating the array AND iterating backwards? Those are two different solutions for similar problems.

This iterator was already not throwing errors, because the iterator is not modified inside the loop (it's a duplicate of the array that is being modified). It's pointless to iterate on it backwards because the list iterated on is not being modified.

henpemaz:
I think you found a hammer and now you want to hit everything with a hammer. Stop, take a break, and think about what problem you are solving. Closing this, open a cleaner PR if there's any of these loops that actually needs a patch.

Online/Entity/OnlineCreature.cs
EdEnStonne: Oh Shor noted this too but I for got to change it back

EdEnStonne:
> I think you found a hammer and now you want to hit everything with a hammer. Stop, take a break, and think about what problem you are solving. Closing this, open a cleaner PR if there's any of these loops that actually needs a patch.

Pretty sure I opened this PR because I saw a foreach loop throwing.
Will nail only the one I'm sure I found and open a new one

PR #1604, opened 2026-08-21
hotfix : Fixed Den not opening when someone left (1.15.1)

EdEnStonne:
unsure why that bug is back, but I basically :

- Removed the realized Player in the ArenaGameSession.Players list when leaving
- Stop counting players with no owners as active
- Opened den in Competitive and TeamBattle when there's 1 player and the countdown is over

Tested in local (3 players) and everything seems to work fine


6fears7:
General overview looks good, will let @One-Letter-Shor add any additional comments

One-Letter-Shor:
Yeah it looks good I think

6fears7 (approved):

PR #1606, opened 2026-08-21
hotfix : Sent the quit RPC to everyone and not just the host (1.15.1)

EdEnStonne:
Changes :

- Woopsies, fixed back #1604 who was only working for the host
- Created the virtual externalArenaGameMode.GetPlayerStillActive which is used in the hook On.ArenaGameSession.PlayersStillActive (fixed some bugs about it too)
- used self.gameSession.thisFrameActivePlayers to check standing players in gamemodes (why didn't we do that before ?)
- Fixed a huge might be null warning in ExternalArenaGameMode, don't mind it.

### DIFF
--- a/Arena/ArenaHooks.cs
+++ b/Arena/ArenaHooks.cs
@@ -2371,7 +2371,7 @@ namespace RainMeadow
         {
             if (isArenaMode(out var arenaOnline))
             {
-                return arenaOnline.externalArenaGameMode.GetPlayerStillActive(self, false).Count;
+                return arenaOnline.externalArenaGameMode.GetActivePlayerACs(arenaOnline, self).Count;
             }
             else
             {
--- a/Arena/ArenaHooks.cs
+++ b/Arena/ArenaHooks.cs
@@ -2221,32 +2221,35 @@ namespace RainMeadow
                     {
                         return;
                     }
-                    if (OnlineManager.lobby.isOwner)
+
+                    for (int i = 0; i < arena.arenaSittingOnlineOrder.Count; i++)
                     {
-                        for (int i = 0; i < arena.arenaSittingOnlineOrder.Count; i++)
+                        OnlinePlayer? onlinePlayer = ArenaHelpers.FindOnlinePlayerByLobbyId(
+                            arena.arenaSittingOnlineOrder[i]
+                        );
+
+                        if (onlinePlayer != null && !onlinePlayer.isMe)
                         {
-                            OnlinePlayer? onlinePlayer = ArenaHelpers.FindOnlinePlayerByLobbyId(
-                                arena.arenaSittingOnlineOrder[i]
-                            );
-                            if (onlinePlayer != null && !onlinePlayer.isMe)
+                            if (OnlineManager.lobby.isOwner)
                             {
                                 onlinePlayer.InvokeOnceRPC(ArenaRPCs.Arena_EndSessionEarly);
                             }
+                            else
+                            {
+                                onlinePlayer.InvokeOnceRPC(
+                                    ArenaRPCs.Arena_RemovePlayerWhoQuit,
+                                    OnlineManager.mePlayer
+                                );
+                            }
                         }
-                        self.manager.RequestMainProcessSwitch(
-                            ProcessManager.ProcessID.MultiplayerResults
-                        );
                     }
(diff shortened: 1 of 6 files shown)


Arena/ArenaHooks.cs
```diff
@@ -2221,32 +2221,35 @@ string message
...
+
+                        if (onlinePlayer != null && !onlinePlayer.isMe)
                         {
-                            OnlinePlayer? onlinePlayer = ArenaHelpers.FindOnlinePlayerByLobbyId(
-                                arena.arenaSittingOnlineOrder[i]
-                            );
-                            if (onlinePlayer != null && !onlinePlayer.isMe)
+                            if (OnlineManager.lobby.isOwner)
                             {
                                 onlinePlayer.InvokeOnceRPC(ArenaRPCs.Arena_EndSessionEarly);
```
6fears7, line 2235: we dont need to invoke right? just call ArenaRPCs.Arena_EndSessionEarly
```suggestion
ArenaRPCs.Arena_EndSessionEarly();
```

EdEnStonne: we need to invoke it for others though

EdEnStonne:
Putting it as draft till it's tested and proved to work again.

Arena/ArenaOnlineGameModes/BaseGameMode.cs
```diff
@@ -1574,6 +1566,74 @@ public virtual bool On_ArenaSitting_PlayerSittingResultSort(
...
+                }
+
+                if (countPlayerAsActive)
+                {
+                    activePlayers.Add(self.Players[i]);
+                }
+            }
+
+            return activePlayers;
+        }
```
One-Letter-Shor: To be honest this method is designed very weirdly.
- Method name is weird (contains 'Still'), is not plural, and does not reflect the player type (contains 'player' instead of 'playerAC')
- bool parameter is named weirdly (contains 'count' instead of 'include'), is negative,  and has a contraction (dont)
- Unnecessary for loops
- Indexing each time to access player ACs and arena players
- Safe casting then using `!` instead of hard casting (this is bad because the intent isn't as clear, and if it does fail you get a NRE instead of cast exception, plus safe casts are technically more expensive than hard casts)
- Unnecessary local bool when `continue`s could be used instead
- Spam of allman braces on ifs that make it harder to read (you should either merge into one if, or use whitespace-only ifs)

Also, I *think* this indirectly fixes a bug. IIRC, PlayerState.playerNumber is always the local player number (so, in arena, it's always 0). Only ArenaPlayer.playerNumber is different between online players.
You may need to check that the arena player is always found, I made the assumption that if the online player can be found from the player AC, the arena player will always be found from the online player.

```suggestion
        public virtual List<AbstractCreature> GetActivePlayerACs(
            ArenaOnlineGameMode arenaOnline,
            ArenaGameSession arenaSession,
            bool includeSandboxLosers = true)
        {
            List<AbstractCreature> activePlayers = [];

            foreach (AbstractCreature playerAC in arenaSession.Players)
            {
                OnlinePlayer? onlinePlayer = playerAC.GetOnlineCreature()?.owner;
                if (onlinePlayer is null)
                    continue;

                if (!playerAC.state.alive
                    || arenaSession.exitManager?.IsPlayerInDen(playerAC) == true
                    || ((Player)playerAC.realizedCreature)?.dangerGrasp is not null)
                {
                    continue;
                }

                ArenaSitting.ArenaPlayer arenaPlayer = ArenaHelpers.FindArenaPlayerByOnlinePlayer(
                    arenaOnline,
                    arenaSession.arenaSitting,
                    onlinePlayer
                )!;

                if (playerAC.Room == arenaSession.game.world.offScreenDen
                    && arenaPlayer.hasEnteredGameArea
                    && !includeSandboxLosers && arenaPlayer.sandboxWin < 0)
                {
                    continue;
                }

                activePlayers.Add(playerAC);
            }

            return activePlayers;
        }
```

EdEnStonne: There's just an or missing but that seems good.
Will lead a playtest and hopefully that should give great results

EdEnStonne:
Tested and functional !

6fears7 (approved):

PR #1611, opened 2026-08-22
Chat Emotes

invalidunits:
image

### DIFF
--- a/Menu/Components/ChatEmote.cs
+++ b/Menu/Components/ChatEmote.cs
@@ -5,29 +5,27 @@ namespace RainMeadow.UI.Components
 {
     class ChatEmote : RectangularMenuObject, ButtonScroller.IPartOfButtonScroller
     {
-        public MeadowProgression.Emote emote; 
         public float Alpha { get; set; }
         public Vector2 sourceSize = Vector2.zero;
         public Vector2 Pos { get => pos; set => pos = value; }
-        public Vector2 Size { get => size; set {}  }
+        public Vector2 Size { get => size + Margin; set {}  }
         public Vector2 Margin = new Vector2(40, 10);
-
+        
         private FSprite[] sprites;
 
-        public ChatEmote(MeadowProgression.Emote emote, Menu.Menu menu, MenuObject owner, Vector2 pos) : 
+        public ChatEmote(MeadowProgression.Character character, MeadowProgression.Emote emote, Menu.Menu menu, MenuObject owner, Vector2 pos) : 
             base(menu, owner, pos, Vector2.zero)
         {
             if (!Futile.atlasManager.DoesContainAtlas("emotes_common"))
             {
                 HeavyTexturesCache.futileAtlasListings.Add(Futile.atlasManager.LoadAtlas("illustrations/emotes/emotes_common").name);
             }
-            var chardata = MeadowProgression.characterData[MeadowProgression.Character.Slugcat];
+            var chardata = MeadowProgression.characterData[character];
             if (!Futile.atlasManager.DoesContainAtlas(chardata.emoteAtlas))
             {
                 HeavyTexturesCache.futileAtlasListings.Add(Futile.atlasManager.LoadAtlas("illustrations/emotes/" + chardata.emoteAtlas).name);
             }
 
-            this.emote = emote;
             this.sprites = new FSprite[2];
             string emote_sprite = (emote.value.StartsWith("emote") ? chardata.emotePrefix + emote.value : emote.value).ToLowerInvariant();
             string background_string = emote.value.StartsWith("emote") ? "emote_background" : "symbols_background";
@@ -66,14 +64,16 @@ namespace RainMeadow.UI.Components
     
     }
 
-    class ChatSpace : RectangularMenuObject, ButtonScroller.IPartOfButtonScroller
+    class ChatEmoteSpace : RectangularMenuObject, ButtonScroller.IPartOfButtonScroller
     {
-        public ChatSpace(Menu.Menu menu, MenuObject owner, Vector2 pos, Vector2 size) : base(menu, owner, pos, size)
+        public ChatEmote origin;
+        public ChatEmoteSpace(Menu.Menu menu, MenuObject owner, Vector2 pos, Vector2 size, ChatEmote origin) : base(menu, owner, pos, size)
         {
+            this.origin = origin;
         }
 
         public float Alpha { get; set; }
         public Vector2 Pos { get; set; }
-        public Vector2 Size { get; set; }
+        public Vector2 Size { get => new Vector2(origin.Size.x, size.y); set => size = value; }
     }
 }
--- a/Menu/Components/ChatEmote.cs
+++ b/Menu/Components/ChatEmote.cs
@@ -1,0 +1,79 @@
+using Menu;
+using UnityEngine;
+
+namespace RainMeadow.UI.Components
+{
+    class ChatEmote : RectangularMenuObject, ButtonScroller.IPartOfButtonScroller
+    {
+        public MeadowProgression.Emote emote; 
+        public float Alpha { get; set; }
+        public Vector2 sourceSize = Vector2.zero;
+        public Vector2 Pos { get => pos; set => pos = value; }
+        public Vector2 Size { get => size; set {}  }
+        public Vector2 Margin = new Vector2(40, 10);
+
+        private FSprite[] sprites;
+
+        public ChatEmote(MeadowProgression.Emote emote, Menu.Menu menu, MenuObject owner, Vector2 pos) : 
+            base(menu, owner, pos, Vector2.zero)
+        {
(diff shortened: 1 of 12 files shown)


Menu/Components/ChatEmote.cs
```diff
@@ -0,0 +1,79 @@
...
+        {
+            if (!Futile.atlasManager.DoesContainAtlas("emotes_common"))
+            {
+                HeavyTexturesCache.futileAtlasListings.Add(Futile.atlasManager.LoadAtlas("illustrations/emotes/emotes_common").name);
+            }
+            var chardata = MeadowProgression.characterData[character];
+            if (!Futile.atlasManager.DoesContainAtlas(chardata.emoteAtlas))
+            {
+                HeavyTexturesCache.futileAtlasListings.Add(Futile.atlasManager.LoadAtlas("illustrations/emotes/" + chardata.emoteAtlas).name);
+            }
```
T1mbits, line 27: `LoadAtlas()` already does a `DoesContainAtlas()` call. I'm also not sure what the reasoning behind using `HeavyTexturesCache` here is and maybe it's different for cross-process and in-game sprites or something but the manual addition to the cache guarded by whether or not the atlas manager has loaded it or not is a bit fishy to me. I've never had a problem in-menus with not loading sprites into `HeavyTexturesCache`

Menu/Components/ChatEmote.cs
```diff
@@ -0,0 +1,79 @@
...
+{
+    class ChatEmote : RectangularMenuObject, ButtonScroller.IPartOfButtonScroller
+    {
+        public float Alpha { get; set; }
+        public Vector2 sourceSize = Vector2.zero;
+        public Vector2 Pos { get => pos; set => pos = value; }
+        public Vector2 Size { get => size + Margin; set {}  }
+        public Vector2 Margin = new Vector2(40, 10);
+        
+        private FSprite[] sprites;
```
T1mbits, line 14: Consider using `RainMeadow.UI.Components.Base.PositionedSprite`, it handles placement of individual `FSprite`s for you and you can just treat it as another menu object with pos aligned to the bottom left corner. I think you'll have to implement `IPartOfButtonScroller`, but that should be like 3 lines. You can remove the `GrafUpdate()` and potentially the `RemoveSprites()` logic if you do this.

Also, something like
```csharp
public PositionedSprite backgroundSprite,
    emoteSprite;
```
is preferred over an arbitrary array of `FSprite[]` which always holds 2 elements

T1mbits (changes requested):

Menu/Components/ChatEmote.cs
```diff
@@ -0,0 +1,79 @@
...
+        public ChatEmoteSpace(Menu.Menu menu, MenuObject owner, Vector2 pos, Vector2 size, ChatEmote origin) : base(menu, owner, pos, size)
+        {
+            this.origin = origin;
+        }
+
+        public float Alpha { get; set; }
+        public Vector2 Pos { get; set; }
+        public Vector2 Size { get => new Vector2(origin.Size.x, size.y); set => size = value; }
+    }
+}
```
One-Letter-Shor, line 79: There are a few issues with this class
- Block scoped namespace (adds unnecessary indents)
- Types had no access modifiers so they defaulted to internal. This seems like a mistake because many of the members were marked as public. Also you should never make people have to guess about the accessibility.
- Array of sprites instead of two fields (you have to use magic numbers to get the sprites from the array)
- Source code frequently breaks 120 characters
- Local variables are snake case instead of camel case, and some do not have casing at all
- The naming variables for the sprite name was just called 'XXXSprite' instead of 'XXXSpriteName'
- A field initializer was ignored (set in the ctor)
- Many lines have trailing whitespace, you should configure your editor to automatically trim whitespace

Also, there are two types in this one file. ChatEmoteSpace should be moved to ChatEmoteSpace.cs

```suggestion
public class ChatEmote : RectangularMenuObject, ButtonScroller.IPartOfButtonScroller
{
    public Vector2 SourceSize { get; set; }
    public float Alpha { get; set; }
    public Vector2 Pos { get => pos; set => pos = value; }
    public Vector2 Size { get => size + Margin; set {} }
    public Vector2 Margin { get; set; } = new(40f, 10f);

    public FSprite
        backgroundSprite,
        emoteSprite;

    public ChatEmote(
        MeadowProgression.Character character,
        MeadowProgression.Emote emote,
        Menu.Menu menu,
        MenuObject owner,
        Vector2 pos) : base(menu, owner, pos, Vector2.zero)
    {
        if (!Futile.atlasManager.DoesContainAtlas("emotes_common"))
        {
            HeavyTexturesCache.futileAtlasListings.Add(
                Futile.atlasManager.LoadAtlas("illustrations/emotes/emotes_common").name
            );
        }

        MeadowProgression.CharacterData characterData = MeadowProgression.characterData[character];

        if (!Futile.atlasManager.DoesContainAtlas(characterData.emoteAtlas))
        {
            HeavyTexturesCache.futileAtlasListings.Add(
                Futile.atlasManager.LoadAtlas("illustrations/emotes/" + characterData.emoteAtlas).name
            );
        }

        string emoteSpriteName = (emote.value.StartsWith("emote")
            ? characterData.emotePrefix + emote.value
            : emote.value)
            .ToLowerInvariant();

        string backgroundSpriteName = emote.value.StartsWith("emote")
            ? "emote_background"
            : "symbols_background";

        backgroundSprite = new FSprite(backgroundSpriteName);
        emoteSprite = new FSprite(emoteSpriteName);
        SourceSize = backgroundSprite.element.sourceSize;
        size = SourceSize * 0.35f;

        backgroundSprite.SetAnchor(0f, 1f);
        backgroundSprite.shader = FShader.defaultShader;
        Container.AddChild(backgroundSprite);
        emoteSprite.SetAnchor(0f, 1f);
        emoteSprite.shader = FShader.defaultShader;
        Container.AddChild(emoteSprite);
    }

    public override void GrafUpdate(float timeStacker)
    {
        base.GrafUpdate(timeStacker);
        backgroundSprite.alpha = Alpha * 0.6f;
        emoteSprite.alpha = Alpha;

        backgroundSprite.SetPosition(Margin + DrawPos(timeStacker));
        backgroundSprite.scaleX = size.x / SourceSize.x;
        backgroundSprite.scaleY = size.y / SourceSize.y;
        emoteSprite.SetPosition(Margin + DrawPos(timeStacker));
        emoteSprite.scaleX = size.x / SourceSize.x;
        emoteSprite.scaleY = size.y / SourceSize.y;
    }

    public override void RemoveSprites()
    {
        base.RemoveSprites();

        backgroundSprite.RemoveFromContainer();
        emoteSprite.RemoveFromContainer();
    }
}

public class ChatEmoteSpace : RectangularMenuObject, ButtonScroller.IPartOfButtonScroller
{
    public ChatEmote ChatEmote { get; set; }
    public float Alpha { get; set; }
    public Vector2 Pos { get; set; }
    public Vector2 Size { get => new(ChatEmote.Size.x, size.y); set => size = value; }

    public ChatEmoteSpace(Menu.Menu menu, MenuObject owner, Vector2 pos, Vector2 size, ChatEmote chatEmote)
        : base(menu, owner, pos, size)
    {
        ChatEmote = chatEmote;
    }
}

```

Menu/Components/ChatScroller.cs
```diff
@@ -0,0 +1,308 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using HarmonyLib;
+using Menu;
+using Menu.Remix.MixedUI;
+using RWCustom;
+using UnityEngine;
+
+namespace RainMeadow.UI.Components
```
One-Letter-Shor, line 10: This should also be a file scoped namespace. I'm not going to leave a full suggestion here because I think that chat itself needs to at least partially be refactored.

One-Letter-Shor (changes requested):

henpemaz:
Looks cute, looking forward to this one

Menu/Components/ChatEmote.cs
henpemaz: Please ignore the following points:

> - Block scoped namespace (adds unnecessary indents)
> - Array of sprites instead of two fields (you have to use magic numbers to get the sprites from the array)
> - Source code frequently breaks 120 characters
> - The naming variables for the sprite name was just called 'XXXSprite' instead of 'XXXSpriteName'

The points above don't need addressing

Menu/Components/ChatScroller.cs
henpemaz: We're not following any patterns with file-scoped or otherwise in this project, just follow your preferences.

Menu/Components/ChatEmote.cs
henpemaz: This was copied from EmoteDisplayer.cs where the manual cache management makes sense. The DoesContainAtlas is a guard so that it's only added to the cache list once.

I imagine there's menu-class helpers for loading and caching atlases? Either way it works

PR #1612, opened 2026-08-23
docs: Tweak CHANGELOG.md

One-Letter-Shor:


CHANGELOG.md
```diff
@@ -1,20 +1,25 @@
...
 - Fixed quitting to menu not ending the match
-### Modders 
-- Added `ExternalGameMode.GetPlayersStillActive` method
+
+### Modders
+- Added `ExternalArenaGameMode.GetActivePlayerACs` method
+
 ## Chat
-- Added a remix option to clear certain chat messages on every session.
+- Added a remix option to clear specific system messages each time a new arena round starts
```
One-Letter-Shor: @EdEnStonne This is correct right? It's specifically only for arena?

One-Letter-Shor:
This was supposed to be a draft, waiting on confirmation from ed that one change is correct

CHANGELOG.md
EdEnStonne: Works with story too, it isn't arena specific, but it's correct.

EdEnStonne: "each time a new arena round or a new cycle starts" would be more accurate

invalidunits (approved):

PR #1617, opened 2026-08-24
fix: unsafe indexing

6fears7:
- fix lookups when arenaSitting player list is cleared
- fix unsafe indexing from FindOnlinePlayerByFakePlayerNumber
- clean up GetAmoebaColor indexing and readability

### DIFF
--- a/Arena/ArenaHelpers.cs
+++ b/Arena/ArenaHelpers.cs
@@ -168,7 +168,7 @@ namespace RainMeadow
 
             return playerNumber == -1
                 ? null
-                : arenaSitting.players[playerNumber];
+                : arenaSitting.players.Find(arenaPlayer => arenaPlayer != null && arenaPlayer.playerNumber == playerNumber);
         }
 
         public static OnlinePlayer? FindOnlinePlayerByFakePlayerNumber(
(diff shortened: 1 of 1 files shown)


Arena/ArenaHelpers.cs
```diff
@@ -168,7 +168,7 @@ public static OnlinePlayer FindOnlinePlayerByStringUsername(string username)
 
             return playerNumber == -1
```
One-Letter-Shor: I think this should be replaced with a comment and a one line of code (comment tweaked as needed to be correct and specific)

Applies to 169-171 (github is shitty and won't let you select multiple lines on mobile it seems)

```cs
// When the round transitions the online list keeps all player numbers but the local list of arena players is cleared. These leaves some niche cases where the player number is out of bounds of the local list.
Return arenaSitting.players.Find(arenaPlayer => arenaPlayer.playerNumber == playerNumber);
```

6fears7: kept null check because of the use of `FindArenaPlayerByOnlinePlayer` in per-frame checks that might still run during NextLevel when the list clears

invalidunits (approved):

PR #1621, opened 2026-08-25
build: Use specific LangVersion

### LINKED ISSUE #1535: Consider locking the project's LangVersion to a specific version
everyone can have different versions of the .net sdk installed on their systems so this would ensure everyone has the same feature set and doesnt accidentally push code that breaks for someone else because they have a different version of .net

microsoft's own [documentation](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version) recommends not using `latest` as well
Image

not making a PR for this because im unsure which c# version to target but imo the best candidates would be either c# 14.0 (latest in .net 10 which is the latest LTS release) or 12.0 (latest in .net 8 which is the previous LTS release)

One-Letter-Shor:
Resolves #1535

### DIFF
--- a/Rain Meadow.csproj
+++ b/Rain Meadow.csproj
@@ -2,7 +2,7 @@
 	<PropertyGroup>
 		<TargetFramework>net48</TargetFramework>
 		<Version>2.0.8</Version>
-		<LangVersion>latest</LangVersion>
+		<LangVersion>14</LangVersion>
 		<Nullable>enable</Nullable>
 		<AllowUnsafeBlocks>true</AllowUnsafeBlocks>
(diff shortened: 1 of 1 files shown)


cgytrus:
should note that most developers working on rain meadow are already using .net 10 and, consequently, c# 14, as it is what the `latest` version is on .net 10, and those that arent wont even be able to build the project right now anyway because some of the code in the project uses c# 14 features, but targetting c# 14 explicitly would show a much clearer error instead of random places in the code just appearing broken

6fears7:
I agree using latest is in some ways asking for trouble

PR #1623, opened 2026-08-27
mod auto-enabler revamp

clkursch:
When players join a lobby, the mod auto-enabler should now also auto-enable dependencies of mods it's slated to enable. same with the auto-disabler.

### DIFF
--- a/ModManager/RainMeadowModManager.cs
+++ b/ModManager/RainMeadowModManager.cs
@@ -125,6 +125,28 @@ namespace RainMeadow
                 var disable = GetRequiredMods().Union(bannedMods).Except(requiredMods).Intersect(active).ToList();
                 var enable = requiredMods.Except(active).ToList();
 
+                //add mod dependencies to the list 
+                int countEnabled = 0;
+                while (countEnabled < enable.Count)
+                {
+                    countEnabled = enable.Count; //loop until no more additions are made
+                    for (int i = 0; i < enable.Count; i++)
+                    {
+                        int index = ModManager.InstalledMods.FindIndex(mod => mod.id == enable[i]);
+                        if (index < 0) continue;
+                        foreach (var req in ModManager.InstalledMods[index].requirements)
+                        {
+                            int reqindex = ModManager.InstalledMods.FindIndex(mod => mod.id == req);
+                            if (reqindex < 0) continue;
+                            if (!(enable.Contains(req) || ModManager.ActiveMods.Contains(ModManager.InstalledMods[reqindex]))) //ignore mods that are already in enable; no change necessary
+                                if (ModManager.InstalledMods.Contains(ModManager.InstalledMods[reqindex])) //if the dependency is installed
+                                    enable.Add(ModManager.InstalledMods[reqindex].id); //enable it            
+                        }
+                    }
+                    //THEN RUN THE CHECK AGAIN WITH THE DEPENDENCIES IN THE ENABLED LIST
+                }
+
+                
                 //clear phony entries to the mod list
                 enable.RemoveAll(mod => mod == null || mod == "" /* || mod == "henpemaz_rainmeadow" */);
                 disable.RemoveAll(mod => mod == null || mod == "" /* || mod == "henpemaz_rainmeadow" */);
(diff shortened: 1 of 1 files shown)


ModManager/RainMeadowModManager.cs
```diff
@@ -125,6 +125,28 @@ internal static void CheckMods(string[] requiredMods, string[] bannedMods, Actio
...
+                            int reqindex = ModManager.InstalledMods.FindIndex(mod => mod.id == req);
+                            if (reqindex < 0) continue;
+                            if (!(enable.Contains(req) || ModManager.ActiveMods.Contains(ModManager.InstalledMods[reqindex]))) //ignore mods that are already in enable; no change necessary
+                                if (ModManager.InstalledMods.Contains(ModManager.InstalledMods[reqindex])) //if the dependency is installed
+                                    enable.Add(ModManager.InstalledMods[reqindex].id); //enable it            
+                        }
+                    }
+                    //THEN RUN THE CHECK AGAIN WITH THE DEPENDENCIES IN THE ENABLED LIST
+                }
+
```
One-Letter-Shor: I think that this code can be simplified.
- Removes unnecessary while loop (FYI, if you did need a while loop, a do while loop would be better)
- Get the mod object directly instead of finding the index first. (This and the next point has the side effect of not having three index variables)
- Removes unnecessary FindIndex of requiredModId
- Inverts nested ifs to be early continues for readability

```suggestion
                // Add mod dependencies to the list.
                // Note: Because detected dependencies immediately get added to the end of enable, the for loop also checks those dependencies' dependencies (basically recursion).
                for (int i = 0; i < enable.Count; i++)
                {
                    string modIdToEnable = enable[i];

                    ModManager.Mod? modToEnable = ModManager.InstalledMods
                        .FirstOrDefault(mod => mod.id == modIdToEnable);
                    if (modToEnable is null) continue;

                    foreach (string requiredModId in modToEnable.requirements)
                    {
                        if (enable.Contains(requiredModId)) continue;
                        if (ModManager.ActiveMods.Any(mod => mod.id == requiredModId)) continue;
                        if (!ModManager.InstalledMods.Any(mod => mod.id == requiredModId)) continue;

                        enable.Add(requiredModId);
                    }
                }
```

ModManager/RainMeadowModManager.cs
```diff
@@ -190,10 +212,15 @@ internal static void CheckMods(string[] requiredMods, string[] bannedMods, Actio
...
-                            disable.Add(mod.id); //disable the mod
+                int countDisabled = 0;
+                while (countDisabled < disable.Count)
+                {
+                    foreach (var mod in ModManager.ActiveMods)
+                        if (!disable.Contains(mod.id)) //ignore mods that are already in disable; no change necessary
+                            if (disable.Exists(id => mod.requirements.Contains(id))) //if one of its dependencies is being disabled
+                                disable.Add(mod.id); //disable the mod
+                    countDisabled = disable.Count;
+                }
```
One-Letter-Shor: I think that this code can be simplified.
- ~~Replaces while loop with a for loop (FYI, if you did need a while loop, a do while loop would be better)~~
- Fixes an issue where the loop only runs once (It seems like the while condition immediately evaluates to false after the first execution)
- Uses Linq's .Any() instead of List's .Exists()
- Inverts nested ifs to be early continues for readability

```suggestion
                bool hasAddedAnyDependents;
                do
                {
                    hasAddedDependent = false;
                    foreach (ModManager.Mod mod in ModManager.ActiveMods)
                    {
                        if (disable.Contains(mod.id)) continue;
                        if (!disable.Any(id => mod.requirements.Contains(id))) continue;

                        disable.Add(mod.id);
                        hasAddedAnyDependents = true;
                    }
                } while (hasAddedAnyDependents);
```

One-Letter-Shor (changes requested):

ModManager/RainMeadowModManager.cs
One-Letter-Shor: Actually this suggestion might be bad, a do while is probably better (I'll check tomorrow)

One-Letter-Shor: Ok, I updated the suggestion to use a do while instead of a for

clkursch: do you know if it works?

One-Letter-Shor: It should work the same at a high level other than the bug fix but it hasn't been fully tested

clkursch: would it be better to merge the tested version and then make a separate pull request or

clkursch: eh I'll just accept it

clkursch:
is this all I need to do?

One-Letter-Shor:
I think you missed the other comment

clkursch:
how about now?

PR #1625, opened 2026-08-27
Fix name tag positioning

cgytrus:
- fixed name tag position being behind by an extra tick
  - theres still another tick of delay but it comes from rain world itself so itd be out of scope for rain meadow to fix
  - more specifically, HUD.Update gets called before almost everything else (including OverWorld.Update) in RainWorldGame.Update, which means that the hud updates with the state of the game from the previous tick before the game is actually updated. this affects *all* huds in rain world, including name tags / player arrows in vanilla, so a proper fix would fix the issue for everything, which would happen to include rain meadow name tags

before | after
--- | ---
<video src="https://github.com/user-attachments/assets/aeaa7d60-a94f-41ae-9025-38909eee07bc"> | <video src="https://github.com/user-attachments/assets/122921e1-f79f-4b50-a3e0-ee713abb22f1">
---

- fixed the arrow being incorrectly offset relative to the shortcut icon when showing a player in the neighboring room

before | after
--- | ---
image | image
---

- fixed the name tag lerping its position when it should change instantly (like when the camera moves or changes its room)
  - note that the tick of delay from the rw bug described above is still present, but the switch is still a lot less jarring

before | after
--- | ---
<video src="https://github.com/user-attachments/assets/843fa682-7073-48da-8377-a6a655d87827"> | <video src="https://github.com/user-attachments/assets/3b51da24-9f97-4131-8ec6-8a380413efa1">
---

- fixed an edge case in clamping logic thatd sometimes cause the name tag to be closer to the edge of the screen than intended
  - when the target player is in a neighbor room and the shortcut is close enough to the edge of the screen that the name tag ends up being less than 40 pixels (2 tiles) away from the edge of the screen before applying the 10 pixel offset

before | after
--- | ---
<video src="https://github.com/user-attachments/assets/a9020917-3c4c-4215-b114-4ed6b1036e36"> | <video src="https://github.com/user-attachments/assets/bb12a20c-3a23-437c-af8e-fdf1da917405">
---

- fixed positioning logic for far away players
  - if a far away player happened to be realized (i think this only happens with jolly coop?) the old logic would technically incorrectly target the main body chunk, it would also only update the target position when the player's abstract creature's position changes regardless of whether the realized creature's position changes
  - also name tags would sometimes just freeze until you get close enough to their player
---

- improved positioning for players really close to edges of the screen (<30px)

before | after
--- | ---
<video src="https://github.com/user-attachments/assets/61885c4f-3969-45bc-b5f4-3b2f35208998"> | <video src="https://github.com/user-attachments/assets/1a71a26c-d234-4382-bf4f-f3f70a04a774">
---

most of these fixes come from just rewriting the positioning logic almost entirely so i had to put it in a single commit which made git generate kind of a bad diff, idk if i can make it look better for easier reviewing, sorry :(

ive tested this quite extensively and couldnt notice any issues (other than ones already present in rain meadow that are out of scope for this PR)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+# Release 1.16.0
+
+## General
+- Improved name tag positioning
+  - Fixed name tag position being behind by an extra tick
+  - Fixed name tag being offset incorrectly when pointing at a player in a neighboring room
+  - Fixed incorrect lerping logic
+  - Fixed an edge case in positioning logic that would sometimes cause it to go closer to the edge of the screen than intended
+  - Fixed positioning logic for far away players
+  - Improved positioning for players really close to edges of the screen
+
 # Release 1.15.2
 
 ## Arena 
--- a/Extensions.cs
+++ b/Extensions.cs
@@ -315,6 +315,11 @@ namespace RainMeadow
                 targetPoint.y >= rect.yMin &&
                 targetPoint.y <= rect.yMax) return targetPoint;
 
+            return rect.GetClosestPointOnEdgeAlongLineFromCenter(targetPoint);
+        }
+
+        public static Vector2 GetClosestPointOnEdgeAlongLineFromCenter(this Rect rect, Vector2 targetPoint)
+        {
             float halfWidth = rect.width * 0.5f;
             float halfHeight = rect.height * 0.5f;
             targetPoint -= rect.center; // from center
--- a/OnlineUIComponents/OnlinePlayerDeathBump.cs
+++ b/OnlineUIComponents/OnlinePlayerDeathBump.cs
@@ -41,9 +41,7 @@ namespace RainMeadow
 
         public void SetPosToPlayer()
         {
-            this.pos = owner.drawpos + new Vector2(0, 30f);
-            this.pos.x = Mathf.Clamp(this.pos.x, 30f, this.owner.camera.sSize.x - 30f);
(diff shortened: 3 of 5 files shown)


cgytrus:
turns out i was dumb and part of the old logic for far away players was correct so i changed that back, it now looks identical to how it was before most of the time so i removed the comparison videos but it is still smoother for jolly coop players now

6fears7:
@cgytrus Thanks for the videos. Can you confirm the following were tested?

- [ ] Player state (alive/dead) in arena & story
- [ ] icon hidden when invis as watcher in arena
- [ ] Team battle icon in team battle
- [ ] State updates (dead, blinded, MVP) in arena
- [ ] Global Mute (check Remix > Gameplay)

cgytrus:
marked as draft while testing and fixing things

cgytrus:
testing progress, removed the one irrelevant entry and reworded others
also idk what blinded is
- [x] story player state (alive/dead)
- [x] arena player state (alive/dead, MVP)
- [x] icon hidden when invis as watcher in arena
- [x] Team battle icon in team battle

cgytrus:
tested more in arena, story and meadow, didnt find any more issues

6fears7 (dismissed):

6fears7:
@cgytrus if you'd like a playtest ping for it, just let me know.

6fears7:
Included in playtest today. No issues

6fears7 (approved):
Tested, worked

PR #1626, opened 2026-08-29
remove nullables from some of TryGet patterns from Extensions

iulh:
it's not a first time i question why `out` arguments of these methods are considered nullable if it doesn't matter much, since they're thrown away if they're `null` on `false` returned.
ideally they're supposed to have a prefix `Try`, and not be an overload. but it's besides the point.

so i've asked choc, who wrote these methods. [quote](https://discord.com/channels/1094716194180841602/1155268800669831270/1542605607763447888):
> can i ask why `GetOnlineObject` has a signature of `bool AbstractPhysicalObject.GetOnlineObject(out OnlinePhysicalObject? opo)`, and not `bool AbstractPhysicalObject.TryGetOnlineObject(out OnlinePhysicalObject opo)`?
> this is like.. a flaw in design. "TryGet" methods are always guaranteed to return true on existing out variable, but it's marked as nullable

then i explained my point and he said i'm free to PR, which i'm doing right now.
PLUS i've fixed some unnecessary checks (i hope i didn't break anything, didn't test much), to show it cuts some warnings for free.

there's also some `TryGet` methods outside of Extensions i found, but the PR doesn't touch them.
and lowering the scope is better to me, since i wont be home for a day if this even matters, and generally smaller scope is better to show the point to show the point.

@One-Letter-Shor also commented he's "working on something rn".
i'm mentioning it since i'm curious

### DIFF
--- a/Arena/AmoebaSummonBehavior.cs
+++ b/Arena/AmoebaSummonBehavior.cs
@@ -38,7 +38,7 @@ public class AmoebaSummonBehavior(VoidSpawn owner) : VoidSpawn.Behavior(owner)
             {
                 if (
                     arenaPlayer.playerNumber
-                    == ArenaHelpers.FindOnlinePlayerNumber(arena, opo!.owner)
+                    == ArenaHelpers.FindOnlinePlayerNumber(arena, opo.owner)
                 )
                 {
                     additionalPoints = arenaPlayer.allKills.Count;
@@ -56,7 +56,7 @@ public class AmoebaSummonBehavior(VoidSpawn owner) : VoidSpawn.Behavior(owner)
             if (!RainMeadow.isArenaMode(out var arena)) return base.SwimTowards;
             VoidSpawn voidSpawn = this.owner;
             // RainMeadow.Debug($"{voidSpawn} choosing a behavior...");
-            
+
 
             // 1. If pointing, go toward the point.
             if (arena.amoebaControl && Input.GetKey(RainMeadow.rainMeadowOptions.PointingKey.Value))
@@ -82,7 +82,7 @@ public class AmoebaSummonBehavior(VoidSpawn owner) : VoidSpawn.Behavior(owner)
                     return pointingVector;
                 }
             }
-            
+
 
             // 2. If spear hit is on, and a target player was found, go toward it.
             Player? ownerPlayer = null;
@@ -106,11 +106,11 @@ public class AmoebaSummonBehavior(VoidSpawn owner) : VoidSpawn.Behavior(owner)
                 }
                 if (!voidSpawn.room.game.GetArenaGameSession.arenaSitting.gameTypeSetup.spearsHitPlayers)
                     continue;
-                
+
                 if (TeamBattleMode.IsTeamBattleMode(out var tb))
                 {
                     ArenaTeamClientSettings? playerTeam =
-                        ArenaHelpers.GetDataSettings<ArenaTeamClientSettings>(oe!.owner);
+                        ArenaHelpers.GetDataSettings<ArenaTeamClientSettings>(oe.owner);
                     if (playerTeam != null && playerTeam.team == arena.arenaTeamClientSettings.team)
                         continue;
                 }
@@ -131,18 +131,18 @@ public class AmoebaSummonBehavior(VoidSpawn owner) : VoidSpawn.Behavior(owner)
                 )
                 {
                     foundPlayer = realizedPlayer;
-                    foundPlayerPriority  = GetPriority(arena, voidSpawn, foundPlayer);
+                    foundPlayerPriority = GetPriority(arena, voidSpawn, foundPlayer);
                     minDistance = distance;
                 }
             }
-            
-            if (foundPlayer != null) 
+
+            if (foundPlayer != null)
             {
                 // RainMeadow.Debug($"(2) Attacking Player !");
                 this.wasCircling = false;
                 return foundPlayer.mainBodyChunk.pos;
             }
-            
+
 
             // 3. If a hostile creature is found, go toward it.
             Creature? foundCreature = null;
@@ -167,7 +167,7 @@ public class AmoebaSummonBehavior(VoidSpawn owner) : VoidSpawn.Behavior(owner)
                 );
                 int aggression = voidSpawn.abstractPhysicalObject.rippleLayer != creature.abstractCreature.rippleLayer
                     ? 0 // can't really aggro if you are not in the same realm
-                    : (ownerPlayer is not null 
+                    : (ownerPlayer is not null
                         ? (int)(abstractCreature.abstractAI?.RealAI?.CurrentPlayerAggression(ownerPlayer.abstractCreature) * 10 ?? 0)
                         : 5); // aggression is from 0 to 1, make it an int from 0 to 10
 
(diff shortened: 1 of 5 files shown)


Extensions.cs
```diff
@@ -41,29 +41,30 @@ public static Color ColorFromHex(int value)
...
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static bool IsLocal(this AbstractPhysicalObject apo, out OnlinePhysicalObject? opo)
+        public static bool IsLocal(this AbstractPhysicalObject apo, out OnlinePhysicalObject opo)
         {
-            opo = null;
-            return OnlineManager.lobby is null || (OnlinePhysicalObject.map.TryGetValue(apo, out opo) && opo.isMine);
+            opo = default!;
+            if (OnlineManager.lobby is null) return false;
+            return OnlinePhysicalObject.map.TryGetValue(apo, out opo) && opo.isMine;
         }
```
One-Letter-Shor, line 60: ~~This breaks API by returning false instead of true when the lobby is null. Some mods rely on this so it should be included in the changelog.
Really this method should probably be removed (including the other `IsLocal` that outs a opo) because the caller should just use `TryGetOnlineObject` instead.~~

~~But if this really needs to be kept, I think it would be better to throw instead of return false when not in a lobby. This will prevent callers from making incorrect code/bad assumptions that go unnoticed. (see #1627)~~

```
/// <exception cref="InvalidOperationException">Thrown if not in a <see cref="Lobby"/>.</exception>
[MethodImpl(MethodImplOptions.AggressiveInlining)]
public static bool IsLocal(this AbstractPhysicalObject apo, out OnlinePhysicalObject opo)
{
    if (OnlineManager.lobby is null)
        throw new InvalidOperationException("Not in a lobby.");

    opo = null!;
    return OnlinePhysicalObject.map.TryGetValue(apo, out opo) && opo.isMine;
}
```

New decision: Keep all IsLocal overloads the same so a different PR can mark them as obsolete and add `IsMine` that throws.

Extensions.cs
```diff
@@ -41,29 +41,30 @@ public static Color ColorFromHex(int value)
...
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static bool GetOnlineObject(this AbstractPhysicalObject apo, out OnlinePhysicalObject? opo) => OnlinePhysicalObject.map.TryGetValue(apo, out opo);
+        public static bool GetOnlineObject(this AbstractPhysicalObject apo, out OnlinePhysicalObject opo) => OnlinePhysicalObject.map.TryGetValue(apo, out opo);
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static OnlineCreature? GetOnlineCreature(this AbstractCreature ac) => GetOnlineObject(ac) as OnlineCreature;
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static bool GetOnlineCreature(this AbstractCreature apo, out OnlineCreature? oc) => (oc = GetOnlineCreature(apo)) is not null;
+        public static bool GetOnlineCreature(this AbstractCreature apo, out OnlineCreature oc) => (oc = GetOnlineCreature(apo) ?? default!) is not null;
```
One-Letter-Shor: This should not be written as a single expression (needlessly complicated/confusing) and this should be prefixed with Try

Extensions.cs
```diff
@@ -41,29 +41,30 @@ public static Color ColorFromHex(int value)
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static bool GetOnlineObject(this AbstractPhysicalObject apo, out OnlinePhysicalObject? opo) => OnlinePhysicalObject.map.TryGetValue(apo, out opo);
+        public static bool GetOnlineObject(this AbstractPhysicalObject apo, out OnlinePhysicalObject opo) => OnlinePhysicalObject.map.TryGetValue(apo, out opo);
```
One-Letter-Shor, line 44: This should be prefixed with Try (I know that you said you didn't want to overload this PR, but I think this is an important improvement)

One-Letter-Shor:
There are a few things I'd like to point out:
- You should not be replacing specified types with `var` in the code
- You should use `null` instead of `default` when you know the type is a reference type
- `GetOnlineObject` and `GetOnlineCreature` should throw when they don't find the object they are looking for considering TryGet versions exist

Extensions.cs
iulh: this is so far the only major overlook, tysm for the check

Extensions.cs
iulh: so i split the methods?

Extensions.cs
6fears7: IsLocal has no need to throw if false because it's still correct: the entity is locally owned. 

Returning a bool instead of throwing also improves soft compatibility, because if my mod is designed to work with or without meadow lobbies, I can use the Boolean to check that without adding an early return. 

Under a throw, I have to now rewrite the mod to perform an early return on orig(), which creates work for others without any real benefit to anyone.

6fears7 (commented):
Placeholder comment because GitHub is wack

iulh:
i'm not sure about prefixing all the TryGet patterns, tho it looks out of standard. like it doesn't fit this PR. i purposefully scoped not to affect much.

throwing an exception is also not a good fit, which is also mentioned by UO to have potential issues with other mods.

and overlooked typo is fixed.
it's also funny how i accidentally highlighted another logic error which wasn't made by me

Extensions.cs
One-Letter-Shor: Returning false from IsLocal allows bugs to show up. If someone thinks they are in a lobby and they aren't, IsLocal will return an unexpected value. This will result in code that throws later or even worse, fails silently.

Also, the function was already sensibly changed to return false when the lobby is null. This allows the out Opo to always be not null when it returns true. This already will break mods that rely on the old (poor) design. If this threw instead it would be so much easier to understand what's happening

Extensions.cs
One-Letter-Shor, line 1: @iulh I think you should revert the changes you made to all IsLocal overloads. A future PR will mark then as Obsolete (and another helper method will be added). This give dependants time to update their code before it breaks. (You will have to make the out variants of IsLocal out a nullable opo though)

One-Letter-Shor (changes requested):

Extensions.cs
iulh: > and another helper method will be added

yo this is actually what i was wondering about.
made something solo but didn't push in the end

One-Letter-Shor:
These changes were not done yet, and it would probably be best as a separate PR @iulh

iulh:
> These changes were not done yet, and it would probably be best as a separate PR @iulh

as you say so

Arena/AmoebaSummonBehavior.cs
henpemaz, line 1: 12 lines changed to remove a single "!"

henpemaz:
This project disregards nullability through most of it's codebase, I don't see the point. I'd probably accept a concise PR if it were only changing the 8 lines that it should be changing rather than 50 lines of whitespace changes.

Arena/AmoebaSummonBehavior.cs
iulh: > 12 lines changed to remove a single "!"

well it is like that for now, Shor was about to PR some changes for clarity but i think he forgot

PR #1628, opened 2026-08-30
refactor(ui): dialogs

T1mbits:
A major refactor of nearly all dialogs in the codebase. This aims to make standard notify dialog and confirmation dialog behaviour consistent for every dialog that shows up in rain meadow and reduce redundancies.

I am fairly confident that every single dialog is working, but I haven't done extensive testing pertaining to timeouts (`DialogAsyncWaitCancellable`), any of the ModApplier dialogs, or the warning popup in the IntroRoll when someone opens Rain Meadow with Inv set as their active campaign.

This does not cover `ArenaPostGameStatsDialog.cs`, `ColorMultipleSlugcatsDialog.cs`, `ColorSlugcatDialog.cs`, or `SlugcatSelector.cs` as they are vastly different from other dialogs. It also does not cover events.


T1mbits:
guh I had an evil stashed change with an insanely large addition

6fears7 (approved):

invalidunits (approved):

PR #1629, opened 2026-08-30
Refactor for the interaction between ButtonScroller and ButtonScrollerObjects

Dustylemny:
This may be subjected to change as I did these in a haste of time and may make a different pr if necessary. this also is just to see things out.

This sort of aims to fix buttonscroller reiterating over it's own subobjects. while also fixes some shortcuts like a following subobject  which is a buttonscroller.Ipartofbuttonscroller of a subobject of a buttonscroller.Ipartofbuttonscroller subobject will be unable to get alpha properly. And also the fix the overreliance to copy paste Alpha, Pos, Size properties. Currently can backtrack for buttonscroller.IPartScroller but if needed can try to use IPartScroller instead on relying on a new interface. Also ButtonScroller will just follow adding Menuobjects since it is what it does.

Additional Todo list:

- [x]  Add a scrollable (supports custom spacing of menuobjects)
- [ ]  Maybe some api for it if necessary
- [ ]  Also look into ArenaLevel SearchBar not taking inputs apparently?


Dustylemny:
@T1mbits i hearby in dire need of your service and music to my eyes

Dustylemny:
Closed due to being merged to #1640

PR #1633, opened 2026-08-31
feat: bundle git info on every debug build

iulh:
this addition probably worth nothing, but in the end we get a clever way to "trace back" when "a change" was made

### DIFF
--- a/Rain Meadow.csproj
+++ b/Rain Meadow.csproj
@@ -45,7 +45,7 @@
 	</ItemGroup>
 
     
-    <Target Name="GitGenerate" BeforeTargets="BeforeCompile" Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <Target Name="GitGenerate" BeforeTargets="BeforeCompile" Condition="'$(Configuration)' == 'Debug'">
         <Exec Command='git branch --show-current' ConsoleToMSBuild="true">
             <Output TaskParameter="ConsoleOutput" PropertyName="GitBranch" />
         </Exec>
--- a/Rain Meadow.csproj
+++ b/Rain Meadow.csproj
@@ -45,7 +45,7 @@
 	</ItemGroup>
 
     
-    <Target Name="GitGenerate" BeforeTargets="BeforeCompile">
+    <Target Name="GitGenerate" BeforeTargets="BeforeCompile" Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
         <Exec Command='git branch --show-current' ConsoleToMSBuild="true">
             <Output TaskParameter="ConsoleOutput" PropertyName="GitBranch" />
         </Exec>
--- a/Rain Meadow.csproj
+++ b/Rain Meadow.csproj
@@ -62,10 +62,10 @@
             <Output TaskParameter="ConsoleOutput" PropertyName="GitHead" />
         </Exec>
 
-        <WriteLinesToFile File="$(IntermediateOutputPath)Git.cs" Overwrite="true" Lines="namespace RainMeadow { public static class Git { public const string GIT_BRANCH_NAME = &quot;$(GitBranch)&quot;%3B public const string GIT_REMOTE_URL = &quot;$(GitRemoteUrl)&quot;%3B public const string GIT_HEAD_COMMIT = &quot;$(GitHead)&quot;%3B } }" />
+        <WriteLinesToFile File="$(IntermediateOutputPath)GitInfo.g.cs" Overwrite="true" Lines="namespace RainMeadow { public static class GitInfo { public const string BRANCH_NAME = &quot;$(GitBranch)&quot;%3B public const string REMOTE_URL = &quot;$(GitRemoteUrl)&quot;%3B public const string HEAD_COMMIT = &quot;$(GitHead)&quot;%3B } }" />
 
         <ItemGroup>
-            <Compile Include="$(IntermediateOutputPath)Git.cs" />
+            <Compile Include="$(IntermediateOutputPath)GitInfo.g.cs" />
         </ItemGroup>
     </Target>
 
--- a/Rain Meadow.csproj
+++ b/Rain Meadow.csproj
@@ -47,22 +47,22 @@
     
     <Target Name="GitGenerate" BeforeTargets="BeforeCompile" Condition="'$(Configuration)' == 'Debug'">
         <Exec Command='git branch --show-current' ConsoleToMSBuild="true">
-            <Output TaskParameter="ConsoleOutput" PropertyName="GitBranch" />
+            <Output TaskParameter="ConsoleOutput" PropertyName="ExistingGitBranch" />
         </Exec>
-        
-        <Exec Command='git config --get branch.$(GitBranch).remote' ConsoleToMSBuild="true">
+
+        <Exec Command='git config --default "" --get branch.$(ExistingGitBranch).remote' Condition="'$(ExistingGitBranch)' == ''" ConsoleToMSBuild="true">
             <Output TaskParameter="ConsoleOutput" PropertyName="GitRemoteName" />
         </Exec>
 
-        <Exec Command='git remote get-url $(GitRemoteName)' ConsoleToMSBuild="true">
+        <Exec Command='git remote get-url $(GitRemoteName)' Condition="'$(GitRemoteName)' != ''" ConsoleToMSBuild="true">
             <Output TaskParameter="ConsoleOutput" PropertyName="GitRemoteUrl" />
         </Exec>
-
+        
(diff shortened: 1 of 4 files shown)


iulh:
look*

RainMeadow.cs
```diff
@@ -178,6 +178,18 @@ private void RainWorld_OnModsInit(On.RainWorld.orig_OnModsInit orig, RainWorld s
 
                 InitializeExtEnums();
 
+#if DEBUG
+                var git = Type.GetType("RainMeadow.Git");
+                if (git == null) RainMeadow.Warn("Git Info wasn't bundled");
```
iulh: is it a good place to add logging?

Rain Meadow.csproj
```diff
@@ -44,6 +44,31 @@
 		</None>
 	</ItemGroup>
 
+    
+    <Target Name="GitGenerate" BeforeTargets="BeforeCompile" Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
```
cgytrus: dont need to check platform here, only configuration

```suggestion
    <Target Name="GitGenerate" BeforeTargets="BeforeCompile" Condition="'$(Configuration)' == 'Debug'">
```

Rain Meadow.csproj
```diff
@@ -44,6 +44,31 @@
...
+        
+        <Exec Command='git config --get branch.$(GitBranch).remote' ConsoleToMSBuild="true">
+            <Output TaskParameter="ConsoleOutput" PropertyName="GitRemoteName" />
+        </Exec>
+
+        <Exec Command='git remote get-url $(GitRemoteName)' ConsoleToMSBuild="true">
+            <Output TaskParameter="ConsoleOutput" PropertyName="GitRemoteUrl" />
+        </Exec>
+
+        <Exec Command="git rev-parse HEAD" ConsoleToMSBuild="true">
```
cgytrus, line 61: maybe consider changing this to `git describe` or at least adding `--short`, see if it looks better
```suggestion
        <Exec Command="git describe --always --dirty" ConsoleToMSBuild="true">
```
```suggestion
        <Exec Command="git rev-parse HEAD --short" ConsoleToMSBuild="true">
```

Rain Meadow.csproj
```diff
@@ -44,6 +44,31 @@
...
+            <Output TaskParameter="ConsoleOutput" PropertyName="GitHead" />
+        </Exec>
+
+        <WriteLinesToFile File="$(IntermediateOutputPath)Git.cs" Overwrite="true" Lines="namespace RainMeadow { public static class Git { public const string GIT_BRANCH_NAME = &quot;$(GitBranch)&quot;%3B public const string GIT_REMOTE_URL = &quot;$(GitRemoteUrl)&quot;%3B public const string GIT_HEAD_COMMIT = &quot;$(GitHead)&quot;%3B } }" />
```
cgytrus: the file name should end with .g.cs or .gen.cs since its auto-generated, and ideally would also have a comment at the top saying that its auto-generated
```suggestion
        <WriteLinesToFile File="$(IntermediateOutputPath)Git.g.cs" Overwrite="true" Lines="/* Auto-generated */ namespace RainMeadow { public static class Git { public const string GIT_BRANCH_NAME = &quot;$(GitBranch)&quot;%3B public const string GIT_REMOTE_URL = &quot;$(GitRemoteUrl)&quot;%3B public const string GIT_HEAD_COMMIT = &quot;$(GitHead)&quot;%3B } }" />
```

RainMeadow.cs
```diff
@@ -178,6 +178,18 @@ private void RainWorld_OnModsInit(On.RainWorld.orig_OnModsInit orig, RainWorld s
...
+                else
+                {
+                    var GIT_HEAD_COMMIT = git.GetField("GIT_HEAD_COMMIT", BindingFlags.Static | BindingFlags.Public)?.GetValue(null);
+                    var GIT_BRANCH_NAME = git.GetField("GIT_BRANCH_NAME", BindingFlags.Static | BindingFlags.Public)?.GetValue(null);
+                    var GIT_REMOTE_URL = git.GetField("GIT_REMOTE_URL", BindingFlags.Static | BindingFlags.Public)?.GetValue(null);
+                    RainMeadow.Info($"You're running a DEBUG build from: {GIT_BRANCH_NAME}\n{GIT_REMOTE_URL}commit/{GIT_HEAD_COMMIT}");
+                }
+#endif
```
cgytrus, line 186: dont need reflection here since youre already inside #if DEBUG and this type will always be included in debug builds

Rain Meadow.csproj
```diff
@@ -44,6 +44,31 @@
...
+
+        <ItemGroup>
+            <Compile Include="$(IntermediateOutputPath)Git.cs" />
```
cgytrus: the file name should end with .g.cs or .gen.cs since its auto-generated
```suggestion
            <Compile Include="$(IntermediateOutputPath)Git.g.cs" />
```

cgytrus (changes requested):
it might be beneficial to also include this in release builds (like if you want to send a release playtesting build to someone it could be useful to have this info) and should definitely somehow be displayed in ui as well, and maybe make ui hide it in unmodified workshop builds (not sure exactly how youd detect this)

also please consider configuring your ide or text editor to not reformat unmodified code so that you dont accidentally push unrelated whitespace/formatting changes

Rain Meadow.csproj
iulh: if this change cuts a hash to 7 letter then i don't know if it's not better to cut them from code if needed instead

RainMeadow.cs
iulh: i really want this to be soft dependendant. even tho it will be always bundled, i don't want it to be a required thing to hold the mod

Rain Meadow.csproj
cgytrus: `git describe` has different behavior, see https://git-scm.com/docs/git-describe

RainMeadow.cs
cgytrus: if its always there theres no way it could fail though?

invalidunits:
This would be nice to add if it effects our client value.

RainMeadow.cs
iulh: there's potentially no way for it to fail, but i feel like explicitly mark it as optional so it's disappearance would change nothing

iulh:
> This would be nice to add if it effects our client value.

relatively? it's a small internal QoL as it could possibly be, mostly for a documentation/tracing reasons

RainMeadow.cs
One-Letter-Shor: If the type is removed this code should also be removed

iulh: > If the type is removed this code should also be removed

whatever, [it's a requirement now](https://github.com/henpemaz/Rain-Meadow/pull/1633/changes/aedbd6a96e1e0a8670dbb7d3a6e725b264bfa75f)

Rain Meadow.csproj
```diff
@@ -44,6 +44,31 @@
...
+
+        <Exec Command='git remote get-url $(GitRemoteName)' ConsoleToMSBuild="true">
+            <Output TaskParameter="ConsoleOutput" PropertyName="GitRemoteUrl" />
+        </Exec>
+
+        <Exec Command="git rev-parse HEAD" ConsoleToMSBuild="true">
+            <Output TaskParameter="ConsoleOutput" PropertyName="GitHead" />
+        </Exec>
+
+        <WriteLinesToFile File="$(IntermediateOutputPath)GitInfo.g.cs" Overwrite="true" Lines="namespace RainMeadow { public static class Git { public const string GIT_BRANCH_NAME = &quot;$(GitBranch)&quot;%3B public const string GIT_REMOTE_URL = &quot;$(GitRemoteUrl)&quot;%3B public const string GIT_HEAD_COMMIT = &quot;$(GitHead)&quot;%3B } }" />
```
cgytrus: the constant should probably not be prefixed with GIT_ because thats already in the type name

also maybe consider doing something like [this](https://github.com/cgytrus/Slughunt/blob/a801ae851381a3ff3c02ecdbcd1f441894a2302b/Slughunt/Slughunt.csproj#L79-L88) here instead of pasting the entire contents of the file directly, would be nicer read and edit

iulh: > the constant should probably not be prefixed with GIT_ because thats already in the type name

yeah i'm doing it. i've got confused with commits order a bit..

> also maybe consider doing something like [this](https://github.com/cgytrus/Slughunt/blob/a801ae851381a3ff3c02ecdbcd1f441894a2302b/Slughunt/Slughunt.csproj#L79-L88) here instead of pasting the entire contents of the file directly, would be nicer read and edit

looks like something Shor would do, it's too much imho. but i see the clarity could be improved

PR #1634, opened 2026-09-01
Optimize registering RPCs and OnlineStates

cgytrus:
significantly speeds up these two stages of startup by checking only assemblies referencing Rain Meadow for registering those automatically instead of every single loaded assembly, making them take up about 25% less time of RM's OnModsInit

also reduces GetExecutingAssembly calls, removes redundant GetCustomAttribute and ToList calls and removes redundant try blocks

before | after
--- | ---
image | image

(look at SetupRPCs and InitializeBuiltinTypes)

also removes OnlineState.RegisterState so technically breaks api/abi but nobody should be using this anyway

### DIFF
--- a/Online/State/OnlineState.cs
+++ b/Online/State/OnlineState.cs
@@ -51,13 +51,11 @@ namespace RainMeadow
         private static Dictionary<StateType, StateHandler> handlersByEnum = new Dictionary<StateType, StateHandler>();
         private static Dictionary<Type, StateHandler> handlersByType = new Dictionary<Type, StateHandler>();
 
-        internal static void InitializeBuiltinTypes()
+        internal static void InitializeBuiltinTypes(IEnumerable<Assembly> assemblies)
         {
             _ = StateType.Unknown; // runs static init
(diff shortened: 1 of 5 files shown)


invalidunits (dismissed):

cgytrus:
improved compatibility (and potentially performance with a lot more mods loaded) by finding assemblies that reference RM then using that list to auto register states/rpcs instead of iterating mod assemblies, so this pr shouldnt change previous behavior now

also removed the nested tries since theyre not needed (GetTypesSafely and IsAssignableFrom shouldnt be able to throw, and if they do something probably went really really wrong anyway)

Online/State/OnlineState.cs
henpemaz, line 1: The shortened assemblies list aside, I don't see the point of the changes in this

cgytrus: The nested try-catches are not needed and RegisterState isn't needed either. The log change is to be consistent with the essentially the same log in RegisterRPCs

PR #1636, opened 2026-09-02
update japanese strings 

solo-snuggles:


One-Letter-Shor:
Just curious, does running `git add . --renormalize` stage text_jap/strings.txt?

With the new override for translation files in .gitattributes it should already be CRLF. I suspect you're telling git to internally store it as CRLF which is not needed

solo-snuggles:
I reformatted because when I had merged main branch -> my local branch, Rider read it as Unix (LF), so I took the time to reformat outside of Rider and that worked for me
it might be just me psyching myself out but i'd rather be safe

PR #1638, opened 2026-09-02
feat: add textfile support for whitelisting mods

6fears7:
Enables users to specify a whitelist of mods, preventing any "client-sided" mods.  Primary use cases are:

1. Users wanting 1-1 parity 
2. Tighter competitive scene

- [x] Playtest


henpemaz (approved):

6fears7:
@invalidunits I don't know if you'd rather me update the lobby creation param like is done here or use the same .AddData() workaround used for cape flying and cheat allowance. Happy to do whatever

invalidunits (approved):

PR #1640, opened 2026-09-02
feat: Slugcat Abilities Menu

EdEnStonne:
**TDLR : The main goal of this PR is to make the slugcat ability menu easier to modify and maintain, by making ui objects specifically suited to Rain Meadow's special syncing system (instead of some dispersed, unorganized code).**
It also makes the UI alot more organized and overhaul prettier (design in video below).

https://github.com/user-attachments/assets/9cccbb0d-f471-4ce8-bcf6-f70b7ed8edbd

This adds alot of new UI elements coming from `OnlineSettingElement`, with some useful functions to properly sync arena values with meadow system, being :
- syncing it with the data holder (usually ArenaOnlineGamemode)
- saving it in the right configurable
- restoring to default value

They also can be positionned using a simple int to tell the order of the parameter, have movement and alpha ease, and have their componant accessible to allow some more elaborated design.

Those elements takes data from the new readonly struct `SettingsTabData` (for tabs) and `SettingsConfigData` (for configurables). They hold all the information those UI needs to properly function.

The new `OnlineSlugcatSettings<TSelf>` is configured to manage those elements autonomously. It takes the data in (before ctor) and use it to make the UI elements, link them properly for controller inputs and reorder them. Those UI elements can be easily accessed later for dynamic updates.

Those are either new classes or subclasses of existing elements. It does not override the existing system, and tabs made by other mods will not be affected.
It is also modder-friendly, as you can use data from other source than ArenaOnlineGamemode.

This PR also comes with :
- @6fears7's Monk shield PR
- redesign of also the score settings menu with those elements
- @Dustylemny's scroller PR

TODO :
- [x] Be able to remake the MSC and Watcher settings with that system.
- [x] Add enum and ExtEnum compat.
- [x] Merge @Dustylemny's scroller system in (pretty sure it's #1629).
- [ ] Make a wiki entry/tutorial on how to use it for modders (in another future PR)
- [x] Testin 'nd Shippin

PR #1642, opened 2026-09-02
feat(arena): add toggelable fruit

6fears7:
depends on #1640  and #1629


6fears7:
Merged in slugcat menu

PR #1645, opened 2026-09-03
fix: arena stat offset midquit

6fears7:
Updates mid game quits to:
- block re-entering on same round
- Clients now read from lobby state of quit users
- Avatars are removed during the PlayerQuit RPC

- [x] tested

### DIFF
--- a/GameModes/ArenaCompetitiveGameMode.cs
+++ b/GameModes/ArenaCompetitiveGameMode.cs
@@ -1185,10 +1185,10 @@ namespace RainMeadow
                 RainMeadow.Debug($"{onlinePlayer} left mid-round; recording as a quitter");
             }
 
-            if (ArenaSession is ArenaGameSession arenaSession)
-            {
-                arenaSession.Players.RemoveAll(ac => ac?.GetOnlineCreature()?.owner == onlinePlayer);
-            }
+            if (ArenaSession is not null) 
+             { 
+                 ArenaSession.Players.RemoveAll(ac => ac?.GetOnlineCreature()?.owner == onlinePlayer); 
+             }
 
             base.PlayerLeftLobby(onlinePlayer);
         }
--- a/GameModes/ArenaCompetitiveGameMode.cs
+++ b/GameModes/ArenaCompetitiveGameMode.cs
@@ -1176,6 +1176,20 @@ namespace RainMeadow
             AllKillsByOPlayer.Remove(onlinePlayer);
             RoundKillsByOPlayer.Remove(onlinePlayer);
 
+            if (OnlineManager.lobby.isOwner
+                && isInGame
+                && arenaSittingOnlineOrder.Contains(onlinePlayer.inLobbyId)
+                && !playersQuitMidRound.Contains(onlinePlayer.inLobbyId))
+            {
+                playersQuitMidRound.Add(onlinePlayer.inLobbyId); // clients receive this via ArenaLobbyData
+                RainMeadow.Debug($"{onlinePlayer} left mid-round; recording as a quitter");
+            }
+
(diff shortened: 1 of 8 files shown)


GameModes/ArenaCompetitiveGameMode.cs
```diff
@@ -164,6 +164,7 @@ public List<string>
         public bool shufflePlayList;
         public List<string> playList = [];
         public List<ushort> arenaSittingOnlineOrder = [];
+        public List<ushort> playersQuitMidRound = [];
```
One-Letter-Shor, line 167: You should be exposing a list of OnlinePlayers instead of ushorts so the data is more meaningful and not able to have invalid values (for the most part)

GameModes/ArenaCompetitiveGameMode.cs
```diff
@@ -1172,6 +1176,20 @@ public override void PlayerLeftLobby(OnlinePlayer onlinePlayer)
...
+                && !playersQuitMidRound.Contains(onlinePlayer.inLobbyId))
+            {
+                playersQuitMidRound.Add(onlinePlayer.inLobbyId); // clients receive this via ArenaLobbyData
+                RainMeadow.Debug($"{onlinePlayer} left mid-round; recording as a quitter");
+            }
+
+            if (ArenaSession is ArenaGameSession arenaSession)
+            {
+                arenaSession.Players.RemoveAll(ac => ac?.GetOnlineCreature()?.owner == onlinePlayer);
+            }
```
One-Letter-Shor: You sure you don't just want a null check? Also can player ACs actually be null?

```suggestion
            if (ArenaSession is not null) 
             { 
                 ArenaSession.Players.RemoveAll(ac => ac?.GetOnlineCreature()?.owner == onlinePlayer); 
             }
```

One-Letter-Shor (changes requested):

GameModes/ArenaCompetitiveGameMode.cs
6fears7: Not here. If a player quits, that OnlinePlayer reference is gone. By tracking the InLobbyid, which will never re-adjust until the lobby lifespan expires, we ensure we never have an off-by-one issue.

GameModes/ArenaCompetitiveGameMode.cs
6fears7: github swallowed my comment i guess. 

AC check being written defensively here doesn't cost us anything to defend against a future change where we might update Creature_Update to actually remove Player Abstract Creatures on abstraction, or for users to remove ACs from a list for their own mods, like in BTW's respawn system.

GameModes/ArenaCompetitiveGameMode.cs
One-Letter-Shor: MeadowPlayerId is the type with a longer lifespan. Arena should probably be changed to only use online player and player id

6fears7: The persistence of the MeadowPlayerId is part of the problem for this. We *dont't* want to track the player persistently (right now).

With a ushort, the quitter keeps their id while they queue for rejoin, and a crashed player returns with another id. 

With a MeadowPlayerId, the returning player matches their own stale entry and we’d have to write explicit cleanup to undo state the ushort can just remove.

Additionally, tracking players who quit as a ushort makes the comparison to the arenaOnlineSittingOrder more straightforward because we're comparing two lists of the same type.  

TL;DR we don't gain anything by converting the type here. If we were to change how stats were managed as being standalone from the position of the user in the InLobbyId, it'd be worth considering

invalidunits (approved):

PR #1650, opened 2026-09-05
Fix story gate transition readiness and premature closing

mkiera:
Story gates can initiate a region switch before the host approves the group, or close the middle door immediately after loading while players are still on the entrance side.

Return from the gate update after advertising local readiness until the shared transition state permits opening. Keep the local readiness flag set when beginning the transition so the host continues waiting for players to pass through. The existing AllPlayersThroughToOtherSide hook clears each flag when that player's local crossing check succeeds.

- A local harness applies the source IL hook to a reduced gate update and uses the source crossing hook and host completion block. It checks waiting for approval, keeping the middle open before crossing, waiting for the slower player, closing after both cross, delayed clients, and offline initiation.
- A two-player test with both players using the patched build reports successful crossing. The inspected client log contains two completed VS to SI world loads in separate attempts and later shelter discovery in SI, without the previous region-request failure loop.
- Host console and exception logs show VS and SI loading, with no gate exception recorded.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Release 1.16.0
 
+## Story
+- Fixed gates starting region transitions before all players are ready and closing the middle door before all players have crossed
+
 ## Arena
 - Fixed infinite tinnitus again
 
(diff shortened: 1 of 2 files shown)


6fears7 (dismissed):

6fears7:
add a changelog update please

6fears7 (approved):

PR #1659, opened 2026-09-13
fix: saint tongue custom color

6fears7:


6fears7 (commented):
Add comment of changed IL

PR #1660, opened 2026-09-14
partially fix menu save game data leaking

r3nsen:
this pr fix some problems related to client local save being used to load data, including:
- loading client local data when menuSaveGameData is null
- "As a client viewing the campaign select, if the host switches to and then away from(?) an empty campaign slot, the client's UI will crash."
- "Joining a story lobby starts on Survivor and scrolls to the current campaign, instead of starting at the current campaign like it used to."

the switch campaign bug was related to switching from a campaign with save data to a one without save data, when menuSaveGameData was null, instead of returning null it was returning the local save, and when the current campaign also had  no data on client side, SlugcatPageContinue tried to get fields from a null class ref

this pr does NOT fix:
- current game time desync
- some special campaign arts desync, like saint The Cycle
- "Mark of communication, the stomach slot, and the creature reputation table all pull from the local meadow save instead of the online meadow save."


r3nsen:
video of the problems:

https://github.com/user-attachments/assets/ae5785db-8989-49a7-8347-27dc77e23298

- showing individual saves in each game instance
- showing that client gets data locally when host campaign has no data
- showing game crashing when transitioning from a saved campaign to a clean campaign (if client also has no data on it)

r3nsen:
The same problems solved

https://github.com/user-attachments/assets/b5279b95-15bc-4e8e-b6a1-1086a460878b

Menu/RainMeadow.MenuHooks.cs
```diff
@@ -65,6 +66,53 @@ private void MenuHooks()
...
+                return false;
+            });
+            c.Emit(OpCodes.Brtrue, skip);
+
+
+
+            c.Emit(OpCodes.Ldarg_0);
+            
+            c.GotoNext(MoveType.After,
+                x => x.MatchCallvirt("System.Collections.Generic.Dictionary`2<SlugcatStats/Name,Menu.SlugcatSelectMenu/SaveGameData>", "set_Item")
```
6fears7, line 112: would be nice to use `typeof` here

Story/StoryOnlineMenu.cs
```diff
@@ -350,6 +350,15 @@ public override void Update()
             {
                 RainMeadow.Debug("page refresh");
                 storyGameMode.needMenuSaveUpdate = false;
+                
+                if (OnlineManager.lobby.isOwner)
+                {
+                    for (int i = 0; i < slugcatColorOrder.Count; i++)
+                    {
+                        saveGameData[slugcatColorOrder[i]] = MineForSaveData(manager, slugcatColorOrder[i]);
```
6fears7, line 358: Are we sure we want to MineForSaveData for every slugcat every frame?

r3nsen: Oh, no. For some reason I thought that would be called only when host changes when I wrote

6fears7: its possible thats the behavior, I just saw it being called in Update() and flagged it

r3nsen: Oh right, I checked here, not calling every frame, but is double calling every time host goes to online menu, one internally by `SlugcatSelectMenu` constructor, and one by this code cause `needMenuSaveUpdate` is setted by `newResourceOwner`, and `newResourceOwner` sets `needMenuSaveUpdate` every time its called and its host, and not only if its a lobby resource as I thought

r3nsen: now it only run this if its a new owner

Menu/RainMeadow.MenuHooks.cs
r3nsen: done

6fears7 (dismissed):
LGTM, great work

Menu/RainMeadow.MenuHooks.cs
```diff
@@ -65,6 +66,53 @@ private void MenuHooks()
             _ = Ext_HUD_OwnerType.RainMeadowOverlay;
             On.ProcessManager.Update += ProcessManager_Update_UpdateOverlay;
             IL.ProcessManager.InitFadeSprite += ProcessManager_InitFadeSprite_SwitchTextSide;
+
+            IL.Menu.SlugcatSelectMenu.ctor += SlugcatSelectMenu_ctor;
+        }
+
+        private void SlugcatSelectMenu_ctor(ILContext il)
```
6fears7, line 73: Please consider wrapping this in a Try/Catch with a logged error for ensuring we grab any errors from future game updates breaking this hook.

r3nsen: done!

6fears7 (approved):

PR #1661, opened 2026-09-16
fix: watcher menu button greyed out, broken ripple screen

6fears7:


6fears7:
@invalidunits not sure if you saw this one needing your review

invalidunits (approved):

